Issue #17, opened 2021-12-13
Run Protractor tests in CI pipeline


AureliaKhorsand:
**Acceptance Criteria**
- At least one protractor end-to-end test runs successfully in the Circle CI environment
- The protractor end-to-end tests are runnable in the docker-compose development environment via script or single command.

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

albertcrowley:
This is a CI pipeline task, so no specific QA outside of code review needed.

WiseQA:
As noted no QA action required on this ticket - moved to closed as ticket complete.

PR #27, opened 2021-12-29
Circleci unit test angular

mjtravers:
Here is the updated CircleCI config.yml with 2 new jobs: one for the Angular unit tests and one for the Angular end-to-end tests.

The empty style.css file needed to be added to fix an issue running the e2e tests.

This PR pertains to issue #17 but I'm not including the magic words (i.e. "resolves" or "fixes") to automatically link to the issue until we find out how that will affect the Jira workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ version: 2.1
 orbs:
   # See the orb documentation here: https://circleci.com/developer/orbs/orb/circleci/python
   python: circleci/python@1.2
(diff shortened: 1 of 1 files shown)


albertcrowley:
Confirmed that the e3e tests are setup and running. It is expected that the actual tests will generate failures, so no problem with that.

The angular unit tests are generating a number of compile errors. They appear to be legitimate errors in the code and not in the configuration of the CI pipeline - so again, these failures are the proper output until we get the code fixed.

Passing code review and merging pull request.

PR #29, opened 2022-01-01
Feature/15 fix django unit tests

albertcrowley:
* Modified settings.py to use the default database as the test database so we can have real data
* Added a default for CommitteeInfo.created_at column
* Fixes to the test_login_success to use proper endpoint
* Added test of the returned JWT to the test_login_success
* Modified committee tests to delete any leftover duplicate test committees during setup.


mjtravers (approved):
Code changes look good for baseline testing and getting CircleCI running.

Issue #47, opened 2022-01-14
Automated front-end tests needs back-end support


albertcrowley:
The Cypress e2e tests need to be able to spin up a baseline test database each time it is run. The database should spin up a baseline Django database and then use Django fixtures to update the database schema to the version being tested.

***User Story***

  As a developer/QA, 

  I would like to have automated end to end testing that exercises full workflows that our end users will run

  so that we can catch and regressions occurring in integration

***Acceptance Criteria***

*   e2e tests should start a working database populated with enough data to run tests
*   e2e job should be configured so that Cypress tests can login and run the existing e2e tests
*   All e2e tests should be passing.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1347](https://fecgov.atlassian.net/browse/FECFILE-1347)

mjtravers:
@albertcrowley Do you envision this database/endpoint being a custom Docker image with a test database and a snapshot of the backend code running? So the bulk of this ticket would be creating a Dockerfile and figuring out an automated way of keeping it up to date?

mjtravers:
This ticket is OBE. e2e tests are too long running to be practical in CircleCI. Will run e2e tests in another automated fashion TBD

mjtravers:
@albertcrowley On second thought, because of the need for an isolated database that needs to be reinitialized before running tests, I'd like to look into if CircleCI can be used ad hoc and/or on a time interval basis. Re-opening ticket.

WiseQA:
QA Review of the following:
- e2e tests should start a working database populated with enough data to run tests
- e2e job should be configured so that Cypress tests can login and run the existing e2e tests
- All e2e tests should be passing.
Automated testing working and passing.
QA review completed and passed.  Moved to Stage Ready

PR #51, opened 2022-01-14
Feature/18 upgrade django

albertcrowley:
* Upgraded Django to 3.2.11 LTS
* Upgraded the minimum number of other python packages needed to make the existing tests pass: cffi, django-appconf, django-compressor, django-storage, djangorestframework
* Modified AppConfig settings to fix breaking change in Django 3.2 
* e2e tests now all passing in development environment.  Issue #47 opened to ensure tests pass in CI
* Made improvements to Protractor e2e tests to validate (as best as possible) the upgrade
  * Added environment variable to configure the base URL for protractor so it can be run in different build environments
  * Added successful login test and view contacts test
  * Wrote a test for add contact, but can not work around bug in state dropdown
  * Now skip all the legacy tests that are failing. (most are soon to be OBE anyway)
  * Small changes to HTML to facilitate testing. Things like adding IDs to tags for easier matching


mjtravers (approved):
The additional tests created are great. One small nit is when creating an xpath string with single-quotes embedded in it, let's wrap the entire string in double-quotes so the inner single-quotes do not have to be escaped. Makes for easier readability. I fixed and committed the changing of the single-quotes in the PR

PR #79, opened 2022-01-20
Feature/41 split repo

albertcrowley:
* Removed Django directories. 
* Modify CI config to remove python jobs
* changed npm package installs in CI to using install-packages so it would be cached
* adding package-lock.json to the repo
* Increased memory limit to 4GB for ng in CI and testing to avoid crashes
* Removed API and DB from docker-compose setup
* Modified the Nginx proxy setting in docker-compose to point to host's localhost:8080 where the API will be running


mjtravers (approved):
Both Docker and CircleCI config files look good. I was able to run the docker-compose set up locally without any issues.

PR #85, opened 2022-01-28
feature/40 deploy to cloud.gov

albertcrowley:


tasks.py
```diff
@@ -0,0 +1,261 @@
...
+import os
+import sys
+
+from shutil import copyfile
+from invoke import task
+
+
+APP_NAME = "fecfile-web-app"
+ORG_NAME = "fec-fecfileonline-prototyping"
+APP_BUILD_NAME = "fec-eFilling"
```
lbeaufort: Do we need `APP_BUILD_NAME`? If so, should it be "fec-eFiling"?

tasks.py
```diff
@@ -0,0 +1,261 @@
...
+        pass_var = f'$FEC_CF_PASSWORD_{space.upper()}'
+
+        if os.getenv(user_var) and os.getenv(pass_var):
+            user = os.getenv(user_var)
+            passwd = os.getenv(pass_var) or ''
+            print(f"passwd is {passwd}")
+            safe_passwd = ('*' * len(passwd )) + os.getenv(passwd)[-3:]
+
+            print(f"${user_var} = {user}");
+            print(f"${pass_var} = {safe_passwd}");
```
lbeaufort: Even though we're only printing user and a partial password, this gives me pause. This gets output to CircleCI, which is public. Would you be open to remove printing out the user and safe password?

tasks.py
```diff
@@ -0,0 +1,261 @@
...
+    if space is None:
+        print("The current configuration does not require a deployment to cloud.gov.   ")
+        return None
+    print("Detected space {space}".format(**locals()))
+    return space
+
+
+def _detect_prod(repo, branch):
+    """Deploy to production if master is checked out and tagged."""
+    if branch != "master":
```
lbeaufort: Let's make this `main`, this is my oversight.

tasks.py
```diff
@@ -0,0 +1,261 @@
...
+            print(f"   cf service-key  [my-server-account-name] [my-service-key-name]")
+
+        exit (1)
+
+def _do_deploy(ctx, space):
+    orig_directory = os.getcwd()
+    os.chdir( os.path.join(orig_directory, 'front-end', 'dist') )
+    print (f'new dir {os.getcwd()}')
+
+    manifest_filename = os.path.join(orig_directory, "deploy-config", f"fecfile-web-app-{space}-manifest.yml")
```
lbeaufort: Can we use `APP_NAME` here?

tasks.py
albertcrowley: Doesn't look like it is needed.  It was present in the original tasks.py.  I'm removing it.

tasks.py
albertcrowley: Changed

tasks.py
albertcrowley: I'm fine removing it.  The purpose is to help with debugging later. I'll print the env var names only and that should be enough to help people out.

tasks.py
albertcrowley: Yes.

lbeaufort (approved):
Looks great, thanks @albertcrowley!

front-end/src/environments/environment.cloud.gov.prod.ts
```diff
@@ -0,0 +1,24 @@
+// This file can be replaced during build by using the `fileReplacements` array.
+// `ng build --configuration=cloud.gov.prod` replaces `environment.ts` with `environment.cloud.gov.prod.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+    production: false,
+    name: 'development',
```
mjtravers: this should be production

front-end/src/environments/environment.cloud.gov.stage.ts
```diff
@@ -0,0 +1,24 @@
+// This file can be replaced during build by using the `fileReplacements` array.
+// `ng build --configuration=cloud.gov.stage` replaces `environment.ts` with `environment.cloud.gov.stage.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+    production: false,
+    name: 'development',
```
mjtravers: this should be stage

front-end/src/environments/environment.cloud.gov.stage.ts
```diff
@@ -0,0 +1,24 @@
...
+    apiUrl: 'https://fecfile-web-api-stage.app.cloud.gov/api/v1',
+    appTitle: 'FECfile',
+    validateSuccess: 'All required fields have passed validation.',
+    awsRegion: 'us-east-1',
+    awsIdentityPoolId: 'us-east-1:f0f414b2-8e9f-4488-9cc1-34a5918a1a1d',
+    ACCESS_KEY: "_process.env.ACCESS_KEY", //TODO: fix this. Hacked to get Angular 8 upgrade working
```
mjtravers: remove "hack" comments

front-end/src/environments/environment.cloud.gov.prod.ts
```diff
@@ -0,0 +1,24 @@
+// This file can be replaced during build by using the `fileReplacements` array.
+// `ng build --configuration=cloud.gov.prod` replaces `environment.ts` with `environment.cloud.gov.prod.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+    production: false,
```
mjtravers: set to true

front-end/src/environments/environment.ts
```diff
@@ -1,18 +1,18 @@
...
-  appTitle: 'FECfile',
-  validateSuccess: 'All required fields have passed validation.',
-  awsRegion: 'us-east-1',
-  awsIdentityPoolId: 'us-east-1:f0f414b2-8e9f-4488-9cc1-34a5918a1a1d',
-  ACCESS_KEY: "_process.env.ACCESS_KEY", //TODO: fix this. Hacked to get Angular 8 upgrade working
-  SECRET_KEY: "_process.env.SECRET_KEY", //TODO: fix this. Hacked to get Angular 8 upgrade working
-  dcfConverterApiUrl: 'https://dev-efile-api.efdev.fec.gov/dcf_converter/v1'
+    production: false,
+    name: 'development',
+    apiUrl: 'https://fecfile-web-api-dev.app.cloud.gov/api/v1',
```
mjtravers: should this be 'localhost' since we have a separate environment for dev.cloud.gov?

front-end/src/environments/environment.ts
```diff
@@ -1,18 +1,18 @@
...
+    appTitle: 'FECfile',
+    validateSuccess: 'All required fields have passed validation.',
+    awsRegion: 'us-east-1',
```
mjtravers, line 11: I don't think we need the AWS config variables, let's remove them

front-end/src/environments/environment.cloud.gov.dev.ts
```diff
@@ -0,0 +1,24 @@
...
+// `ng build --configuration=cloud.gov.dev` replaces `environment.ts` with `environment.cloud.gov.dev.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+    production: false,
+    name: 'development',
+    apiUrl: 'https://fecfile-web-api-dev.app.cloud.gov/api/v1',
+    appTitle: 'FECfile',
+    validateSuccess: 'All required fields have passed validation.',
+    awsRegion: 'us-east-1',
```
mjtravers, line 11: Are these AWS variables needed for cloud.gov? If not let's remove them from here and the other environment.*.ts files

mjtravers (changes requested):
Found some changes to be made in the environment.*.ts files that need to be made. Go ahead and merge after fixes are made, move the ticket to QA and assign it to Shelly

front-end/src/environments/environment.cloud.gov.prod.ts
albertcrowley: fixed

front-end/src/environments/environment.cloud.gov.stage.ts
albertcrowley: fixed

front-end/src/environments/environment.cloud.gov.stage.ts
albertcrowley: fixed

front-end/src/environments/environment.cloud.gov.prod.ts
albertcrowley: fixed

front-end/src/environments/environment.ts
albertcrowley: Yes.  Changed.

front-end/src/environments/environment.ts
albertcrowley: The awsRegion and awsIdentityPoolId is actually used in the upload-contact.service.ts and upload-trx.service.ts.  Do we still want to remove them from environemnt files/

albertcrowley: Just did a test and "ng server" generates errors is we remove awsRegion or awsIdentityPoolId so I am going to leave them in place.

front-end/src/environments/environment.cloud.gov.dev.ts
albertcrowley: Just did a test and "ng server" generates errors is we remove awsRegion or awsIdentityPoolId so I am going to leave them in place.

mjtravers (approved):

PR #93, opened 2022-02-01
feature/59-sonarcloud

albertcrowley:
* added sonar-project.properties


mjtravers (approved):
Looks good. Noted that the CircleCI config.yml will need to be updated after the Angular unit tests are fixed

PR #106, opened 2022-02-04
Adding manifest files for prod and stage

albertcrowley:
When updating and testing the acceptance criteria for the issue, I saw that we will have to actually deploy to Stage and Prod for this to be testable.  In doing that testing, I saw that the stage and prod manifest files were not yet created.  This follow up PR adds them in to the repo with the same defaults (other than route) as the development manifest.


mjtravers (approved):
Looks good. Will merge.

PR #117, opened 2022-02-14
Updated settings to make staging environment match the production environment

lbeaufort:
Per @mjtravers, updated settings to make staging environment match the production environment so we can catch issues sooner.


mjtravers (approved):
Thanks. I'll look into which option is the troublemaker by way of the new ticket.

PR #124, opened 2022-02-16
Upgraded code base to Angular v13

mjtravers:


albertcrowley (approved):
It builds for me and test run on my local environment.  I tried some manual testing on the front end and I got an error adding a contact which had been working previously for me.  I think we were expecting some regressions so that is not a showstopper for me. Just wanted you to be aware.

Approving the PR to merge into develop.

PR #141, opened 2022-02-19
Added temporary snyk exception and adjusted angular budget limits

mjtravers:


toddlees (approved):
Ignoring typeahead vulnerability because there is no impact by the expiry date

PR #150, opened 2022-03-02
Fixed contacts list after Angular 13 upgrade

mjtravers:


front-end/src/app/contacts/contacts-table/contacts-table.component.ts
```diff
@@ -299,7 +299,7 @@ export class ContactsTableComponent implements OnInit, OnDestroy {
         }
 
         this._contactsService.addUIFileds(res);
-        this._contactsService.mockApplyFilters(res, this.filters);
+        // this._contactsService.mockApplyFilters(res, this.filters);
```
toddlees, line 302: Is it worth keeping this line as a comment?

mjtravers: Yes, for now as the filter code that is not working is in that call so we know where the call was made later on.

toddlees (approved):

Issue #156, opened 2022-03-08
Reorganize web-app code base


mjtravers:
Business Reason

In order to be able to develop and test the existing and new features of the app, move existing code that is not part of OP1 to the code backlog.

***Tasks***

*   Identify code that is not necessary for OP1 features and move to the code backlog in the repo.
*   Update package.json to suspend phased out dependencies

Acceptance Criteria

*   Users can log into site
*   Dashboard with navigation presents at login

QA

*   Can log into site and view dashboard

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1271](https://fecgov.atlassian.net/browse/FECFILE-1271)

WiseQA:
Log into https://fecfile-web-app-dev.app.cloud.gov/ using test account.   Verified Dashboard appeared.  (Note links not working just display of naming).

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/74eab81c-27ca-4189-bf87-6bf59c4c36c4)

QA review completed.  Moved to Stage Ready.

PR #158, opened 2022-03-10
Reorganized code to optimize compilation

mjtravers:
Issue #156


front-end/package.json
```diff
@@ -29,38 +31,22 @@
     "@angular/platform-browser": "~13.2.0",
     "@angular/platform-browser-dynamic": "~13.2.0",
     "@angular/router": "~13.2.0",
-    "@fortawesome/fontawesome-free": "~5.15.4",
-    "@kolkov/angular-editor": "~2.0.0",
-    "@ng-bootstrap/ng-bootstrap": "~11.0.0",
-    "@ng-select/ng-select": "~8.1.1",
+    "@ngrx/effects": "^13.0.2",
+    "@ngrx/store": "^13.0.2",
+    "@ngrx/store-devtools": "^13.0.2",
```
toddlees, line 36: should this be a dev dependancy?

front-end/src/app/dashboard/dashboard.component.ts
```diff
@@ -1,199 +1,8 @@
...
-    if (reason === ModalDismissReasons.ESC) {
-      return 'by pressing ESC';
-    } else if (reason === ModalDismissReasons.BACKDROP_CLICK) {
-      return 'by clicking on a backdrop';
-    } else {
-      return `with: ${reason}`;
-    }
-  }
-}
+export class DashboardComponent {}
```
toddlees, line 8: 👍 👍 👍

front-end/src/app/shared/models/contact.model.ts
```diff
@@ -0,0 +1,71 @@
+import { plainToClass } from 'class-transformer';
+import { BaseModel } from './base.model';
+import { LabelList } from '../utils/label.utils';
+
+export enum ContactTypes {
+  CANDIDATE = 'cand',
```
toddlees, line 6: Are these the values you were saying we'd need to use in the api?  What's the source of these?

front-end/src/app/shared/scss/_global.scss
```diff
@@ -1,24 +1,21 @@
...
 @import '../../../../node_modules/bootstrap/scss/grid';
-@import '../../../../node_modules/bootstrap/scss/dropdown';
-@import '../../../../node_modules/bootstrap/scss/tooltip';
-@import '../../../../node_modules/bootstrap/scss/card';
-@import '../../../../node_modules/bootstrap/scss/nav';
-@import '../../../../node_modules/bootstrap/scss/navbar';
-@import '../../../../node_modules/bootstrap/scss/modal';
-@import '../../../../node_modules/bootstrap/scss/utilities';
-@import '../../../../node_modules/bootstrap/scss/forms';
+// @import '../../../../node_modules/bootstrap/scss/dropdown';
```
toddlees, line 5: Can we remove these?

front-end/src/app/shared/services/api.service.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+
+describe('ApiService', () => {
+  let service: ApiService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(ApiService);
+  });
+
+  xit('should be created', () => {
```
toddlees, line 13: does `xit` disable the test?  do we want these to be running?

toddlees (approved):
This is great! Overall approved

front-end/src/app/shared/models/contact.model.ts
mjtravers: Those are not the correct values. I have the correct values in the branch I'm working on now. I will push at the end of the day and send you a link to see them.

front-end/src/app/shared/services/api.service.spec.ts
mjtravers: No, not yet. Work needs be done in the regular tickets for the tests to get up and running.

front-end/src/app/shared/scss/_global.scss
mjtravers: There are a number of files that are hold overs to keep the login pages up and running until we migrate over to the new login system. These directories will be phased out after we have the new login setup:

- app/login
- app/shared/interceptors
- app/shared/services/AuthService
- app/shared/services/SessionService
- app/shared/scss

front-end/package.json
mjtravers: I thought so too but the install instructions had a reason and also a setting so that it wasn't active in production. However, it doesn't seem to be working that well. I noticed that it isn't keeping up in the chome UI plugin what is actually in the store. I installed the "hydration" package that syncs the store with the browser localStorage and that is working great and allows better tracking. I am going to remove the store-devtools in the next ticket

Issue #145, opened 2022-02-23
Allow user to edit an existing Committee contact


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to edit an existing Committee contact so that I can maintain an accurate database of contacts as contact metadata such as addresses, employers, etc. change over time.

 Acceptance Criteria 
 

**If** I am a Committee Administrator
**When** I access my committee's contact list and select a Committee type contact
**Then** I will be able to edit and save any of that contact's data **except Type**

Dev Tasks:
1. refactor old code
2. unit tests
3. documentation
4. logging in the back-end
5. accessible in the UI

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

AureliaKhorsand:
Pointed at 3 based on 2/23/22 backlog refinement discussion. Based off pointing of #68.

WiseQA:
Contacts screen selecting Committee contact:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/bb139985-7033-4606-97e3-61eebe0840e0)

Successful - Contact Updated Popup Screen Shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/da80583c-d589-4c3a-be8e-aed5ee174b42)

Contacts screen showing Committee edit name successful:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/5a009073-972a-45fa-ac5a-a88c801d7121)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/fbf4a25f-de64-4238-9f13-40d61ccf9ae9/download](https://app.zenhub.com/files/306118986/fbf4a25f-de64-4238-9f13-40d61ccf9ae9/download)

QA review completed.  Moved to Stage Ready.

MitchellTCG:
Story Points increased because Todd needs to address "Except Type" portion of the acceptance criteria for all four "type"

WiseQA:
Verified patch 145 implemented.  Once contact is created (e.g. Individual, Committee, Organization and/or Candidate) the contact type cannot be modified or changed.   All 4 contacts passed this merge.

QA review completed on this patch and passed.

PR #176, opened 2022-03-23
Patch 145 disable edit of contact type

toddlees:
#145 
When editing an existing contact, the type field is disabled

### Testing
open the contacts page
click the "new +" button to create a new contact.
you should be able to change the contact type in this modal
save the contact
find the contact in the contacts list and click the edit ✏️ button
you should NOT be able to change the contact type in this modal


rfultz (approved):
Looks right to me!

Issue #144, opened 2022-02-23
Allow user to edit an existing Candidate contact


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to edit an existing Candidate contact so that I can maintain an accurate database of contacts as contact metadata such as addresses, employers, etc. change over time.

 Acceptance Criteria 
 

**If** I am a Committee Administrator
**When** I access my committee's contact list and select a Candidate type contact
**Then** I will be able to edit and save any of that contact's data **except Type**

 Dev Notes 
 
So, that a Contact name will appear on the Contact Page, find a way for the name to display in the  Contact Table:
 **First Name**  **Last Name**

Dev Tasks:
1. refactor old code
2. unit tests
3. documentation
4. logging in the back-end
5. accessible in the UI

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

AureliaKhorsand:
Pointed at 3 based on 2/23/22 backlog refinement discussion. Based off pointing of #68.

WiseQA:
Contacts screen selecting Candidate contact:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/818cdbba-07ea-4f70-abbb-a44cf0764972)

Successful - Contact Updated Popup Screen Shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/f73c4322-e2d7-44eb-be33-a10ed303d86d)

Contacts screen showing Candidate edit name successful:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/dec8a14d-7a59-42ae-a13b-f6a73d58c839)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/d57f3256-d38c-4117-bcc7-f2e2bab0b700/download](https://app.zenhub.com/files/306118986/d57f3256-d38c-4117-bcc7-f2e2bab0b700/download)

QA review completed.  Moved to Stage Ready.

Issue #68, opened 2022-01-18
Allow user to edit an existing Individual contact


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to edit an existing contact so that I can maintain an accurate database of contacts as contact metadata such as addresses, employers, etc. change over time.

 Acceptance Criteria 
 

**If** I am a Committee Administrator
**When** I access my committee's contact list and select an Individual type contact
**Then** I will be able to edit and save any of that contact's data **except Type**

 Dev Notes 
 
So, that a Contact name will appear on the Contact Page, find a way for the name to display in the  Contact Table:
 **First Name**  **Last Name**

Dev Tasks:
1. refactor old code
2. unit tests
3. documentation
4. logging in the back-end
5. accessible in the UI

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

AureliaKhorsand:
Per Ryan 2/10/22 :  [Committee and Candidate IDs] should be editable in case there is a reason they entered the incorrect ID. There are instances where committees and candidates register more than once, which creates multiple IDs.  These scenarios are rare, but again we should allow the user to update if needed. For now, we want to give users maximum flexibility in their data entry.

AureliaKhorsand:
**Refinement Meeting 2/23/22:** Split out ticket by contact type to aid in testing. Additional contact type tickets will also be a 3, team agreed on call.

WiseQA:
Contacts screen selecting Individual contact:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/68d594fa-dfa6-42e4-9c50-7107de63f5a3)

Successful - Contact Updated Popup Screen Shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/28afe654-0dd1-4c77-9b90-6c0c97d5fac2)

Contacts screen showing Individual edit name successful:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/ac16577b-55f3-469f-8a57-9e442c143f9a)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/930a2c98-3ee0-4ba8-a953-862e1a2b0b98/download](https://app.zenhub.com/files/306118986/930a2c98-3ee0-4ba8-a953-862e1a2b0b98/download)

QA review completed.  Moved to Stage Ready.

PR #200, opened 2022-04-01
Changed name display for individual and candidate contacts

mjtravers:
#68 #144


toddlees (approved):
Approving assuming the ticket language was incorrect initially and the design team has changed it to read `FIRSTNAME LASTNAME`

Issue #87, opened 2022-01-31
Allow user to view Committee Profile data (read-only)


AureliaKhorsand:
Business Reason

As a Committee Administrator, I will be able to view my Committee Profile data so that I can verify it is all correct and determine if changes are needed.

Technical Tasks

*   Set up API pulls from the FECFile web app to open FEC API (public key) to get the most recent committee account data to provide the data needed for the front end web page.
*   Create read-only Angular screen under Profile -> Account menu item
*   API endpoint is [/committee/committee_id/]([https://api.open.fec.gov/developers/#/committee/get_committee__committee_id__](https://api.open.fec.gov/developers/#/committee/get_committee__committee_id__))

Acceptance Criteria

***If*** I am a Committee Administrator

  ***When*** I go to the Account page

  ***Then*** I will see Committee Profile data from Form 1 as read-only.

Fields to be shown (if no data from FEC, will be blank):

General

*   Committee ID
*   Committee Name
*   Type of Committee _(pull directly from committee_type_full field in API)_
*   Address
*   Apartment, Suite, Etc.
*   City
*   State/Province
*   Zip/Postal Code
*   Email 1
*   Website

Treasurer

*   First Name
*   Last Name
*   Address
*   Apartment, Suite, Etc.
*   City
*   State/Province
*   Zip/Postal Code
*   Telephone

Custodian of Records _(pull directly from custodian_name_full field in API)_

_Note: remove Candidate and Bank information currently in app._

![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/dbe1c288-cfa4-43a3-895b-71e4d2f0228a](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/dbe1c288-cfa4-43a3-895b-71e4d2f0228a))

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1319](https://fecgov.atlassian.net/browse/FECFILE-1319)

AureliaKhorsand:
Per email from Ryan Lanz over email on 2/16/22, Candidate information should be removed since it does not apply to Form 3X. We will need to add it back in, but it should only appear for Form 3 and Form 3P committees. This will need additional research when we pick this up in OP3, such as a link to the candidate's Form 2.

[...]

PR #210, opened 2022-04-04
Allow user to view Committee Profile data (read-only)

toddlees:
Issue #87


front-end/src/app/layout/header/header.component.ts
```diff
@@ -54,7 +54,7 @@ export class HeaderComponent implements OnInit {
         label: 'Profile',
         icon: 'pi pi-fw pi-user',
         items: [
-          { label: 'Account' },
+          { label: 'Account', routerLink: ['/profile'] },
```
mjtravers: Let's change the /profile url to /account so it matches more closely with the menu label for clarity

front-end/src/app/profile/profile-routing.module.ts
```diff
@@ -1,7 +1,15 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
+import { ProfileComponent } from './profile.component';
 
-const routes: Routes = [{ path: '**', redirectTo: '' }];
+const routes: Routes = [
+  {
+    path: '',
+    component: ProfileComponent,
```
mjtravers: This is a little confusing naming the account info component the same as the module it is in. Let's rename the new component "AccountComponent" and move it into its own subdirectory: src/app/profile/account. The path should be "/account" to match the path in the menu

front-end/src/app/profile/profile.component.html
```diff
@@ -0,0 +1,223 @@
...
+        </p>
+      </div>
+    </div>
+    <div class="col-9">
+      <div class="field">
+        <h4>WEBSITE</h4>
+        <p>
+          {{ (committeeAccount$ | async)?.website }}
+        </p>
+      </div>
```
mjtravers: app210-email-website-overlap
The email and website content is overlapping

front-end/src/app/shared/models/committee-account.model.ts
```diff
@@ -2,74 +2,6 @@ import { plainToClass } from 'class-transformer';
...
-    CommitteeTypes.QUALIFIED_NON_PARTY_WITH_SEP_ACCOUNT,
-    'Qualified Non-Party with Separate Contribution Account with Carey cover letter',
-  ],
-  [CommitteeTypes.PARTY_NON_QUALIFIED, 'Party Non-Qualified'],
-  [CommitteeTypes.PARTY_QUALIFIED, 'Party Qualified'],
-  [CommitteeTypes.PARTY_NON_FEDERAL_ACCOUNTS, 'Party Non-Federal Accounts - Schedules L & I'],
-];
-
 export enum CommitteeDesignations {
   PRINCIPLE_CAMPAIGN_COMMITTEE = 'P',
```
mjtravers: I think we can get rid of the CommitteeDesignations enum, CommitteeDesignation type, and CommitteeDesignationLabels too

front-end/src/app/shared/services/FecApiService/fec-api.service.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            "count": 1,
+            "pages": 1
+        },
+        results: [committeeAccount]
+      }
+
+      service.getDetails("C00601211")
+        .subscribe(committeeAccountData => {
+          expect(committeeAccountData).toEqual(committeeAccount);
+        });
```
mjtravers: So the service call to the FEC endpoint is being intercepted and mocked, yes?

front-end/src/app/shared/services/FecApiService/fec-api.service.ts
```diff
@@ -0,0 +1,35 @@
+import { HttpClient } from '@angular/common/http';
+import { environment } from '../../../../environments/environment';
+import { Injectable } from '@angular/core';
+import { map, Observable } from 'rxjs';
+import { CommitteeAccount } from '../../models/committee-account.model';
+import { FecApiPaginatedResponse } from 'app/shared/models/fec-api.model';
+
```
mjtravers: Would you move the fec-api.service.ts and fec-api.service.spec.ts up a directory into the shared/services directory? I left AuthService and SessionService in own directories because they will be removed after Login.gov set up. It's legacy code

front-end/src/app/shared/services/committee-account.service.ts
```diff
@@ -1,20 +1,27 @@
...
    * @return     {Observable}  The commitee details.
    */
   public getDetails(): Observable<CommitteeAccount> {
-    return this.apiService.get<CommitteeAccount>('/core/get_committee_details');
+      const userLoginData$ = this.store.select(selectUserLoginData);
+      return userLoginData$.pipe(
+        concatMap(userLoginData => {
+          return this.fecApiService.getDetails(userLoginData.committee_id);
+        })
+      );
```
mjtravers: I had to look at this and how it is pulled into the ngrx effect a couple of times to sort out the mergeMap and concatMap are working together

sonar-project.properties
```diff
@@ -11,6 +11,8 @@ sonar.sources=front-end
 # Encoding of the source code. Default is default system encoding
 #sonar.sourceEncoding=UTF-8
 
+# Exclude spec files from coverage
+sonar.coverage.exclusions=**.spec.ts
 sonar.javascript.lcov.reportPaths=front-end/coverage/front-end/lcov.info
```
mjtravers, line 16: Nice find on the sonar config variable

front-end/src/app/profile/profile.component.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+import { provideMockStore } from '@ngrx/store/testing';
+
+import { ProfileComponent } from './profile.component';
+
+describe('ProfileComponent', () => {
+  let component: ProfileComponent;
+  let fixture: ComponentFixture<ProfileComponent>;
+
+  beforeEach(async () => {
+    let committeeAccount: any = {
```
mjtravers: change the type from any to CommitteeAccount

mjtravers (changes requested):
This is really nice and clean. Great job figuring out the *.spec.ts code. My comments are mostly rearranging things for file organization and naming.

mjtravers:
I ran ANDI against the account page and the only issues that were reported were in the layout menu items. So this PR passes the 508 compliance check.

front-end/src/app/shared/services/FecApiService/fec-api.service.spec.ts
toddlees: not actually calling FEC. httpTestingController tells the test if that url is requested (test part 1) and returns the mock response which line 40 checks (test part 2)

front-end/src/app/shared/services/committee-account.service.ts
toddlees: is mergemap a better option?

toddlees:
@mjtravers I think i resolved the email overlap thing.  It was happening because email and website were both fields where the value is a single 'word'.  Normally wordwrap doesn't break words.  I added a css class to these two fields to break long words.  I made a judgment call over truncating the email and website with ellipses.  What do you think?
image

mjtravers:
> @mjtravers I think i resolved the email overlap thing. It was happening because email and website were both fields where the value is a single 'word'. Normally wordwrap doesn't break words. I added a css class to these two fields to break long words. I made a judgment call over truncating the email and website with ellipses. What do you think? image

That's a good way to handle it.

mjtravers (approved):
Change request have been made. Merging into devleop.

Issue #119, opened 2022-02-14
Allow user to enter Candidate Office/State/District info when creating Candidate Contact


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to enter Candidate Office/State/District info when creating Candidate Contact so that I can track Candidate information as required for transactions.

 Acceptance Criteria 
 

**If** I am a Committee Administrator
**When** I access my committee's contact list and am adding a new "Candidate" contact
**Then**  I will be required to input the following:
- CANDIDATE OFFICE (Dropdown: House, Senate, Presidential. Required)
- CANDIDATE STATE (A/N-2, _should be a dropdown, and required if Office = House or Senate_)
- CANDIDATE DISTRICT (A/N-2, required if Office = House. _List should reflect State selected_)

**And** the dropdowns appear as documented in the below xls file (tab 3 Congressional Districts):
https://docs.google.com/spreadsheets/d/17YmnaZgqvFOUyv4-Qknd8ikHe8o9LoDb/edit#gid=1465207356

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
Wireframe

![Candidate.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/fa81e131-a7ee-4494-8572-ef08f7861ebf)

![Candidate Office– Required Fields.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/908a876e-342a-4c5b-9344-e49fc0c18689)

AureliaKhorsand:
**3/16/22 Refinement Meeting:** Mitchell will confirm the dropdown lists with FEC, then load the lists to be validated to the ticket.

MitchellTCG:
@mjtravers 
Please make sure the candidate state dropdown should consist of the states and territories in row 1 of the excel doc. 

The Foreign Countries option should not appear in either candidates dropdowns — Address nor Office.

WiseQA:
Verified candidate office drop down for Presidential no Candidate State or Candidate District fields were displayed.
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/5c7fd835-12a5-464e-be36-851d3bfb1405)

Verified candidate office drop down for Senate ONLY Candidate State field is displayed.
[...]

PR #213, opened 2022-04-11
Add conditional display of candidate contact dropdowns

mjtravers:
#119 

Added conditional display of candidate state and candidate district based on value selected in candidate office dropdown.


lbeaufort (approved):
Looks good to me!

As an aside, for OP2: at some point when we do Schedule B we'll need to have the ability to enter non-federal candidates (for example, gubernatorial candidates), so office may not apply. Spoke w/ @AureliaKhorsand and we put in this placeholder ticket: https://github.com/fecgov/fecfile-web-app/issues/225

PR #214, opened 2022-04-11
set front-end fec api key

toddlees:
#87 
These are public keys that will be used on the front end api calls


lbeaufort (approved):
Per our discussion, I'll put in a follow-up ticket to move the committee info retrieval to the API so we can keep this call server-side

Issue #220, opened 2022-04-13
Correct app lint warnings and errors


mjtravers:
Follow on to #126, the adding of the lint check to CircleCI.

The task of this ticket is to clean up the current lint warnings and errors.

Technical Notes

*   Restore the full lint check that was limited in CircleCI to pass tests as a stop gap measure. The rules that were turned off can be removed from the .eslintrc.json file and are: no-extra-boolean-cast, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unused-vars

Acceptance Criteria

*   The app lint check reports 0 errors and warnings
*   No rule exceptions are declared in the .eslintrc.json file

QA

*   To verify that the lint check reported zero errors, expand the "Run linter" section of the CircleCI report [here]([https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/335/workflows/3de0a427-0371-45f2-afc8-050cd8e22916/jobs/1164](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/335/workflows/3de0a427-0371-45f2-afc8-050cd8e22916/jobs/1164))
*   To verify that there are no rule exceptions in the .eslintrc.json file, see [this file for before with the 'off' rules]([https://github.com/fecgov/fecfile-web-app/blob/9cd523ac4500e593c9d6eeff0a4b3d1ddfc8a035/front-end/.eslintrc.json#L27-L31](https://github.com/fecgov/fecfile-web-app/blob/9cd523ac4500e593c9d6eeff0a4b3d1ddfc8a035/front-end/.eslintrc.json#L27-L31)) and see [this file for after without 'off' rules]([https://github.com/fecgov/fecfile-web-app/blob/develop/front-end/.eslintrc.json#L27-L28](https://github.com/fecgov/fecfile-web-app/blob/develop/front-end/.eslintrc.json#L27-L28))

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1230](https://fecgov.atlassian.net/browse/FECFILE-1230)

WiseQA:
Verified Unit test passed as denoted in the acceptance criteria.
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/05876a94-8dfc-4d2a-92e7-ea0f7094e2a4)

Verified file before the "off rules" as denoted in the acceptance criteria.
[...]

PR #227, opened 2022-04-18
Fix lint warnings and errors

mjtravers:
#220 

All lint error and warning messages were clean up except for 8 errors regarding the rule: @typescript-eslint/no-explicit-any

Those 8 were mitigated using no-qa commenting "eslint-disable-line" where the any type is necessary for variables holding JSON data that could contain any JSON data structure.

An additional 10 errors of the no-explicit-any type were no-qa accepted but reside in the files session.service.ts and confirm-two-factor.component.ts which will be refactored out of the code base after the Login.gov service is integrated into the app


front-end/src/app/shared/components/table-list-base.component.ts
```diff
@@ -87,32 +88,37 @@ export abstract class TableListBaseComponent<T> {
...
-    //   icon: 'pi pi-exclamation-triangle',
-    //   accept: () => {
-    //     this.itemService.delete(item).subscribe((response: null) => {
-    //       this.item = this.getEmptyItem();
-    //       this.messageService.add({ severity: 'success', summary: 'Successful', detail: 'Item Deleted', life: 3000 });
-    //     });
-    //   },
-    // });
+    this.confirmationService.confirm({
+      message: 'Are you sure you want to delete this item?',
```
toddlees, line 92: Might be nice when we go through with ux to say what item is being deleted.  Like, "Are you sure you want to delete <item.name>"

even better, it seems like the standard now is soft deletes and undos

toddlees (approved):

Issue #192, opened 2022-03-30
Integrate validate npm package into INDIVIDUAL contacts ADD form validation screens


mjtravers:
Business Reason

As a user, I want the validation check on the contact add forms to use the validation javascript library so it validates against the validation schema.

Acceptance Criteria

*   Current validation code for the 4 contact add screens replaced by javascript validation npm package work as they did previously.
*   All 4 contact add screens (Individual, Organization, Committee, Candidate) pass the sprint-5 regression test

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1247](https://fecgov.atlassian.net/browse/FECFILE-1247)

WiseQA:
Individual Contact field messages screen shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/5277170e-2cbc-4765-ad30-db0a8adb0e7d)

Individual Contact field error messages screen shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/7add78aa-ec9e-4b1e-af45-a68f9ca35505)

Individual Contact "required fields" screen shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/65621499-e5dc-4369-8579-070b9f3040b2)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/d1f48d1a-c6e5-4745-852d-50777de28b42/download](https://app.zenhub.com/files/306118986/d1f48d1a-c6e5-4745-852d-50777de28b42/download)

QA review completed.  Moved to Stage Ready.

Issue #189, opened 2022-03-29
Integrate validate npm package into INDIVIDUAL contacts EDIT form validation screens


mjtravers:
Business Reason

As a user, I want the validation check on the contact edit forms to use the validation javascript library so it validates against the validation schema.

Acceptance Criteria

*   Current validation code for the 4 contact edit screens replaced by javascript validation npm package work as they did previously.
*   All 4 contact edit screens (Individual, Organization, Committee, Candidate) pass the sprint-6 regression tests.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1249](https://fecgov.atlassian.net/browse/FECFILE-1249)

AureliaKhorsand:
**Refinement Meeting 3/30/22:** @mjtravers to split out into a ticket for each contact type, and point each at 5.

WiseQA:
Contacts screen selecting Individual contact:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/68d594fa-dfa6-42e4-9c50-7107de63f5a3)

Successful - Contact Updated Popup Screen Shot:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/28afe654-0dd1-4c77-9b90-6c0c97d5fac2)

Contacts screen showing Individual edit name successful:
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/ac16577b-55f3-469f-8a57-9e442c143f9a)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/9c05e433-b64d-4420-b7d6-de0a1393f102/download](https://app.zenhub.com/files/306118986/9c05e433-b64d-4420-b7d6-de0a1393f102/download)

QA review completed.  Moved to Stage Ready.

PR #241, opened 2022-04-19
Integrate validate schema into contact forms

mjtravers:
#189 #192


front-end/src/app/contacts/contact-detail/contact-detail.component.ts
```diff
@@ -194,11 +200,22 @@ export class ContactDetailComponent implements OnInit {
...
-    return Contact.getFieldsByType(type).reduce(
-      (isInvalid: boolean, fieldName: string) => isInvalid || !!this.form?.get(fieldName)?.invalid,
-      false
-    );
+  /**
+   * Given the type of contact given, return the appropriate JSON schema doc
+   * @param {ContactTypes} type
+   * @returns {JsonSchema} schema
+   */
+  private getSchemaByType(type: ContactTypes): JsonSchema {
```
toddlees, line 208: could you make something like ContactTypeLabels where you map the types to thier schema and have it live in contact.model? then you wouldn't have to write all these conditionals in the component

toddlees (approved):

front-end/src/app/contacts/contact-detail/contact-detail.component.ts
mjtravers: @toddlees Per our chat, moving the schema lookup into the contact model makes a lot of sense at first glance but because of wanting to keep the schema loading in the contact module for lazy loading strategies, we'll revisit this if that bit of code needs to be duplicated down the road. Great talking that issue through to look at alternatives

PR #251, opened 2022-04-25
Change schema url from https to http in unit test

mjtravers:
The unit test was throwing an error caused by the https in the schema draft url


mjtravers:
Need to make this change on a different branch. Closing PR

Issue #165, opened 2022-03-15
Allow user to view reports


AureliaKhorsand:
Business Reason

As a Committee Administrator, I will be able to view a list of reports associated with my committee account so that I can manage reports as needed.

Technical Notes

*   The status and date/time fields will be placeholders until the data is added to the system
*   Create a Django fixture that creates a set of test reports for testing
*   All columns in the wireframe should be sortable. Sort coverage dates by ending coverage date.

Acceptance Criteria

***Given*** I am a Committee Administrator

  ***When*** I go to the reports management page for my committee account

  ***Then*** I will see a list of reports associated with my committee account and the following information for each:

*   ***Form type*** (e.g., Form 3X)
*   ***Type of report*** (e.g., April Monthly (M4))
*   ***Coverage dates***
*   ***Status*** - _Note: this is a placeholder for now_
*   ***Version*** all will be Original for now unless specifically created for testing purposes_
*   ***Date filed***

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1267](https://fecgov.atlassian.net/browse/FECFILE-1267)

MitchellTCG:
New wire in [fecfile-project-management#44](https://github.com/fecgov/fecfile-project-management/issues/44)

AureliaKhorsand:
![image.png](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/beba2dab-e476-4d6b-925b-0ff1d582f730)

MitchellTCG:
Wireframe to code

![Manage Reports – Form type – 1.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/d9f6e2d9-bb11-4af8-bc06-8b576b5e4eb3)

MitchellTCG:
@WiseQA 
I asked Matt to code to the wireframe. I will update the ticket. This is my bad. Sorry

WiseQA:
Verified list of "Manage reports" page contains the following and each column is sortable:
Form type (e.g., Form 3X)
Type of report (e.g., April Monthly (M4))
Coverage dates
Status - Note: this is a placeholder for now
Version all will be Original for now unless specifically created for testing purposes_
Date filed
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/57f44bf0-5818-4fd0-8cd9-3ec4f50de5fe)

[...]

PR #254, opened 2022-04-28
Add reports list page

mjtravers:
#165 

- Two known issues are being flagged by the ANDI accessibility tester in the list table: disabled buttons and color contrast. The disabled buttons will be addressed in a follow up ticket. The color contrast will be addressed when the UX updates the site color scheme.
-


front-end/src/app/reports/report-list/report-list.component.html
```diff
@@ -0,0 +1,79 @@
...
+    [showCurrentPageReport]="true"
+  >
+    <ng-template pTemplate="caption">
+      <div class="flex align-items-center justify-content-between">
+        <h5 class="m-0">Recent reports</h5>
+      </div>
+    </ng-template>
+    <ng-template pTemplate="header">
+      <tr role="row">
+        <th pSortableColumn="TBD" id="form-type-column" role="columnheader">
```
toddlees, line 38: can this be unset instead of an invalid value?

toddlees: or does that change the functionality?

front-end/src/app/shared/models/f3x-summary.model.ts
```diff
@@ -0,0 +1,247 @@
...
+  | F3xReportCodes.M6
+  | F3xReportCodes.M7
+  | F3xReportCodes.M8
+  | F3xReportCodes.M9
+  | F3xReportCodes.M10
+  | F3xReportCodes.M11
+  | F3xReportCodes.M12;
+
+// NOTE: If this label list is updated, please sync up the values in the api backend as well:
+// https://github.com/fecgov/fecfile-web-api/blob/develop/django-backend/fecfiler/f3x_summaries/views.py
```
toddlees, line 82: this could be more lift than it's worth, could we just have this mapping live in the f3x view and provide an endpoint to retrieve it?

front-end/src/app/shared/pipes/fec-date.pipe.ts
```diff
@@ -0,0 +1,12 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { DateTime } from 'luxon';
+
+@Pipe({
+  name: 'fecDate',
+})
+export class FecDatePipe implements PipeTransform {
+  transform(value: string): string {
+    const date = DateTime.fromFormat(value, 'yyyyMMdd');
```
toddlees, line 9: not something we'd want to deal with now but i would love to store dates as dates and worry about the fec `yyyymmdd` format only when we submit to fec.  storing data in mismatched types (e.g. dates in strings) when we don't really have to opens us up to a lot of pain

front-end/src/app/shared/models/f3x-summary.model.ts
toddlees: OR have it live here and pass the mapping in the list request. I just want to avoid the duplicate definition

mjtravers: The label values are stable enough that I'd rather not have an endpoint lookup. I did look into passing the alias field, report_code_label, back to the client as part of the result set but did not find a way to do that readily available. We can break that task out as a separate ticket as I should think it possible in the serializer somehow

front-end/src/app/shared/pipes/fec-date.pipe.ts
mjtravers: This could be done in the model fromJSON() method, converting the yyyymmdd format into a Date type.

front-end/src/app/reports/report-list/report-list.component.html
mjtravers: The primeng widget needs to have some sort of string identifier declared (not just an empty string) or else the sort button becomes disabled.

toddlees (approved):

Issue #261, opened 2022-05-02
Allow user to create a Form 3X QUARTERLY REPORT for a SPECIAL REPORTS: (Create Form 3X-Part 1)


MitchellTCG:
Business Reason 
 
As a Committee Administrator, I will be able to create a Form 3X QUARTERLY Special Report.

 Acceptance Criteria 
 
Given I am a Committee Administrator
When I create a new Form 3X quarterly Special Report 
Then I will be required to select a quarterly reporting frequency (stop-gap measure)
Then I will be required to select a report type with the following logic:

If Committee has a QUARTERLY Filing frequency:
**and** creates a report for an ELECTION YEAR:
•	12-Day Pre-Election Primary (12P)
•	12-Day Pre-Election Runoff (12R)
•	12-Day Pre-Election Convention (12C)
•	12-Day Pre-Election Special (12S)
•	30-Day Post-Election Runoff (30R)
•	30-Day Post-Election Special (30S)

If 12-DAY or 30-Day is selected:
•	Election on date picker appears.  And
•	In the state/province of dropdown appears

 Dev Note	
 
When a quarterly report type has been selected by clicking the correct radio button, the coverage dates section will appear per ticket (web-app #137)
 

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
![Custom Size – 29.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/dbd1fa97-24a8-4179-ac1a-66553358bee3)

![Custom Size – 38.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ba55e815-a4e5-4876-8e97-fdbfd06ad850)

![Custom Size – 41.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ec457a9c-84b1-4d83-9e70-274657339e0a)

WiseQA:
Verified capability to create a Form 3X Quarterly Report for Special reports by selecting the radio buttons denoted in Wireframe.
![Ticket #261 - Quarterly Report Special Report 12 and 30 day.JPG](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/87bbb048-a398-4b1d-b8b7-b55666f165bf)

[...]

Issue #260, opened 2022-05-02
Allow user to create a Form 3X QUARTERLY REPORT for a NON-ELECTION YEAR: (Create Form 3X-Part 1)


MitchellTCG:
Business Reason 
 
As a Committee Administrator, I will be able to create a Form 3X QUARTERLY Report for a non-election year.

 Acceptance Criteria 
 
Given I am a Committee Administrator
When I create a new Form 3X quarterly report for an election year
Then I will be required to select a quarterly reporting frequency (stop-gap measure)
Then I will be required to select a report type with the following logic:

If Committee has a QUARTERLY Filing frequency:
**and** creates a report for an ELECTION YEAR:
•	APRIL 15 QUARTERLY REPORT (Q1) 
•	JULY 31 MID-YEAR REPORT (NON-ELECTION YEAR ONLY) (MY)
•	OCTOBER 15 QUARTERLY REPORT(Q3) 
•	JANUARY 31 YEAR-END (YE)
•	TERMINATION REPORT (TER)

  Dev Note 
 	
When a quarterly report type has been selected by clicking the correct radio button, the coverage dates section will appear per ticket (web-app #137)
 
![Custom Size – 29.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ac9b8099-b504-4711-8683-9dfd0d9db4ce)

![Custom Size – 37.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/f74e8661-d342-47e3-bf84-31e4f62f615c)

![Custom Size – 40.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/6b6b9f59-da8a-4d45-a377-aede0c7ad4df)

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

WiseQA:
Verified capability to create a Form 3X QUARTERLY Report for a non-election year by selecting the radio buttons denoted in Wireframe.
![Ticket #260 - Quarterly Report Non Election Year.JPG](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/d6823bba-a273-488f-a4ca-cced8b240712)

Test Case/Test Script attached showing ticket “PASS”.
[https://app.zenhub.com/files/306118986/2c118049-fbad-4176-9e47-6ce5282b591e/download](https://app.zenhub.com/files/306118986/2c118049-fbad-4176-9e47-6ce5282b591e/download)

[...]

Issue #259, opened 2022-05-02
Allow user to create a Form 3X QUARTERLY REPORT for an ELECTION YEAR: (Create Form 3X-Part 1)


MitchellTCG:
Business Reason 
 
As a Committee Administrator, I will be able to create a Form 3X QUARTERLY Report for an election year.

 Acceptance Criteria 
 
Given I am a Committee Administrator
When I create a new Form 3X quarterly report for an election year
Then I will be required to select a quarterly reporting frequency (stop-gap measure)
Then I will be required to select a report type with the following logic:

If Committee has a QUARTERLY Filing frequency:
**and** creates a report for an ELECTION YEAR:
•	APRIL 15 QUARTERLY REPORT (Q1) 
•	JULY 15 QUARTERLY REPORT (Q2) 
•	OCTOBER 15 QUARTERLY REPORT(Q3) 
•	12-DAY PRE-ELECTION GENERAL (12G) 
•	30-DAY POST-ELECTION GENERAL (30G) 
•	JANUARY 31 YEAR-END (YE)
•	TERMINATION REPORT (TER)

If 12-DAY PRE-ELECTION GENERAL (12G) or 30-DAY POST-ELECTION GENERAL (30G) is selected:
•	Election on date picker appears.  And
•	In the state/province of dropdown appears

 Dev Note	
 
When a quarterly report type has been selected by clicking the correct radio button, the coverage dates section will appear per ticket (web-app #137)

![Custom Size – 29.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/44ed3753-2bba-4346-ab2f-abdc8ec9519f)

![Custom Size – 30.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/dcf6730f-5b5d-434b-9698-b9a4547c794e)

![Custom Size – 39.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/946d3518-609b-47ab-ad0a-6e63c4002455)

![Custom Size – 42.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/adbe877f-db98-4c5c-af12-8564c53706cf)

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

WiseQA:
Verified capability to create a Form 3X Quarterly Report for a Election Year by selecting the radio buttons denoted in Wireframe.
[...]

Issue #258, opened 2022-05-02
Allow user to create a Form 3X MONTHLY REPORT for a NON-ELECTION YEAR  (Create Form 3X-Part 1)


MitchellTCG:
Business Reason 
 
As a Committee Administrator, I will be able to create a Form 3X Monthly Report for an election year.

 Acceptance Criteria 
 
Given I am a Committee Administrator
When I create a new Form 3X Monthly report for a non-election year
Then I will be required to select a monthly reporting frequency (stop-gap measure)
Then I will be required to select a report type with the following logic:

If Committee has a MONTHLY Filing frequency:
**and** creates a report for a NON-ELECTION YEAR:
•	FEBRUARY 20 MONTHLY REPORT (M2)
•	MARCH 20 MONTHLY REPORT (M3)
•	APRIL 20 MONTHLY REPORT (M4)
•	MAY 20 MONTHLY REPORT (M5)
•	JUNE 20 MONTHLY REPORT (M6)
•	JULY 20 MONTHLY REPORT (M7)
•	AUGUST 20 MONTHLY REPORT (M8)
•	SEPTEMBER 20 MONTHLY REPORT (M9)
•	OCTOBER 20 MONTHLY REPORT (M10)
•	NOVEMBER 20 MONTHLY REPORT (M11)
•	DECEMBER 20 MONTHLY REPORT (M12)
•	JANUARY 31 YEAR-END (YE)
•	TERMINATION REPORT (TER)

 Dev Note	
 
When a monthly report type has been selected by clicking the correct radio button, the coverage dates section will appear per ticket (web-app #137)

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
![Custom Size – 31.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/f1a53748-fa27-4eb9-9fc2-97a33da0b7de)

![Custom Size – 33.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/34130430-d09d-4dcb-83c8-233b4d5ef963)

![Custom Size – 36.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/79f69aa4-e709-4fe6-8573-2d5f99dbea67)

WiseQA:
Verified capability to create a Form 3X Monthly Report for a non-election year by selecting the radio buttons denoted in Wireframe.
[...]

PR #263, opened 2022-05-03
Feature/135 reports detail

toddlees:
#257 #258 #259 #260 #261 #137


front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  userCanSetFilingFrequency: boolean = environment.userCanSetFilingFrequency;
+  stateOptions: PrimeOptions = [];
+  formSubmitted = false;
+
+  form: FormGroup = this.fb.group(this.validateService.getFormGroupFields(this.formProperties));
+
+  /**  This different label list is 'necessarry' because the labels in the wireframe are
+   * different between the list and the creation steps
+   * */
+  f3xReportCodeCreationLabels: LabelList = [
```
toddlees, line 47: @mjtravers this is failing the sonarcloud test for code duplication because honestly it is pretty close to duplicate code.  The labels in the reports list are unfortunately different from the ones in the creation screens

mjtravers: We'll need to investigate further a way to flag this as an exception or maybe increase the passing limit percentage for duplicate code.

mjtravers (approved):
Made some formatting changes and added a boolean type to a new Subject().
All else looks great. Passes CR

PR #266, opened 2022-05-05
use params instead of headers

toddlees:


front-end/src/app/shared/services/f3x-summary.service.ts
```diff
@@ -19,14 +19,14 @@ export class F3xSummaryService {
   public create(f3xSummary: F3xSummary, fieldsToValidate: string[] = []): Observable<F3xSummary> {
     const payload = f3xSummary.toJson();
     return this.apiService
-      .post<F3xSummary>(`/f3x-summaries/`, payload, { fields_to_validate: fieldsToValidate })
+      .post<F3xSummary>(`/f3x-summaries/?fields_to_validate=${fieldsToValidate}`, payload)
       .pipe(map((response) => F3xSummary.fromJSON(response)));
   }
```
mjtravers, line 24: I think we should move the query params implementation from F3xSummariesService and into the ApiService. Future forms and maybe transactions will also be using it.

Also, the HttpClient post() and put() methods can accept query params in the method Options parameter: https://angular.io/api/common/http/HttpClient#post rather than embedding directly in the URL string

We can replace the headersToAdd with a queryParams object in ApiService.post() and .put()

mjtravers (changes requested):

front-end/src/app/shared/services/f3x-summary.service.ts
toddlees: I like that!

mjtravers (approved):

Issue #215, opened 2022-04-11
System to auto-populate Committee Address on Form 3X and inform user to notify FEC within 10 days of Address Change


AureliaKhorsand:
Business Reason

The system will auto-populate Committee address on a Form 3X report so that the user does not have to manually enter information that rarely changes, reducing the chance for user error and enhancing the user experience. As a user, I will be able to edit the address and mark it as a change if needed so that FEC is notified of an address change upon report submission.

Acceptance Criteria

***When*** a Form 3X report gets created

  ***Then*** the Committee Address will auto-populate on the Form 3X report.

***Given*** I am a Committee Administrator or Backup Committee Administrator accessing an open report

  ***Then*** I will see information letting me know that I must notify FEC within 10 days of an address change and they will not see this change until I submit my Form 3X report, and a link to the webform for Form 1 will be available and will open in a new tab. The link to Form 1 is [https://webforms.fec.gov/webforms/form1/index.htm](https://webforms.fec.gov/webforms/form1/index.htm)

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1234](https://fecgov.atlassian.net/browse/FECFILE-1234)

MitchellTCG:
Mocks in fecfile-project-management#42

AureliaKhorsand:
@MitchellTCG We need to update the message on the wireframe to say the following, based on the 4/18/22 requirements meeting: 

Important notice: 
Within 10 days of an address change you must notify the FEC by amending your Form 1, in addition to updating your address here. 
Update your committee address (link).

WiseQA:
Needs more work - moved back to In-Progress

AureliaKhorsand:
@MitchellTCG @toddlees isn't this the ticket that was getting bumped back? Or is this where it was going? Is the only blocker getting the correct text?

WiseQA:
Changed title of ticket as to what the Acceptance Criteria is and what is actually being tested.  Address change was covered in Ticket 133.  This ticket about 10 days to notify FEC about address change.

WiseQA:
“Within 10 days of an address change you must notify the FEC by amending your Form 1, in addition to updating your address here.” Update your committee address page is displayed.
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/c52abf57-380e-4434-99bf-56901118b137)

Disclaimer for 10 day notice page is displayed:
[...]

PR #268, opened 2022-05-06
Link to form 1 in f3x step 2

toddlees:
#215

I need someone to provide the form 1 link.

I'd imagine that is this: https://webforms.fec.gov/webforms/form1/index.htm but i'm not sure


lbeaufort:
@toddlees that's the correct link to the Form 1 webform.

mjtravers (approved):

Issue #230, opened 2022-04-19
Allow users to enter Group B Transaction


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to enter a transaction for **"Offsets to Operating Expenditures"** on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 

**Given** I am a Committee Administrator editing an open Form 3X report
**Then** I can enter a transaction with the type "Offsets to Operating Expenditures" from the Transaction Screen (for Individual and Organization Only)
**Then** I will be required to enter the following information (unless otherwise noted):

# Contributor 
#
-	**CONTRIBUTOR TYPE** 
Field description in API: ENTITY TYPE
Data field: Dropdown
Dropdown options: Individual, Organization
---
_if Entity = Organization_
-	**ORGANIZATION NAME** (A/N-200, Required)
Field description in API: CONTRIBUTOR ORGANIZATION NAME
Data field: textbox
---
_if Entity = Individual_
-	**LAST NAME** (A/N-30, Required)
Field description in API: CONTRIBUTOR LAST NAME
Data field: textbox
-	**FIRST NAME** (A/N-20, Required)
Field description in API: CONTRIBUTOR FIRST NAME
Data field: textbox
-	**MIDDLE NAME** (A/N-20, Optional)
Field description in API: CONTRIBUTOR MIDDLE NAME
Data field: textbox
-	**PREFIX** (A/N-10, Optional)
Field description in API: CONTRIBUTOR PREFIX
Data field: textbox
-	**SUFFIX** (A/N-10, Optional)
Field description in API: CONTRIBUTOR SUFFIX
Data field: textbox
---

# Address 
#
-	**STREET ADDRESS** (A/N-34, Required)
Field description in API: CONTRIBUTOR STREET 1
Data field: textbox
-	**APARTMENT, SUITE, ETC.** (A/N-34, Optional)
Field description in API: CONTRIBUTOR STREET 2
Data field: textbox
-	**CITY** (A/N-30, Required)
Field description in API: CONTRIBUTOR CITY
Data field: textbox
-	**STATE/PROVINCE** (ALPHA-2, Required)
Field description in API: CONTRIBUTOR STATE
Data field: Dropdown
          Dropdown options: States without military
          https://docs.google.com/spreadsheets/d/17YmnaZgqvFOUyv4-Qknd8ikHe8o9LoDb/edit#gid=494776297
[...]

PR #270, opened 2022-05-09
Add group B transaction page

mjtravers:
#230

### DIFF
--- a/front-end/src/app/shared/services/validate.service.spec.ts
+++ b/front-end/src/app/shared/services/validate.service.spec.ts
@@ -48,7 +48,7 @@ describe('ValidateService', () => {
     expect(result).toBe(null);
   });
 
-  it('#formValidator should validate enum properties correctly', () => {
+  it('#formValidator should validate boolean properties correctly', () => {
     const fb: FormBuilder = new FormBuilder();
     service.formValidatorSchema = f3xSchema;
     service.formValidatorForm = fb.group(service.getFormGroupFields(service.getSchemaProperties(f3xSchema)));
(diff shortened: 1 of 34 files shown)


front-end/src/app/shared/services/validate.service.spec.ts
```diff
@@ -60,14 +60,19 @@ describe('ValidateService', () => {
     let result: ValidationErrors | null = validator(service.formValidatorForm.get('change_of_address') as FormControl);
     expect(result).not.toBe(null);
     if (result) {
-      expect(result['pattern'].requiredPattern).toBe('must be boolean,null');
+      expect(result['pattern'].requiredPattern).toBe('Allowed values: X, ');
```
toddlees: So even thought we merged the updates into the validate repo, we want to point to the old schemas?

toddlees (changes requested):

front-end/src/app/transactions/transaction-group-b/transaction-group-b.component.ts
```diff
@@ -1,8 +1,158 @@
...
+  save(navigateTo: 'list' | 'add another') {
+    this.formSubmitted = true;
+
+    if (this.form.invalid) {
+      return;
+    }
+
+    // Force selection of checkbox to true/false even if never checked
+    this.form.patchValue({
+      memo_code: !!this.form.get('memo_code')?.value,
```
toddlees: do we get the same functionality if we set the memo_code to false by default in the template?

front-end/src/app/shared/services/validate.service.spec.ts
toddlees: if we point to a later commit in fec-validate's history (HEAD of develop for example) this will break

mjtravers: Yes, good catch. My local validate repo got behind. I updated both the fecfile-web-app and fecfile-web-api repos to point back to the develop branch.

front-end/src/app/transactions/transaction-group-b/transaction-group-b.component.ts
mjtravers: For the p-checkbox that is set to binary mode ([binary]="true") there is no attribute to set in the template to make the default  value "false", it is returning an empty string. I updated the form initialization to patchValue a "false" value and moved that force patch value out of the save() method.

toddlees (approved):

Issue #228, opened 2022-04-19
Create Schedule A Transaction Type selection screen


AureliaKhorsand:
***Note:*** _These links/items will be inactive until we develop transactions, and they will become active as we go._

  

Business Reason

As a Committee Administrator within an existing Form 3X report, I will be able to see a list of available Schedule A transactions grouped by type so that in the future I can easily see and select what I need to enter.

Acceptance Criteria

***Given*** I am editing an open Form 3X report

  ***When*** I go to enter a Schedule A transaction

  ***Then*** available Schedule A transactions will be grouped by the following:

—

1.  Contributions from Individuals/Persons

       #

*   Individual Receipt
*   Tribal Receipt
*   Partnership Receipt
*   Reattribution
*   In-Kind Receipt
*   Returned/Bounced Receipt (Individual)
*   Earmark Receipt
*   Conduit Earmark (Deposited)
*   Conduit Earmark (Undeposited)
*   Unregistered Receipt from Person
*   Unregistered Receipt from Person - Returned/Bounced Receipt

       —

1.  Contributions from Registered Filers

       #

*   Party Receipt
*   Party In-Kind
*   Returned/Bounced Receipt (Party)
*   PAC Receipt
*   PAC In-Kind
*   PAC Earmark Receipt
*   PAC Conduit Earmark (Deposited)
*   PAC Conduit Earmark (Undeposited)
*   Returned/Bounced Receipt (PAC)

       —

1.  Transfers

       #

*   Transfers
*   JF Transfers
*   In-Kind Transfer
*   In-Kind Transfer-FEA
*   JF Transfer - National Party Recount Account
*   JF Transfer - National Party Convention Account
*   JF Transfer - National Party Headquarters Account

       —

1.  Refunds

       #

*   Refunds of Contributions to Registered Committees
*   Refunds of Contributions to Unregistered Committees

       —

1.  Other

       #

*   Offsets to Operating Expenditures
*   Other Receipts
*   Ind. Receipt - Non-Contribution Account
*   Other Committee Receipt - Non-Contribution Account
*   Business/Labor Org. Receipt - Non-Contribution Account
*   Individual Recount Receipt
*   Party Recount Receipt
*   PAC Recount Receipt
*   Tribal Recount Receipt
*   Individual National Party Recount Account
*   Party National Party Recount Account
*   PAC National Party Recount Account
*   Tribal National Party Recount Account
*   Individual National Party Headquarters Buildings Account
[...]

PR #274, opened 2022-05-11
Adds Schedule A Transaction Selection

toddlees:
Issue #228 

@mjtravers I tried andi on the page and it alerted me to the use of aria-hidden.  Based on this [article](http://web-accessibility.carnegiemuseums.org/code/accordions/) I believe primeng is using aria-hidden appropriately and the alert can be ignored


front-end/src/app/transactions/transaction-type-picker/transaction-type-picker.component.ts
```diff
@@ -0,0 +1,94 @@
...
+import { LabelList } from 'app/shared/utils/label.utils';
+
+@Component({
+  selector: 'app-transaction-type-picker',
+  templateUrl: './transaction-type-picker.component.html',
+  styleUrls: ['./transaction-type-picker.component.scss'],
+})
+export class TransactionTypePickerComponent implements OnInit {
+  scheduleATransactionTypeLabels: LabelList = ScheduleATransactionTypeLabels;
+  report: F3xSummary | undefined;
```
mjtravers: I think we can change this F3xSummary type to the Report interface type. Is this component only for F3X transactions or will their be transactions for other form types down the road? If not, we can leave it as F3xSummary

mjtravers (changes requested):
Per our discussion, please rename the component so it specifically references being an F3X-centric component and the 4 code smells. Thx

mjtravers (changes requested):
I am seeing errors in the unit test output. They don't stop the testing but it should be sorted out before merging to develop. Here's a screen shot:
cr-app228-unit-test-errors

toddlees:
image
@mjtravers addressed the messages coming up during the unit tests.  adding routertestingmodule silenced them

mjtravers (approved):

PR #280, opened 2022-05-16
Feature/328 django login-dot-gov django login dot gov

dheitzer:


Elaine-Krauss-TCG (approved):
The code looks great :+1:

mjtravers (changes requested):
One change please that I didn't think to mention earlier on code organization: Would you move the login-page.guard.ts into a newly created directory: src/app/shared/guards ?

front-end/src/app/login/guards/login-page.guard.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LoginGuard implements CanActivate {
+  constructor(private apiService: ApiService, private router: Router) {}
+  canActivate(): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    if (this.apiService.isAuthenticated()) {
+      this.router.navigate(['dashboard']);
```
toddlees: Navigate is [asynchronous](https://github.com/angular/angular/blob/88e16553e88356339db4fbc6aa1b78a9dfdfc8d2/packages/router/src/router.ts#L1215), could you add a return here so that in the case where navigate errors, the user isn't passed through the guard

front-end/src/environments/environment.cloud.gov.dev.ts
```diff
@@ -5,12 +5,17 @@
 export const environment = {
   production: false,
   name: 'development',
-  apiUrl: 'https://fecfile-web-api-dev.app.cloud.gov/api/v1',
+  apiUrl: 'https://fecfile-web-api-login-dev.app.cloud.gov/api/v1',
```
toddlees: this needs to be updated before merging into dev

tasks.py
```diff
@@ -7,7 +7,7 @@
 from invoke import task
 
 
-APP_NAME = "fecfile-web-app"
+APP_NAME = "fecfile-web-app-login"
```
toddlees: needs to be update before merging into dev

toddlees (changes requested):

front-end/src/environments/environment.cloud.gov.dev.ts
dheitzer: updated in the latest version

dheitzer:
> 

This was addressed

front-end/src/environments/environment.cloud.gov.dev.ts
```diff
@@ -11,6 +11,11 @@ export const environment = {
   fecApiUrl: 'https://api.open.fec.gov/v1/',
   fecApiKey: 'DVoLzo07NBfrbDZPj0LJs3PS0GIRL3fk4eOp7Zo6',
   userCanSetFilingFrequency: true,
+  loginDotGovAuthUrl: 'https://fecfile-web-api-dev.app.cloud.gov/oidc/authenticate',
+  loginDotGovLogoutUrl: 'https://fecfile-web-api-dev.app.cloud.gov/oidc/logout/',
```
mjtravers, line 15: Looks like we need the equivalent loginDotGov environment variables set in the stage and prod env files as well.

mjtravers (changes requested):
See inline comment about environment files for stage and prod needing the loginDotGov env vars as well.

front-end/src/environments/environment.cloud.gov.dev.ts
dheitzer: I went ahead and added these in the latest commit.

mjtravers (approved):

toddlees (approved):

Issue #130, opened 2022-02-17
System to direct user to fec.gov (Form 1) to edit Committee Account data


AureliaKhorsand:
Business Reason

As a Committee Administrator, I will see instructions on the Committee Account page telling me to access fec.gov and resubmit a Form 1 to update any Committee Account information so that I can edit the information as needed if it changes.

Acceptance Criteria

***Given*** a Committee Administrator is on the Committee Account page (Profile > Account)

  ***Then*** I will see instructions directing me to resubmit Form 1 through FEC if edits are needed 

  ***And*** a link to my committee's most recently submitted Form 1

  ***And*** each link will open a new tab when clicked.

### QA Notes ###

_Note: Remove numerous "Edit" buttons._

For example: “You can view your committee's most recent Form 1 here: (link to most recent F1 on fec.gov for that committee). You can update your committee’s official information disclosed on Form 1 here. (Link:  https://webforms.fec.gov/webforms/form1/index.htm ).”

Dev Notes: 

- For the most recent Form 1 for a committee, pull the "pdf_url" from the following API call FOR EXAMPLE [https://api.open.fec.gov/v1/filings/?api_key=DEMO_KEY&sort=-receipt_date&per_page=1&page=1&cycle=2022&committee_id=C00344234&form_type=F1](https://api.open.fec.gov/v1/filings/?api_key=DEMO_KEY&sort=-receipt_date&per_page=1&page=1&cycle=2022&committee_id=C00344234&form_type=F1) 
- (LB Note): This endpoint (/filings/) is updated nightly. In order to get the most up-to-date data, we should also include a query to the raw/efile endpoints. These endpoints don't have as good filters though, so we'll need to loop through some of it on our own. Start with https://api.open.fec.gov/v1/efile/filings/?api_key=DEMO_KEY&committee_id=C00344234 and then get the most recent filing where `form_type` starts with F1. We can probably wait until we move this to the API. We can't just use efiling endpoint because it's only 3-4 months of data
- Need to create three FEC API keys, and also have public API keys and rate limit them by IP.
- Need to rename the existing API key to private

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1289](https://fecgov.atlassian.net/browse/FECFILE-1289)

AureliaKhorsand:
[...]

PR #301, opened 2022-05-24
Committee Detail Links

toddlees:
Issue: #130 

On the Committee Details page, link buttons at the bottom will direct the user to their most recent F1 filing PDF or to the site where they can file a new F1.

To get the correct F1 pdf link, we check the realtime efile/filings/ endpoint.  If a recent F1 is not found in that response, we check the filings/ endpoint which is updated nightly but retains older reports.


Elaine-Krauss-TCG (approved):

lbeaufort (approved):
Looks great, thanks @toddlees!

mjtravers (approved):

PR #331, opened 2022-06-10
Feature/123 refactor sorting of report code column

Elaine-Krauss-TCG:
Implements the front-end elements of [API ticket 123](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/fecgov/fecfile-web-api/123)

Paired with [API feature/123](https://github.com/fecgov/fecfile-web-api/tree/feature/123-Refactor-Sorting-of-Report-Code-Column)


front-end/src/app/shared/pipes/report-code-label-list.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { ReportCodeLabelList, ReportCodeLabel } from '../utils/reportCodeLabels.utils';
+
+@Pipe({ name: 'findOnReportCodePipe' })
+export class FindOnReportCodePipe implements PipeTransform {
+  transform(list: ReportCodeLabelList | null, reportCode: string) {
+    let label: ReportCodeLabel = { label: '', report_code: '' };
+    let found: ReportCodeLabel | undefined = undefined;
+    if (list != null) {
```
toddlees: could you combine these two if statements into `if (list?.length > 0)`.  The code is generally easier to understand if there are less conditional scopes

front-end/src/app/shared/pipes/report-code-label-list.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { ReportCodeLabelList, ReportCodeLabel } from '../utils/reportCodeLabels.utils';
+
+@Pipe({ name: 'findOnReportCodePipe' })
+export class FindOnReportCodePipe implements PipeTransform {
+  transform(list: ReportCodeLabelList | null, reportCode: string) {
+    let label: ReportCodeLabel = { label: '', report_code: '' };
```
toddlees: You don't need to make this change, but it could be helpful to the coder using the pipe for the pipe to return `undefined` if no report code is found.  Without having to understand the nature of the pipe, they can simply check if anything came back.  By returning an object with empty string label and code, conventionally checking if the pipe returned a defined value would give a false positive.  consider the find() function you're using below.  It return `undefined` if nothing was found.

Like i said, not a necessary change by any means, but worth considering.

front-end/src/app/shared/pipes/report-code-label-list.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { ReportCodeLabelList, ReportCodeLabel } from '../utils/reportCodeLabels.utils';
+
+@Pipe({ name: 'findOnReportCodePipe' })
+export class FindOnReportCodePipe implements PipeTransform {
```
toddlees, line 5: This is a really useful pipe!

toddlees (changes requested):
Looks good!  I would just reduce the verbosity of the conditionals in the label pipe

toddlees (approved):

Issue #201, opened 2022-04-01
Allow users to view existing users


AureliaKhorsand:
Business Reason

As a Committee Administrator, I will be able to view existing users in my Committee Account so that I can manage them appropriately.

UPDATED Acceptance Criteria

***Given*** I am a Committee Administrator

  ***When*** I access the user management page

  ***Then*** I will able to see existing users for the Committee Account, including:

*   Name
*   Email
*   Role
*   Status

***And*** the list will be sorted alphabetically (A-Z) by name in ascending order.

Dev Note

Please make the Name column sortable by ***Last Name***, First Name

_Note: Additional functionality will be later features._

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1242](https://fecgov.atlassian.net/browse/FECFILE-1242)

AureliaKhorsand:
@MitchellTCG @mary-tcg This is the story that will eventually need the Manage Users wireframe you all have been working on.

AureliaKhorsand:
![image.png](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/3d185fdd-fbdc-4bff-a06d-2c5cbf2851d9)

AureliaKhorsand:
**4/6/22 Refinement Meeting:** This ticket covers functionality and not specific look-and-feel. Clarified that users should only see users from the Committee Account they are logged in to and not others.

Elaine-Krauss-TCG:
To-do list:
- [x] Implement an endpoint that allows the user to retrieve a list of users in the same committee
- [x] Implement a new front-end component per the wireframe
- [x] Add unit tests for code coverage

Elaine-Krauss-TCG:
Pull requests open for this ticket on the [Front-End](https://github.com/fecgov/fecfile-web-app/pull/337) and the [Back-End](https://github.com/fecgov/fecfile-web-api/pull/143)

For QA: You can find the results of the unit tests [here](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-api/482/workflows/4b007b0c-76d5-48c4-9d5a-37cc3a19cabe/jobs/1342)

mjtravers:
[...]

PR #337, opened 2022-06-15
Feature/201 allow users to view existing users

Elaine-Krauss-TCG:
#201

Creates a page for managing users which lists the users belonging to the same committee as the logged in user, sort-able on name, email, and role.  Requires the endpoint created in [this](https://github.com/fecgov/fecfile-web-api/pull/143) PR.


front-end/src/app/shared/models/user.model.ts
```diff
@@ -12,6 +13,30 @@ export type UserLoginData = {
...
+  role: string | null = null;
+  is_active: boolean | null = null;
+
+  // prettier-ignore
+  static fromJSON(json: any): CommitteeUser { // eslint-disable-line @typescript-eslint/no-explicit-any
+    return plainToClass(CommitteeUser, json);
+  }
+
+  public getRoleLabel(): string | null {
+    const role: string | null = this.role;
```
toddlees: RoleLabels is a LabelList which has a static function `get`.
We could use it like this:
```
public getRoleLabel(): string | null {
  return LabelList.get(RoleLabels, this.role);
}
```

front-end/src/app/shared/services/users.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    return this.apiService.get<ListRestResponse>(`/committee/users/${parameter_string}`).pipe(
+      map((response: ListRestResponse) => {
+        response.results = response.results.map((item) => CommitteeUser.fromJSON(item));
+        return response;
+      })
+    );
+  }
+
+  public create(user: CommitteeUser): Observable<CommitteeUser> {
+    console.log('Create', user);
```
toddlees, line 29: We'll want to remove console statements.

front-end/src/app/users/user-list/user-list.component.html
```diff
@@ -0,0 +1,61 @@
...
+      <tr role="row">
+        <th pSortableColumn="name" id="name-column" role="columnheader">Name <p-sortIcon field="name"></p-sortIcon></th>
+        <th pSortableColumn="email" id="email-column" role="columnheader">Email <p-sortIcon field="email"></p-sortIcon></th>
+        <th pSortableColumn="role" id="role-column" role="columnheader">Role <p-sortIcon field="role"></p-sortIcon></th>
+        <th pSortableColumn="is_active" id="active-column" role="columnheader">Status <p-sortIcon field="is_active"></p-sortIcon></th>
+      </tr>
+    </ng-template>
+    <ng-template pTemplate="body" let-item>
+      <tr role="row">
+        <td>{{ item.last_name }}, {{ item.first_name }}</td>
```
toddlees, line 52: I noticed that you added a `name` alias to the django view, which combines the first and last names into a third field.  We should stick to one way of combining the name fields.  I'm on the fence over whether to do it on the front end or on the backend.  Which do you think is better?

front-end/src/app/users/users.module.ts
```diff
@@ -0,0 +1,43 @@
...
+import { TableModule } from 'primeng/table';
+import { DropdownModule } from 'primeng/dropdown';
+import { ToastModule } from 'primeng/toast';
+import { ToolbarModule } from 'primeng/toolbar';
+import { FileUploadModule } from 'primeng/fileupload';
+import { ButtonModule } from 'primeng/button';
+import { CheckboxModule } from 'primeng/checkbox';
+import { ConfirmDialogModule } from 'primeng/confirmdialog';
+import { DialogModule } from 'primeng/dialog';
+import { InputTextModule } from 'primeng/inputtext';
```
mjtravers: A number of these PrimeNG modules, such as FileUploadModule, are not used by the Users module and can be removed. You will know it isn't used if you remove the import line and import declaration below and the app recompiles without error

mjtravers (dismissed):
See the inline comments for requested changes.

front-end/src/app/users/user-list/user-list.component.html
Elaine-Krauss-TCG: In an outside conversation, we've agreed that this is not an issue as the `name` annotation from the back-end is only used for sorting and does not get passed by the serializer to the front-end.

front-end/src/app/users/users.module.ts
Elaine-Krauss-TCG: I've removed all of the PrimeNG imports and added them back one-by-one as was necessary to get the code to compile.  There should be no more unnecessary imports present.

front-end/src/app/shared/models/user.model.ts
toddlees: looks like that opens up a tiny can of worms.  for one it would be `LabelUtils.get()` instead of `LabelList.get`.  Two, you'd need to change the arguments of `get` to `get(labelArrays: LabelList, key: string | null)`.  Third now the function returns `''` instead of `null`.  I still stand by my suggestion in a previous pr on returning null on a miss ([reason](https://stackoverflow.com/a/1626620/18957717)), but in this scenario `''` seems rational enough since we're just going to throw the result into the html.

Seems like a lot of work but it reduces code duplication/divergence and you'd be improving the LabelUtil.get which i'm positive we will be using many more times in the app

toddlees: looks like that opens up a tiny can of worms. for one it would be LabelUtils.get() instead of LabelList.get. Two, you'd need to change the arguments of get to get(labelArrays: LabelList, key: string | null). Third now the function returns '' instead of null. I still stand by my suggestion in a previous pr on returning null on a miss ([reason](https://stackoverflow.com/a/1626620/18957717)), but in this scenario '' seems rational enough since ~~we're just going to throw the result into the html~~ the function is "what should i call this" not "give me this object so i can work with it".

Seems like a lot of work but it reduces code duplication/divergence and you'd be improving the LabelUtil.get which i'm positive we will be using many more times in the app

toddlees (changes requested):

front-end/src/app/shared/models/user.model.ts
Elaine-Krauss-TCG: It does reduce code duplication, so it's worth doing.  Plus, the only thing relying on the method returning null is the unit tests, so those can be changed.

The User model still has the `.getRoleLabel()` method, but it's just a wrapper that returns null if the object's `.role` attribute is set to null and otherwise returns the result from `LabelUtils.get()`

Elaine-Krauss-TCG: And after further consideration, the null-handling is now handled by changes to `LabelUtils.get()` in line with Todd's suggestions

toddlees (approved):

PR #348, opened 2022-06-17
Enable Committee Entity Type For Group B

toddlees:
Adds Committee entity type to group b dropdown and labels the organization_name input "COMMITTEE NAME" when the entity type is committee.  Both organization and committee use the organization_name field in the fec spec

### DIFF
--- a/front-end/src/app/transactions/transaction-group-b/transaction-group-b.component.html
+++ b/front-end/src/app/transactions/transaction-group-b/transaction-group-b.component.html
@@ -18,11 +18,15 @@
         </div>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


Elaine-Krauss-TCG (approved):
Took me a hot second to figure out what was going on with `form.get('entity_type')?.value === ContactTypes.ORGANIZATION ? 'ORGANIZATION NAME' : 'COMMITTEE NAME'`

Under normal circumstances I'd want to see it simplified, but this genuinely is the best solution for doing this inside of a template

dheitzer (approved):

PR #358, opened 2022-06-24
Feature/233 Add Group E transaction type page

dheitzer:


Elaine-Krauss-TCG (approved):
Looks good all around

front-end/src/app/transactions/transaction-group-e/transaction-group-e.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    protected override validateService: ValidateService,
+    protected override fb: FormBuilder,
+    protected override router: Router
+  ) {
+    super(messageService, transactionService, validateService, fb, router);
+  }
+
+  override ngOnInit(): void {
+    super.ngOnInit();
+    this.form.get('entity_type')?.setValue(this.contactTypeOptions[0].code);
```
dheitzer: Just wanted to point this out - I had to add this because  p-dropdown displays the first item in the list (autoDisplayFirst defaults to true https://www.primefaces.org/primeng/dropdown) however it doesn't set it in the form.  Therefore, I was getting "IND" when I saved (instead of "COM" since this is the first in the list).  I think we'll need to do add this for the other transactions so we may want to consider adding it to the base class?

toddlees: Isn't the only option COM based on line 33?  If primeng is picking "IND" that means we're providing it the wrong list of options somehow right?

mjtravers (changes requested):
All looks good except I am not getting the red required error message for Date Received when saving and the calendar input is blank. See here:
image

front-end/src/app/transactions/transaction-group-e/transaction-group-e.component.ts
mjtravers: I traced this and the problem is in the resetForm() method in the transaction-type-base.component.ts. A better solution is to create a defaultEntityType property to the base component class and have it default to ContactTypes.INDIVIDUAL. Subsequent iterations of the form can set the defaultEntityType to values other than ContactTypes.INDIVIDUAL. The line at 150 in transaction-type-base.components.ts can replace ContactTypes.INDIVIDUAL with this.defaultEntityType and the override of ngOnInit can be removed from transaction-group-e-component.ts

toddlees: @mjtravers if we don't care about which option of the narrowed-down options comes first, we could just replace line 150 with @dheitzer 's code: `entity_type: this.contactTypeOptions[0].code` since that variable is available in the parent class.  We'd lose the control to pick a different option as default though

toddlees (approved):

mjtravers (approved):
The required warning for the date field is not working but looks to be broken in develop branch now, too. Passing ticket for CR and the date required warning will be examined/fixed in a follow on ticket.

PR #367, opened 2022-07-01
Feature/262 overlapping coverage dates

dheitzer:
I'm wondering - do we need to check on the backend for valid coverage dates as well? (in case someone tries to POST through something other than the UI).


toddlees (approved):
Adding unit-test coverage after this pr

PR #369, opened 2022-07-01
Feature/125 limit transactions table to include from only one report

Elaine-Krauss-TCG:
Associated with App Ticket 125

Paired with this [api branch](https://github.com/fecgov/fecfile-web-api/tree/feature/125-Limit-transactions-table-to-only-include-from-one-report)


front-end/src/app/shared/components/table-list-base/table-list-base.component.ts
```diff
@@ -80,16 +80,17 @@ export abstract class TableListBaseComponent<T> implements OnInit, AfterViewInit
...
     let ordering: string = event.sortField ? event.sortField : '';
     if (ordering && event.sortOrder === -1) {
       ordering = `-${ordering}`;
     }
 
-    this.itemService.getTableData(pageNumber, ordering).subscribe((response: ListRestResponse) => {
+    this.itemService.getTableData(pageNumber, ordering, params).subscribe((response: ListRestResponse) => {
       this.items = [...response.results];
-      this.totalItems = response.count;
+      this.totalItems = this.items.length;
```
mjtravers: Will this give the total number of records for the query and not just the number of records for the page?

front-end/src/app/shared/components/table-list-base/table-list-base.component.ts
```diff
@@ -167,4 +168,9 @@ export abstract class TableListBaseComponent<T> implements OnInit, AfterViewInit
       this.messageService.add({ severity: 'success', summary: 'Successful', detail: 'Items Deleted', life: 3000 });
     });
   }
+
+  public getGetParams(): { [param: string]: string | number | boolean | ReadonlyArray<string | number | boolean> } {
```
mjtravers, line 182: It's not obvious what why this method is here, please add a comment using [JSDocs](https://devhints.io/jsdoc). See OnSelectAllChange() method above for an example

front-end/src/app/reports/f3x/create-workflow/create-f3x-step3.component.ts
```diff
@@ -31,6 +31,12 @@ export class CreateF3xStep3Component extends TableListBaseComponent<Transaction>
   protected getEmptyItem(): Transaction {
     return {} as Transaction;
   }
+
+  override getGetParams(): { [param: string]: string | number | boolean | ReadonlyArray<string | number | boolean> } {
+    const reportId = this.activatedRoute.snapshot.params['reportId'];
+    const params = { report_id: reportId };
```
mjtravers: What happens if the url does not have a value for :reportId? Does an error get thrown?

mjtravers (changes requested):
Some inline questions added. The significant one is the pagination total count change.

front-end/src/app/shared/components/table-list-base/table-list-base.component.ts
Elaine-Krauss-TCG: That's a very good question...

After a little digging and some experimentation with Postman, it does appear that the .count attribute is intended to reflect the total number of records _before_ pagination

front-end/src/app/reports/f3x/create-workflow/create-f3x-step3.component.ts
Elaine-Krauss-TCG: The routing requires a value for :reportId.  If no value is provided, it redirects you to the Manage Reports page.  If the value is a string, then it evaluates to "api/v1/f3x-summaries/NaN/" and you get a 404 error.

front-end/src/app/shared/components/table-list-base/table-list-base.component.ts
Elaine-Krauss-TCG: Reverted to the prior implementation, `this.totalItems = response.count`

front-end/src/app/shared/components/table-list-base/table-list-base.component.ts
Elaine-Krauss-TCG: Commenting added per the request.  It's definitely a little obtuse, so this was a necessary change.

mjtravers (approved):

PR #374, opened 2022-07-06
Update routing from transaction entry page

mjtravers:
Changed routing when leaving transaction entry page to transaction list from report list.


dheitzer (commented):
LGTM

PR #377, opened 2022-07-07
Feature/46 updates to schema specs

mjtravers:
Updates to the schema files based on information collected by requirements.

[fecfile-project-management#46](https://github.com/fecgov/fecfile-project-management/issues/46)

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
@@ -93,7 +93,7 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy 
     });
 
     if (this.transaction?.transaction_type_identifier) {
-      let fieldsToValidate: string[] = this.getFieldsToValidate();
+      let fieldsToValidate: string[] = this.validateService.getSchemaProperties(this.schema);
       // Remove properties populated in the back-end from list of properties to validate
       fieldsToValidate = fieldsToValidate.filter((p) => p !== 'transaction_id' && p !== 'donor_committee_name');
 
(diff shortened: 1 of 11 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -93,10 +93,9 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     });
 
     if (this.transaction?.transaction_type_identifier) {
-      // Remove transaction_id from list of validate properties because it will be added in the back end.
-      const fieldsToValidate: string[] = this.validateService
-        .getSchemaProperties(this.schema)
-        .filter((p) => p !== 'transaction_id');
+      let fieldsToValidate: string[] = this.getFieldsToValidate();
+      // Remove properties populated in the back-end from list of properties to validate
+      fieldsToValidate = fieldsToValidate.filter((p) => p !== 'transaction_id' && p !== 'donor_committee_name');
```
toddlees, line 98: it seems like we'd have to do this after every time we call `getFieldsToValidate()`.  If that's the case, should we move it inside that function?

toddlees (dismissed):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
mjtravers: I had created that function to give more control to any inherited transaction type component more control over the validation fields but eventually did not need it on how the refactoring turned out. I can roll back that additional method.

Elaine-Krauss-TCG (approved):

PR #388, opened 2022-07-13
Defect/365 istrusted logindotgov code error

dheitzer:


front-end/src/app/dashboard/dashboard.component.ts
```diff
@@ -1,8 +1,38 @@
...
+          environment.ffapiEmailCookieName),
+        is_allowed: true,
+        token: null
+      }
+      this.cookieService.delete(environment.ffapiCommitteeIdCookieName)
+      this.cookieService.delete(environment.ffapiEmailCookieName)
+      this.store.dispatch(userLoggedInAction({ payload: userLoginData }));
+    }
+  }
+}
```
mjtravers, line 38: Can we drop the use of cookies and use the ngrx store 100% or is the use of cookies required?

[googling around...]

I see that the CORS_ALLOW_CRENDENTIALS setting in Django may require the use of cookies. Is that what you found, David?

Elaine-Krauss-TCG:
Do we know what's going on with the circleci test failure?

`Error response from daemon: unauthorized: authentication required`

Edit:  Seems to be working now!

mjtravers (changes requested):
Per my inline comment, is it reasonable to eliminate the use of cookies completely and rely solely on the ngrx store?

dheitzer:
> Do we know what's going on with the circleci test failure?
> 
> `Error response from daemon: unauthorized: authentication required`



> Do we know what's going on with the circleci test failure?
> 
> `Error response from daemon: unauthorized: authentication required`

Nice catch.  That was odd.  I tried rerunning the build and it worked fine now.

front-end/src/app/dashboard/dashboard.component.ts
dheitzer: @mjtravers well, the problem was that the server needs to send this data down to the client alongside a redirect.  Because I couldn't use a JSON body in the redirect, I felt my only option was to pass them using cookies.  This code then deletes the cookies once they are received following login.

dheitzer: It actually does ultimately store the cookie data in the store UserLoginData object, but only receives it from the server via cookies.  I'm open for any suggestions on how to send these from the server though if we want to use something else.

mjtravers: That works for me. Thanks for the clarification

mjtravers (approved):

Issue #350, opened 2022-06-21
Link pages to Report Left Side Navigation


AureliaKhorsand:
Purpose

Create the Report Workflow Accordion Side Navigation.

Business Reason

As a Committee Administrator, I will be able to navigate thru the Form3X report workflow. I will be able to move back and forth through the workflow

Dev Notes

This is base on the FEC Pattern library component
[https://fec-pattern-library.app.cloud.gov/components/detail/side-navigation.html](https://fec-pattern-library.app.cloud.gov/components/detail/side-navigation.html)

  The angular routes may have to be revisited if information that is needed in the current routes is missing. (e.g. :reportId)

Building Specs

### QA Notes ###
**Accordion Navigation**

**Tier 1:**
ENTER A TRANSACTION  - [ALL CAPS]
- Manage your transactions _(Go to View All Transactions for report)_
- Add a receipt _(Go to Add a Transaction screen)_
- Add a disbursements [30% Opacity] **(NOT BUILT YET-LEAVE UNLINKED)**
- Add loans and debts [30% Opacity] **(NOT BUILT YET-LEAVE UNLINKED)**
- Add other transactions [30% Opacity] **(NOT BUILT YET-LEAVE UNLINKED)**

**Tier 2:**
REVIEW A REPORT
- View summary page _(Go to Summary Page)_
- View detailed summary page _(Go to Detailed Summary Page)_
- View print preview _(Go to PDF/Print Preview page)_**(NOT BUILT YET-LEAVE UNLINKED)**
- Add a report level memo _(Go to Report Level Memo page)_**(NOT BUILT YET-LEAVE UNLINKED)**

**Section 3:**
SUBMIT YOUR REPORT
- Confirm information _(Go to Email and Verify Address page)_
- Submit report _(Go to Sign and Submit page)_
- Report status _(Go to Report Submission Status Page)_

## QA Notes ##
To display the report management menu in the left sidebar, click on "Reports" in the top menu and edit a report.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1152](https://fecgov.atlassian.net/browse/FECFILE-1152)

mjtravers:
[...]

Issue #346, opened 2022-06-17
Add Accordion Side Navigation


MitchellTCG:
Purpose 
 
Create the Report Workflow Accordion Side Navigation.

 Business Reason 
 
As a Committee Administrator, I will be able to navigate thru the Form3X report workflow. I will be able to move back and forth through the workflow

  Dev Notes 
 
This is base on the FEC Pattern library component
https://fec-pattern-library.app.cloud.gov/components/detail/side-navigation.html
The blue background is a gradient that goes from #1A4F82 to #112E51

 Building Specs 

**Accordion Navigation**

Note: Clicking the top level of the "accordion" menu expands the menu. Clicking on another top level item will collapse any others. Users are not taken to a page until the select an option _within_ the menu itself.

**Section 1:**
ENTER A TRANSACTION  - [ALL CAPS]
- Manage your transactions
- Add a receipt
- Add a disbursements [30% Opacity] - disabled
- Add loans and debts [30% Opacity] - disabled
- Add other transactions [30% Opacity] - disabled

**Section 2:**
REVIEW A REPORT
- View summary page
- View detailed summary page
- View print preview
- Add a report level memo

**Section 3:**
SUBMIT YOUR REPORT
- Confirm information
- Submit report

![NAV – 1.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/9b63349c-f2c1-4b81-85ed-c8a0d79ee51d)

### QA Notes ### 
  
This a visual check. Links will be add in next Sprint. The report accordion menu shows up on all site pages for this ticket and the info about the report above the accordion menu is place holder information. The display of the accordion menu and the proper display of the report info will be handled in the follow up ticket, #350.

### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
I can't add all the links that exist until Dev and Stage are back on.

MitchellTCG:
@AureliaKhorsand 

Updated

AureliaKhorsand:
@MitchellTCG updated requirement from PI3 Planning - the Sign and Submit step will actually be on the same page, and will be the same action for the user. I suspect we want to combine these in the menu. Otherwise, this looks good and is ready for ADTR.

MitchellTCG:
@AureliaKhorsand 
[...]

PR #393, opened 2022-07-18
Feature/346 Create report left nav menu

mjtravers:
#346 

Note to code reviewers: The scope of this ticket is to get the HTML/CSS part of the left report menu up and running. The follow on ticket #350 will link the menu items to the angular routing and also make the report menu disappear when the routing is not focused on a particular report. It's okay in this iteration to have the hard coded text in the report description part of the left nav.


Elaine-Krauss-TCG (approved):
Looks great

PR #394, opened 2022-07-18
Feature/328 logout for login dot gov

dheitzer:


dheitzer:
This one is ready for CR

Elaine-Krauss-TCG (approved):

MitchellTCG:
@toddlees Can you please CR this pull?

front-end/src/app/shared/services/api.service.ts
```diff
@@ -48,6 +48,14 @@ export class ApiService {
     return this.http.get<T>(`${environment.apiUrl}${endpoint}`, { headers: headers, params: params, withCredentials: true });
   }
 
+  public getAbsoluteUrl<T>(
```
toddlees: I think the purpose of the api url is that it's for interactions with the (FECfile) api.  Requests to other api's should be in their own services

front-end/src/app/shared/services/login.service.ts
```diff
@@ -66,12 +66,10 @@ export class LoginService {
...
       this.store.dispatch(userLoggedOutAction());
     } else {
-      this.apiService.postAbsoluteUrl(`${environment.loginDotGovLogoutUrl}`, null).pipe(
-        tap(() => {
-          this.clearUserLoggedInCookies();
-          this.store.dispatch(userLoggedOutAction());
-        })
-      ).subscribe(() => undefined);
+      this.cookieService.delete('csrftoken');
+      window.location.href = environment.loginDotGovLogoutUrl || "";
```
toddlees: Seems like this is the fix 👍 .  As for the code before, `pipe(tap(...))` [is for side effects](https://rxjs.dev/api/operators/tap) and is redundant to the subscribe below where line 71 and 72 would be more appropriate

toddlees (approved):

front-end/src/app/shared/services/login.service.ts
toddlees: This seems to be causing the unit tests to fail.  Maybe there's a way to turn this off when it's running unit tests

toddlees (changes requested):

toddlees (approved):

Issue #334, opened 2022-06-10
Allow user to enter confirmation email and verify address before submission


AureliaKhorsand:
Business Reason

As a user, I will be able to enter up to two email addresses to receive FEC confirmation of submission and verify that the committee address has or has not changed so that I can send the most updated information for FEC upon submission.

Acceptance Criteria

***Given*** the user is on the "Confirm Information" page

  ***Then*** they will be able to enter two Confirmation Emails, and the system will validate they are in the correct email format:

  1\. Email # 1 - this should default to the email from Form 1, and the user has the option to change if needed. This is a required field.

  2\. Email # 2 - this field defaults to blank, and is optional.

***And*** they will be able to indicate if their address has changed or not, and this will always default to ***NO***. 

  ***And*** if the user selects "YES", then the user will see their current address of the committee in editable textboxes, and have the ability to edit the address

*   STREET ADDRESS (A/N-34, required)
*   APARTMENT, SUITE, ETC. (A/N-34, not required)
*   CITY (A/N-30, required)
*   STATE/PROVINCE (required. Should be a dropdown) {note: will show written out state name and A/N-2 on back-end}
*   ZIP/POSTAL CODE (A/N-9, required)

***And*** they can Save and Confirm, and go back and edit as needed before the report is submitted.

[https://images.zenhubusercontent.com/306118986/7c61f9db-194f-4526-bf2b-34aa50dac37c/submit.mp4](https://images.zenhubusercontent.com/306118986/7c61f9db-194f-4526-bf2b-34aa50dac37c/submit.mp4)

![Confirm Information - 1 Blank.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/3e1b1946-49b3-4ae7-afbb-8f434d474643](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/3e1b1946-49b3-4ae7-afbb-8f434d474643))

### QA Notes ###
 
To get to the confirmation page, edit a report and the report navigation in the left sidebar will appear. Select "SUBMIT YOUR REPORT" -> "Confirmation information"

### DEV Notes ###
 
 null

### Design ###
 
 null

 See full ticket and images here: [FECFILE-1163](https://fecgov.atlassian.net/browse/FECFILE-1163)

MitchellTCG:
@AureliaKhorsand  Submit will be two steps. I'm not sure how you want to split the pages. We can discuss. Please review the prototype video.

[...]

PR #395, opened 2022-07-19
Feature/334 report email confirmation form

Elaine-Krauss-TCG:
#334 

Paired with this [API branch](https://github.com/fecgov/fecfile-web-api/pull/177)

Re: Cloudsonar feedback:

The code smell is a block that was commented out on Mitchell's request.  It contained a line that is *probably* unnecessary now that we are collecting the change of address at the end of the report instead of the beginning.  Mitchell wants it preserved in a comment in case this proves incorrect.

The security vulnerability is the regex that provides simple checking for whether or not an email is valid.  Supposedly it could be used in a Denial-of-service attack?  A saved email could be difficult for the local validator in the browser to parse later.

I don't think the fear is relevant given the small maximum length allowed.


front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
```diff
@@ -0,0 +1,184 @@
...
+    const email_1 = this.form?.get('confirmation_email_1')?.value;
+    const email_2 = this.form?.get('confirmation_email_2')?.value;
+
+    return email_1 != null && email_1.length > 0 && email_1 === email_2;
+  }
+
+  public save(jump: 'continue' | 'back' | null = null): void {
+    this.formSubmitted = true;
+
+    console.log(this.report);
```
mjtravers: Please remove all console.log() calls.

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.destroy$.complete();
+  }
+
+  public buildEmailValidator(valueFormControlName: string): ValidatorFn {
+    return (): ValidationErrors | null => {
+      const email: string = this.form?.get(valueFormControlName)?.value;
+
+      if (email?.length === 0 && valueFormControlName === 'confirmation_email_1') {
+        return { required: true };
+      }
```
mjtravers: I believe this can be replaced by adding the 'required' attribute to the HTML tag for the confirmation_email_1 form input. Or better yet, add Validators.required per the comment above about using the Angular provided validation callbacks

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
```diff
@@ -0,0 +1,184 @@
...
+      .pipe(takeUntil(this.destroy$))
+      .subscribe((committeeAccount) => this.setDefaultFormValues(committeeAccount));
+
+    this.reportCodeLabelList$ = this.store.select<ReportCodeLabelList>(selectReportCodeLabelList);
+    this.store.dispatch(updateLabelLookupAction());
+
+    // Initialize validation tracking of current JSON schema and form data
+    this.validateService.formValidatorSchema = f3xSchema;
+    this.validateService.formValidatorForm = this.form;
+    this.form.controls['confirmation_email_1'].addValidators(this.buildEmailValidator('confirmation_email_1'));
```
mjtravers: Can this be replaced with the angular provided validators? Something like: .addValidators([Validators.required, Validators.maxLength(44), Validators.email] ? That would spare us the maintenance of the custom code and eliminate that RegEx code smell. You'll still need some custom code to handle the duplicate email part.

mjtravers (changes requested):
See inline comments. Feel free to debate my recommendations.
Unit testing is a couple of percentage points shy of 80%

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
Elaine-Krauss-TCG: The Angular-provided required and maxLength validators are a great addition.

The Angular email validator can be made to work with the form as seen here:
image

But the detection is somewhat poor
image

Might still be worth it anyways to make CloudSonar happy

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
Elaine-Krauss-TCG: Removed!  I thought I'd gotten all of these...

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
Elaine-Krauss-TCG: Great suggestion!

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
Elaine-Krauss-TCG: We decided to continue using the regex-based validation, but in tandem with the Angular-provided "maxLength" and "Required" validators

mjtravers (changes requested):
This is good. I like the way you remembered to dispatch to the store to refresh the committee account info and make sure the reportCodeList has data.
Only request is to expand the unit testing to get it over 80%:
image


Oh and one ANDI issue. The other issues are with the left nav and can be ignored but here is a label issue on the email input elements of the form:
image

mjtravers (approved):

Issue #291, opened 2022-05-18
Allow user to submit Form 3X report (front-end only, not connected to FEC)


AureliaKhorsand:
Business Reason

As a user, I will be able to submit a completed Form 3X report so that I can generate a *.fec file and send it to the FEC.

Acceptance Criteria

### QA Notes ###
**Given** the Form 3X report includes all required information
**When** a user is on the "Submit Report" page 
**Then** the will be required to check the box with the following text before being able to submit:
> I certify that I have examined this Statement and to the best of my knowledge and belief it is true, correct, and complete. NOTE: Submission of false, erroneous, or incomplete information may subject the person signing this Statement to the penalties of 52 U.S.C. § 30109.

**And** when the user clicks "Submit your Form3X", they will receive a confirmation popup with the following language: 
> Are you sure you want to submit this form electronically? Please note that you cannot undo this action. Any changes needed will need to be filed as an amended report.

**And** when the user confirms submission
**Then** the user will be redirected to the View Reports page with an indication of submission.

- Dev Note: Dev will need to set up an API endpoint that provides a mock response to the front-end similar to one coming from the FEC WebUpload API

https://images.zenhubusercontent.com/306118986/de0e6507-e4db-4d9e-959d-35e208fb353d/submit.mp4

![Submit Report - Empty – 1.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/b007d232-534b-483b-b497-33db1a15901b)
![Submit Report - Empty – 2.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/9ca4d6f1-67c1-4bb4-9030-d661db7fa24e)

![Submit Report - Amend – 3.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/c49e3e4a-d97d-4ae4-8e7b-8a577cb72083)

TEXT:
If you any questions on how to enroll, confirming your email, two-factor authentication, or your Personel Key please go to the following page: **FEC Electronic Filing Password Assignment System Help**

link for password assignment: https://webforms.fec.gov/psa/help.htm

## QA Notes ##
To get to the confirmation page, edit a report and the report navigation in the left sidebar will appear. Select "SUBMIT YOUR REPORT" -> "Submit report"

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1185](https://fecgov.atlassian.net/browse/FECFILE-1185)

[...]

PR #400, opened 2022-07-21
Feature/291 report submission page

Elaine-Krauss-TCG:
#291 

Adds step 2 of the report submission process.

Unit tests are insufficient at the moment, but other than that it should be good to go.
There's also a code smell in regards to the number of parameters in a constructor?  That seems a little arbitrary...


front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
```diff
@@ -93,14 +93,24 @@ export class SubmitF3xStep1Component implements OnInit, OnDestroy {
...
-      street_2: this.report?.street_2 ? this.report.street_2 : committeeAccount?.street_2,
       city: this.report?.city ? this.report.city : committeeAccount?.city,
       state: this.report?.state ? this.report.state : committeeAccount?.state,
       zip: this.report?.zip ? this.report.zip : committeeAccount?.zip,
       confirmation_email_1:
         this.report?.confirmation_email_1 !== null ? this.report?.confirmation_email_1 : committeeAccount?.email,
       confirmation_email_2: this.report?.confirmation_email_2 !== null ? this.report?.confirmation_email_2 : null,
     });
+
+    //Get `street_2` from the same place the form got `street_1`
```
toddlees, line 104: I think this can be accomplished how it was before, but just making the condition street1 rather than street2:
```
street_2: this.report?.street_1 ? this.report.street_2 : committeeAccount?.street_2,
```

mjtravers (approved):

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step2.component.ts
```diff
@@ -0,0 +1,192 @@
...
+  ) {}
+
+  ngOnInit(): void {
+    // Refresh committee account details whenever page loads
+    this.store.dispatch(refreshCommitteeAccountDetailsAction());
+    this.stateOptions = LabelUtils.getPrimeOptions(StatesCodeLabels);
+    this.countryOptions = LabelUtils.getPrimeOptions(CountryCodeLabels);
+
+    this.report = this.activatedRoute.snapshot.data['report'];
+    if (this.report?.report_code) {
```
toddlees, line 64: `this.report_code = this.report?.report_code || ''` should accomplish the same goal.

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step2.component.ts
```diff
@@ -0,0 +1,192 @@
...
+      treasurer_last_name: this.report?.treasurer_last_name
+        ? this.report.treasurer_last_name
+        : committeeAccount?.treasurer_name_2,
+      treasurer_first_name: this.report?.treasurer_first_name
+        ? this.report.treasurer_first_name
+        : committeeAccount?.treasurer_name_1,
+    });
+
+    //If the report provided them, take the remaining fields from the report
+    if (this.report?.treasurer_last_name && this.report?.treasurer_first_name) {
```
toddlees, line 98: is there a reason that `treasurer_last_name` and `treasurer_first_name` are not handled with this condition?
It seems possible that we could end up with the first name from the report and last name from committee.  I don't think that's ever a desired outcome.  If we can assign first and last name the same way as the following fields, we could make this:

```
reportHasTreasurer = this.report?.treasurer_last_name && this.report?.treasurer_first_name;
treasurerSource = reportHasTreasurer ? this.report : this.committeeAccount;
this.form.patchValue({
  treasurer_last_name: treasurerSource.treasurer_last_name,
  treasurer_first_name: treasurerSource.treasurer_first_name,
  treasurer_middle_name: treasurerSource.treasurer_middle_name,
  treasurer_prefix: treasurerSource.treasurer_prefix,
  treasurer_suffix: treasurerSource.treasurer_suffix,
});
```

toddlees (changes requested):

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step2.component.ts
Elaine-Krauss-TCG: Great idea

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step2.component.ts
Elaine-Krauss-TCG: Love the idea, used it for handling the addresses even, but it doesn't work here because the name attributes are named differently on the CommitteeAccount than they are on the Report. (e.g. "treasurer_first_name" to "treasurer_name_1")

toddlees: oh yeah that's a bummer!  what do you think about the split conditions for first and last names, though?  Do you think they need to be handled separately or can they be included in this if/else?

Elaine-Krauss-TCG: As to the condition you describe where the treasurer_last_name comes from one source and the first name from another, this can only happen if you submit bad data to the API by going around the required fields on the front-end.  That said, I'll update this to avoid that possibility anyways.

front-end/src/app/reports/f3x/submission-workflow/submit-f3x-step1.component.ts
Elaine-Krauss-TCG: Wound up solving this by using the solution you described below for the treasurer name fields.  I feel like that's just a super-readable solution.

PR #427, opened 2022-08-03
Defect/173 report coverage date overlap

Elaine-Krauss-TCG:
API Ticket [173](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/fecgov/fecfile-web-api/173)

Updates the coverage date overlap checking to catch an edge case where a new report is created whose dates entirely encompass another report's dates.  Adds unit tests to verify that the edge case has been caught.

### DIFF
--- a/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
+++ b/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
@@ -116,6 +116,40 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
     }, 1000);
   }
 
+  /**
+   * Checks if a field's date is within another report's dates or if 
+   * another report's dates fall within the form's "from" and "through" dates
+   * 
+   * @param fieldDate the date of the field being checked
+   * @param fromDate the form's "from" date
+   * @param throughDate the form's "through" date
+   * @param targetDate the object whose date is being checked for an overlap
+   * @returns true if there is an overlap in dates
+   */
+  checkForDateOverlap(
+    fieldDate: Date,
+    fromDate: Date,
+    throughDate: Date,
+    targetDate: F3xCoverageDates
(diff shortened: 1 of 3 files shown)


Elaine-Krauss-TCG:
Shelly has pointed out that a comment on the ticket adds fixing the coverage date fields to the requirements.  A fix for that has been added.

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -101,29 +101,68 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
...
+      this.buildCoverageDatesValidator('coverage_through_date'),
+      Validators.required
+    ]);
 
     // Initialize validation tracking of current JSON schema and form data
     this.validateService.formValidatorSchema = f3xSchema;
     this.validateService.formValidatorForm = this.form;
+
+    //Re-checks the validators on the coverage date fields every second
+    setInterval(()=>{
```
toddlees: I think we really want to avoid running validation on a time schedule.  What prompted this addition?

toddlees (dismissed):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
Elaine-Krauss-TCG: Without this, the validation for a field only runs on form submission and on interaction with the given field.  Normally, this is just fine, but with the edge case this ticket is trying to solve, both fields need to show the validation error because neither one field is at fault.  This is something I would avoid normally, but it seems necessary here, and it comes at a negligible performance cost.  (I ran a performance profile on my machine using the dev panel, and the total runtime of timer events in a 30 second window came out to 180ms)

mjtravers (approved):

Issue #415, opened 2022-07-27
Fix display of report left nav for edit transaction screens


mjtravers:
The urls assigned to the edit transaction screens need to be changed so that the report left nav displays and shows the "ENTER A TRANSACTION" -> "Manage your transactions" menu item

1.  1.  Acceptance Criteria ##

             WHEN a user clicks on "Manage your transactions" in the left report nav when editing a report, and then edits a transaction, THEN the left report nav will still be displayed the the selected item will be "Manage your transactions"

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1123](https://fecgov.atlassian.net/browse/FECFILE-1123)

mjtravers:
This ticket was split off from the left nav work done in #350: https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/fecgov/fecfile-web-app/350

mjtravers:
Passed CR by Elaine. Sending to QA

WiseQA:
QA review verified the report left navigation for edit transaction screens display was fixed.

1.  Click on "Manage your transactions" in the left navigation when editing a report.

2.  After editing and saving a report verify  "Manage your transactions" from the left navigation is still displayed and selected

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/9711b502-ce96-4a1f-8c16-f836e0531d36)

QA Review Completed.   Moved to Stage Ready.

PR #437, opened 2022-08-08
Bug/415 Rearrange transaction routing to accomodate report left nav actions

mjtravers:
To solve this I did not move the transactions module into the reports module, afterall. Instead, the report step 3 component was moved from the reports module to the transactions module and renamed transaction-list component

#415


Elaine-Krauss-TCG (approved):
Looks pretty straightforward.  I like the idea of moving the manage transactions page to be under the transactions app.

Issue #355, opened 2022-06-22
Allow user to view submitted Form 3X reports (submitted to FEC)


AureliaKhorsand:
Business Reason

As a user, I will be able to view closed/submitted F3X reports as read only.

Acceptance Criteria

***If*** a report has been submitted to FEC

  ***Then*** it will have a Status of "Submitted" on the Manage Reports Page

  ***And when*** the user opens the report (or immediately upon submission - update from #291)

  ***Then*** the following pages will be available:

***Accordion Navigation***

REVIEW A REPORT

*   View summary page
*   View detailed summary page
*   View print preview

SUBMIT YOUR REPORT

*   View Submission Status _(this will be the default landing page for opening a Submitted Report)_

![Custom Size – 91.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/0e5d4294-931d-414d-8587-7f41d73966d1](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/0e5d4294-931d-414d-8587-7f41d73966d1))

![Print Preview – 1.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/56a061e5-ca62-48e9-ad69-515c3289bb54](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/56a061e5-ca62-48e9-ad69-515c3289bb54))

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1147](https://fecgov.atlassian.net/browse/FECFILE-1147)

mjtravers:
Passes CR. Sending to QA

mjtravers:
Bug found in QA: The left report nav does not switch to read only  on submission, only when the report is left and returned to.

mjtravers:
Elaine's code updates to fix the report left menu look good. Sending to QA.

WiseQA:
Moved back to QA review

MitchellTCG:
![Screen Shot 2022-08-16 at 2.06.32 PM.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/2cf8bf08-f32a-4043-9743-0242e8938250)

![Screen Shot 2022-08-16 at 2.06.42 PM.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7c9746ec-1371-4ffc-a280-14c05f2518b0)

[...]

PR #455, opened 2022-08-12
Feature/355 read only reports

Elaine-Krauss-TCG:
#355 

Blocks the displaying of and navigation to certain links after a report has been submitted and while it is either Processing or Accepted.


front-end/src/app/shared/services/report-is-editable.service.ts
```diff
@@ -0,0 +1,24 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class ReportIsEditableService {
+  constructor(private store: Store){}
+
+  isEditable(): Observable<boolean> {
+    let observe = new Observable<boolean>();
+    this.store.select(selectActiveReport).subscribe((report: Report | null)=>{
+      const status = report?.upload_submission?.fec_status;
```
toddlees: `fec_status` only comes into play once it's been submitted to fec.  We don't want the report to be editable from the moment the user makes the submission.  I think this condition should take into account the `fecfile_task_state`

```
  isEditable(): Observable<boolean> {
    return this.store.select(selectActiveReport).pipe(
      map((report: Report | null) => {
        const uploadSubmission = report?.upload_submission;
        const fecStatus = report?.upload_submission?.fec_status;
        const fecfileTaskState = report?.upload_submission?.fecfile_task_state;
        return !uploadSubmission || fecStatus == 'REJECTED' || fecfileTaskState == 'FAILED';
      })
    );
  }
```

toddlees (dismissed):

front-end/src/app/shared/services/validate.service.spec.ts
```diff
@@ -83,6 +83,7 @@ describe('ValidateService', () => {
     let result: ValidationErrors | null = validator(
       service.formValidatorForm.get('L6a_cash_on_hand_jan_1_ytd') as FormControl
     );
+    expect(result).not.toEqual(null);
```
toddlees: I saw this removed in a pr from @mjtravers.  Is it supposed to be back?

front-end/src/app/shared/guards/report-is-editable.guards.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Observable } from 'rxjs';
+import { ReportIsEditableService } from '../services/report-is-editable.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ReportIsEditableGuard implements CanActivate {
+  constructor(private editableService: ReportIsEditableService) {}
+
+  canActivate(): Observable<boolean> {
```
toddlees: This is pretty dang cool!

front-end/src/app/shared/services/report-is-editable.service.ts
Elaine-Krauss-TCG: Implemented verbatim.  Works great!

front-end/src/app/shared/services/validate.service.spec.ts
Elaine-Krauss-TCG: It was a check that was failing because it didn't like using `toBe(null)`

It could be removed, it could be changed, I'd fixed it separately and I just kept it in.

front-end/src/app/shared/guards/report-is-editable.guards.ts
Elaine-Krauss-TCG: The guard pattern is quite nifty

front-end/src/app/shared/services/validate.service.spec.ts
mjtravers: It can be in or out. The test is not consequential which is why I removed it.

mjtravers (approved):

Issue #290, opened 2022-05-18
 Allow user to create a PDF/print preview of the Form 3X report (Page 1, SP, DSP, SA)


AureliaKhorsand:
Business Reason

As a user, I will be able to generate and view a PDF of my report so that I can see how Form 3X looks as a paper form.

Acceptance Criteria

***Given*** a report in any status

  ***When*** the user goes to the "PDF/Print Preview" page

  ***Then*** they can generate a PDF of their report in it's current state

  ***And*** when they click to generate a PDF, they will receive a message indicating "This may take a moment..."

  ***And*** if after 5 seconds the PDF:

*   IS ready: the user will have the option to download it as a PDF by opening it in a new tab; the system will provide a date/time stamp for when the PDF was generated; and the button to compile a new PDF will be available.
*   IS NOT ready: the system will indicate that "Your report is still being processed. Please check back later to access your PDF" and the process will continue in the background until it is ready _(Dev note: check status endpoint each time page is accessed)_

***And*** once the URL is no longer available from WebPrint service _(after 24 hours)_, the page resets _(only one button for Compile PDF)_.

Dev Tasks

  Create an endpoint on the webservice worker to submit a fec file to the FEC WebPrint service. Return the WebPrint response to the Celery broker

Documents: [https://efoservices.fec.gov/static/webprint_help.html#webservice](https://efoservices.fec.gov/static/webprint_help.html#webservice)
[https://webforms.fec.gov/psa/getstarted.htm](https://webforms.fec.gov/psa/getstarted.htm)

![Compile 2.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/8b1a919e-3366-4a4b-b92f-de32397e2e85](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/8b1a919e-3366-4a4b-b92f-de32397e2e85))

![Custom Size – 85.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/a6a03b6c-d74d-4476-bb14-a4976ba485a2](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/a6a03b6c-d74d-4476-bb14-a4976ba485a2))

[...]

PR #458, opened 2022-08-12
Feature/290 web print submission

Elaine-Krauss-TCG:
#290 

Allows the user to submit a report to the FEC Web Print service through the front-end


mjtravers (approved):
I renamed some components to match convention and the status models to match the database table names. I also did some prettier formatting to the files and updated the left nav menu links.

PR #494, opened 2022-08-23
Feature/480 angular upgrade

toddlees:


front-end/src/app/reports/f3x/submission-workflow/submit-f3x-status.component.html
```diff
@@ -66,12 +66,12 @@ <h3 class="no-vert-margins">
 </div>
 <div class="grid">
   <div class="col-12">
-    <p *ngIf="report.upload_submission?.fec_status!=='ACCEPTED' && report.upload_submission?.fec_status !=='REJECTED'">
+    <p *ngIf="!['ACCEPTED', 'REJECTED'].includes(report.upload_submission?.fec_status || '')">
```
Elaine-Krauss-TCG, line 69: This is just a really elegant change, :+1:

Elaine-Krauss-TCG (approved):

dheitzer (approved):

front-end/src/app/reports/reports-routing.module.ts
```diff
@@ -67,6 +67,7 @@ const routes: Routes = [
     path: 'f3x/memo/:reportId',
     component: ReportLevelMemoComponent,
     canActivate: [ReportIsEditableGuard],
+    resolve: { report: ReportResolver },
```
mjtravers: This can be removed. Because an issue where the ReportIsEditableGuard needed to have the full report object, it pulls the report like the ReportResolver and makes it available in the ngrx store. So in effect, only the ReportIsEditableGuard or the ReportResolver is needed but not both, otherwise we end up making 2 API calls to the backend where only one is needed. There are some inline comments squirreled away in the code regarding this

mjtravers (changes requested):
See comment

mjtravers (approved):

PR #541, opened 2022-08-29
Feature/111 intl phone num take two

dheitzer:


Elaine-Krauss-TCG:
I was interested in how the field would actually look on the front-end, so I pulled and went to make a Contact.  Unfortunately, I think I found a bug?

When I selected a country code (Malaysia, chosen at random), if I typed in a phone number, it shows an error "this field must contain 10 alphanumeric characters" when there are 6 characters entered,
image

but then it shows "this field cannot contain more than 10 alphanumeric characters" at 7 entered characters.
image

Elaine-Krauss-TCG:
The error message appears at varying points depending on the length of the country code
image

image

dheitzer:
@Elaine-Krauss-TCG did you rebuild your client on that branch to pull in the latest fecfile-validation change? There was a needed validation change for this field that could be causing this behavior.

Elaine-Krauss-TCG:
> @Elaine-Krauss-TCG did you rebuild your client on that branch to pull in the latest fecfile-validation change? There was a needed validation change for this field that could be causing this behavior.

After talking with David about this, we found that fully rebuilding the front-end did solve my problem, but there is another error altogether.  If you create a new contact and type in a phone number without changing the country code, it does not recognize the phone number as having a valid length when it should.

dheitzer:
> 

@Elaine-Krauss-TCG this bug has been addressed and this should be ready for CR now.

Elaine-Krauss-TCG (approved):

Elaine-Krauss-TCG (approved):

front-end/src/app/shared/components/fec-intl-tel-input/fec-intl-tel-input.component.html
```diff
@@ -0,0 +1,5 @@
+<div>
```
toddlees: We've been following a convention of avoiding abbreviation when possible.  Clarity of code is highly impactful and shrinking variable/class/filenames doesn't gain us much.  I'd highly recommend spelling out `intl` and `tel` as `international` and `telephone` or just `phone`

toddlees (changes requested):
I think there's an unimplemented requirement:
>ZIP (A/N-9, not required only if Country is NOT United States) [validate that optional appears if state chosen]

Additionally there seems to be a problem with the requirements.  Having the number after the country code always be 10 (no more no less) breaks functionality for countries with more (China 11 digits) and less (Chili 9 digits) digits.  We can probably pass this and make a followup ticket for expanding that rule.  Similarly we'll want to make a ticket for allowing spaces, parentheses, and hyphens in the input (stripping them out to save).

front-end/src/app/shared/components/fec-intl-tel-input/fec-intl-tel-input.component.ts
```diff
@@ -0,0 +1,77 @@
...
+@Component({
+  selector: 'app-fec-intl-tel-input',
+  templateUrl: './fec-intl-tel-input.component.html',
+  styleUrls: ['./fec-intl-tel-input.component.scss'],
+})
+export class FecIntlTelInputComponent implements AfterViewInit, OnDestroy, 
+  ControlValueAccessor {
+  @ViewChild('telInput') telInput: ElementRef<HTMLInputElement> | null = null;
+
+  private iti: intlTelInput.Plugin | null = null;
```
toddlees: I see that this comes from the example code from their site, but could we spell this out more.  When i first encountered `iti` in the spec above I couldn't even guess what it meant.

Elaine-Krauss-TCG (approved):

Elaine-Krauss-TCG (approved):

toddlees (approved):

Issue #445, opened 2022-08-11
System to check "Memo Item" as read-only on PAC Joint Fundraising Transfer Memo


AureliaKhorsand:
PAC Joint Fundraising Transfer Memo will always have Memo Item checked, and it cannot be unchecked by the user, so that the transaction is never aggregated onto the Detailed Summary page.

Acceptance Criteria

***Given*** the user is entering or editing a PAC Joint Fundraising Transfer Memo

  ***Then*** the "Memo Item" checkbox will be checked and read-only.

![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/f7364dc6-f128-451a-9750-c7a3b01c18a5](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/f7364dc6-f128-451a-9750-c7a3b01c18a5))

### QA Notes ###

No additional information needed for this ticket.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1105](https://fecgov.atlassian.net/browse/FECFILE-1105)

WiseQA:
QA Review verified the Memo Item checkbox is "Read Only" for PAC Joint Fundraising Transfer Memo

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/050c1846-8e5c-4dc1-bfb1-a0747ad15c3f)

QA Review Completed.   Moved to Stage Ready

PR #543, opened 2022-08-30
Feature/read only memo item for jf tran memo

Elaine-Krauss-TCG:
#445 

Creates a list of ReadOnlyMemoItems that tracks SchATransactions that can _only_ be memo items.  The Group F Transaction Form automatically checks the memo item checkbox if the transaction type is found in the previously mentioned list.  This checkbox is then set to read-only.

Paired with [validate](https://github.com/fecgov/fecfile-validate/pull/111) and [api](https://github.com/fecgov/fecfile-web-api/pull/216) PR's


toddlees (dismissed):
This can actually be addressed largely at the schema level.  In the JF_TRAN_PAC_MEMO, memo_code should be given the property `const: true`.  We can use that property to assign the value in the form and the presence of a const in the schema tells us to make the input read only.

front-end/src/app/transactions/transaction-group-f/transaction-group-f.component.ts
```diff
@@ -38,8 +44,12 @@ export class TransactionGroupFComponent extends TransactionTypeBaseComponent imp
     protected override transactionService: TransactionService,
     protected override validateService: ValidateService,
     protected override fb: FormBuilder,
-    protected override router: Router
+    protected override router: Router,
+    protected activatedRoute: ActivatedRoute
   ) {
     super(messageService, transactionService, validateService, fb, router);
+    console.log(this.readOnlyMemo);
```
dheitzer: If this code stays (per Todd's comment), maybe remove these log statements.

front-end/src/app/transactions/transaction-group-f/transaction-group-f.component.ts
```diff
@@ -38,8 +43,32 @@ export class TransactionGroupFComponent extends TransactionTypeBaseComponent imp
     protected override transactionService: TransactionService,
     protected override validateService: ValidateService,
     protected override fb: FormBuilder,
-    protected override router: Router
+    protected override router: Router,
+    protected activatedRoute: ActivatedRoute
   ) {
     super(messageService, transactionService, validateService, fb, router);
+
+    of(this.activatedRoute.snapshot.data['transactionType']).subscribe((transactionType: TransactionType) => {
```
toddlees: https://angular.io/api/router/ActivatedRoute#description shows how activatedRoute provides the observable for you to subscribe to.

toddlees (approved):

Issue #550, opened 2022-09-01
Remove old "page two" of Report Creation Workflow


AureliaKhorsand:
Task: Remove "step2" page of report creation, since that data entry has now been moved to the submit report process and the page for adding a report level memo and the page is no longer needed. "Save and Continue" after coverage date selection should lead the user directly to "Manage Transactions".

Acceptance Criteria

***Given*** the user has selected a report type and coverage dates for a new report

  ***When*** they click "save and continue"

  ***Then*** they will be on the "Manage Transactions" page for that new report.

"Save and Continue" button from Create Report page 1:

  ![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/fd5dc96b-9ccc-44af-b376-a3294bb9bb31](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/fd5dc96b-9ccc-44af-b376-a3294bb9bb31))

Current "step two" page to be removed:

  ![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/35ba3909-3e6b-46b2-bb08-a8db9231118f](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/35ba3909-3e6b-46b2-bb08-a8db9231118f))

Manage Transactions page:

  ![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/9e0c262b-e0be-4a63-80cf-cee9878ac9a2](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/9e0c262b-e0be-4a63-80cf-cee9878ac9a2))

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1022](https://fecgov.atlassian.net/browse/FECFILE-1022)

Elaine-Krauss-TCG:
Step 2 of the Report creation process has been removed.  All unit tests are passing.

![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/5355fafa-aa94-44b1-81f1-b1495673536c)

mjtravers:
One code issue was found in CR. Elaine is making a code change to flatten out a couple of rxjs subscriptions and then will resubmit to CR

mjtravers:
Code change requests have been made. Ticket passes CR and sending to QA

WiseQA:
[...]

PR #564, opened 2022-09-12
Feature/550 remove report creation step 2

Elaine-Krauss-TCG:
#550 

Removes Report Creation Step 2


front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -240,7 +241,16 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
         // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
         this.reportService.getTableData().subscribe();
 
-        this.router.navigateByUrl(`/reports/f3x/create/step2/${report.id}`);
+        this.store
```
toddlees: Nested subscriptions should be avoided (like nested callbacks or promises https://angular-checklist.io/default/checklist/rxjs/Z1eFwa9).
We can flatten this by changing 239 to:
```
createSummary$ = this.f3xSummaryService.create(summary, this.formProperties);
selectCashOnHand$ = this.store.select(selectCashOnHand);
createSummary$.pipe(combineLatestWith(selectCashOnHand$)).subscribe(([report, cashOnHand]) => {
```
combineLatestWith https://rxjs.dev/api/operators/combineLatestWith

then, at line 244 you can just write your condition without setting up another subscription

toddlees (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
Elaine-Krauss-TCG: That's a super interesting idea.  I'll take a look.

mjtravers: Yes, agree with Todd's comment about not nesting subscribes. @Elaine-Krauss-TCG : Also take a look at mergeMap() which I have found to be a super useful pattern to know as well: https://www.learnrxjs.io/learn-rxjs/operators/transformation/mergemap

mjtravers (approved):
Looks good. Can be merged to develop after the code change is made to flatten the nested subscriptions as noted by Todd

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -235,22 +236,40 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
...
 
-    this.f3xSummaryService.create(summary, this.formProperties).subscribe((report: F3xSummary) => {
-      if (jump === 'continue') {
-        // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
-        this.reportService.getTableData().subscribe();
+    //Observables are *defined* here ahead of their execution
+    const create = this.f3xSummaryService.create(summary, this.formProperties);
+    // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
+    const tableData = this.reportService.getTableData();
+    const cashOnHand = this.store.select(selectCashOnHand);
```
mjtravers: A small nit-pick: the convention is to add a "$" suffix to variables that contain observables. e.g. tableData$, cashOnHand$

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -235,22 +236,40 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
...
-    });
+    //Create the report, update cashOnHand based on all reports, and then retrieve cashOnHand in that order
+    create
+      .pipe(
+        switchMap((report) => {
+          return zip(of(report), tableData);
+        }),
+        switchMap(([report, _]) => {
+          return zip(of(report), cashOnHand);
+        }),
```
mjtravers, line 251: This seems like a lot of work to subscribe to 3 independent observables. Did you consider forkJoin()?  ... and maybe tap() the tableData$ so not to have the extra _ array element?

Something like:
forkJoin([create$, cashOnHand$].pipe(tap(tableData$.subscribe()).subscribe([report, coh]) => {}

mjtravers (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -235,22 +236,40 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
...
     }
 
-    this.f3xSummaryService.create(summary, this.formProperties).subscribe((report: F3xSummary) => {
-      if (jump === 'continue') {
-        // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
-        this.reportService.getTableData().subscribe();
+    //Observables are *defined* here ahead of their execution
+    const create = this.f3xSummaryService.create(summary, this.formProperties);
+    // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
+    const tableData = this.reportService.getTableData();
```
mjtravers: I think this will need a pipe(takeUntil(this.destroy$))

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
mjtravers: I think this will need a pipe(takeUntil(this.destroy$)), too

Elaine-Krauss-TCG: :+1:

toddlees: not necessary since there's one subscription

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
toddlees: the issue with using tap here is that we wouldn't get the assured sequence of operation since tap is for side effects and returns the original observable.  We need it to return a new observable dependent on getTableData$ completing

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
Elaine-Krauss-TCG: :+1:

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
Elaine-Krauss-TCG: :+1:

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
toddlees: no need for takeUntil.  There is one subscription in this function and it has a takeUntil at like 255

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -235,22 +237,39 @@ export class CreateF3XStep1Component implements OnInit, OnDestroy {
...
 
-    this.f3xSummaryService.create(summary, this.formProperties).subscribe((report: F3xSummary) => {
-      if (jump === 'continue') {
-        // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
-        this.reportService.getTableData().subscribe();
+    //Observables are *defined* here ahead of their execution
+    const create$ = this.f3xSummaryService.create(summary, this.formProperties).pipe(takeUntil(this.destroy$));
+    // Save report to Cash On Hand in the store if necessary by pulling the reports table data.
+    const tableData$ = this.reportService.getTableData().pipe(takeUntil(this.destroy$));
+    const cashOnHand$ = this.store.select(selectCashOnHand).pipe(takeUntil(this.destroy$));
```
toddlees: @mjtravers @Elaine-Krauss-TCG these takeUntil's a redundant if we've refactored this into a single subscription.  If we have one takeUntil at the end of the pipe below, the subscription ends when the component is destroyed.

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
toddlees: to get rid of the tabledata in the subsequent switchmap signature, instead of zipping the report and tabledata in the previous switchmap, we'd have to switchmap the tabledata to return the report:
```
    createReport$.pipe(
      switchMap((report) =>  getTableData$.pipe(switchMap(()=>of(report)))),
      switchMap((report)=> zip(of(report), cashOnHand$)),
      takeUntil(this.destroy$)
    ).subscribe(([report, cashOnHand]) => {
```

createReport$ resolves -> getTableData$ resolves and is mapped return the report -> resolve and return both of(report) and cashOnHand$ -> clean up this flow/subscription when this component is destroyed

even better!
```
    createReport$.pipe(
      switchMap((report) =>  getTableData$.pipe(map(()=>report))),
      switchMap((report)=> zip(of(report), cashOnHand$)),
      takeUntil(this.destroy$)
    ).subscribe(([report, cashOnHand]) => {
```

Elaine-Krauss-TCG:
The switchmaps no longer have an unused variable, so that should no longer be an issue, @toddlees 

image

toddlees:
> The switchmaps no longer have an unused variable, so that should no longer be an issue, @toddlees
> 
> image

👍 I see you hid it in the response array.  if you want to drop the tabledata after the observable resolves you can map it to report like below, but it works that way too!

```
    createReport$.pipe(
      switchMap((report) =>  getTableData$.pipe(map(()=>report))),
      switchMap((report)=> zip(of(report), cashOnHand$)),
      takeUntil(this.destroy$)
    ).subscribe(([report, cashOnHand]) => {
```

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
mjtravers: @Elaine-Krauss-TCG Todd has a good catch with not needing the takeUntil()'s if one is placed after the switchMaps inthe the create$.pipe: https://stackoverflow.com/questions/58520753/angular-should-i-unsubscribe-from-switchmap

toddlees (approved):

mjtravers (approved):

Issue #72, opened 2022-01-18
Allow user to use Lookup within Transactions for existing Individual Contacts-Part 1


AureliaKhorsand:
Business Reason

The system will employ a Lookup feature when entering transactions in order to allow the user to pull existing contact information from the committee account's Contact Management table. This will also assist the FEC in making more accurate data available to the public.

Acceptance Criteria

_Note: The transaction entry form will _only_ show the contact lookup until a contact is selected. Once a contact is selected, the transaction form will show._

***Given*** the user is entering a transaction

  ***And*** they have indicated an Entity Type of "IND" (or "IND" is required entity type)

  ***Then*** they will be able to search their committee account's Contact Management table by _name_ using a Lookup feature to find and select an existing contact

  ***And*** the results will including the following information in the dropdown under the text "Select an existing individual contact":

*   Last Name, First Name
*   "In Contacts"

***And*** the Lookup results will:

*   highlight the matching text from the search box within the results _(QA note: excluding delimiters such as spaces and commas)_
*   highlight a selection grey if hovered over with the pointer
*   be limited to 10 maximum _(Dev/QA note: for now, users need to type more to narrow it down. #542 will address when results are more than 10.)_

***And*** when I select the contact I want, the system will pull in the contact information necessary for the transaction and show the full transaction entry form. 

  ***(Note information to include Contributor, Address Fields and Employer and Occupation Fields)***

  ***If*** there are no results, the system will indicate as such.

BA notes

FECFile Online basic requirements as follows:

*   Users will be able to search their committee account's Contact Management table using a Lookup feature to find and select an existing contact for a transaction
*   When the user selects the contact, the system will pull in the contact information necessary for the transaction.
*   The transaction will be associated with that contact (used for aggregations/itemization logic)

DEV notes

From this point on, contacts and transactions need to be connected. _Each transaction must be associated with a contact, such as a UUID._

QA notes

Relevant existing Transactions to test:

*   Individual Receipt
*   Offsets to Operating Expenditures
*   Other Receipts

***(Note information when selecting a contact from the lookup includes Contributor, Address Fields and Employer and Occupation Fields)***

To test, identify an existing individual contact in Contact Management for your committee account and use that name to search.

Wireframes: 

  (Note: down arrow next to "individual" only if other entity types are valid for selection.)

[...]

PR #567, opened 2022-09-13
Feature/72 individual lookup

dheitzer:
#72


front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
```diff
@@ -1,29 +1,55 @@
...
+          optionLabel="name"
+          optionValue="code"
+          [readonly]="contactTypeReadOnly"
+          [styleClass]="contactTypeStyleClass">
+        </p-dropdown>
+      </div>
+      <div class="col-2 pl-0">
+        <p-autoComplete 
+          [style]="{'width':'325px'}" 
+          [inputStyle]="{'width':'325px'}"
```
mjtravers: What is fixing the width of the autocomplete explicitly fix? It's not working dynamically when the screen size changes? Could we use percentages rather than pixels?

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -158,4 +158,17 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
       contribution_purpose_descrip: this.contributionPurposeDescrip,
     });
   }
+
+  onContactLookupSelect(value: Contact) {
+    if (value && typeof value.id === 'number') {
```
mjtravers: type of value.id === 'number' will not work after we switch to UUID primary keys on all the models/tables

front-end/src/app/shared/models/contact.model.ts
```diff
@@ -73,42 +73,100 @@ export class Contact extends BaseModel {
...
+
+export class FecfileIndividualLookupData extends Contact {
+  static toSelectItem(data: FecfileIndividualLookupData,
+    inContacts: boolean): ContactLookupSelectItem<Contact> {
+    return (
+      data && {
+        label: `${data.last_name}, ${data.first_name}`,
+        value: data,
+        inContacts: inContacts,
+      }
```
mjtravers, line 157: What happens when data is null or undefined? Does the toSelectItem() return the boolean false?

mjtravers (changes requested):
Just a couple of questions and a heads up about the impending UUID change to the table primary keys.

Also, there is a tiny SonarCloud code smell to clean up

front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
```diff
@@ -1,29 +1,55 @@
...
+          dropdownMode="current"
+          dropdownIcon="pi pi-search"
+          [suggestions]="contactLookupList"
+          formControlName="selectedContact"
+          placeholder="Search for a contact"
+          [forceSelection]="true"
+          (onSelect)="onContactSelect($event)">
+          <ng-template let-contact pTemplate="item">
+            <div>
+              <span [innerHTML]="contact.label | highlightTerm:searchTerm"
```
toddlees: this is really nifty

front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
dheitzer: Originally I had to set this to be able to resize the autocomplete component.  I was able to remove it now as it seems using bootstrap grid fixed this issue.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
dheitzer: Nice catch.  I went ahead and reworked some of the model classes to allow us to check instanceof instead.

front-end/src/app/shared/models/contact.model.ts
dheitzer: I reworked the model classes to use 'this' instead and change this method from static, so it should now never be null.

dheitzer:
> Just a couple of questions and a heads up about the impending UUID change to the table primary keys.
> 
> Also, there is a tiny SonarCloud code smell to clean up

these should be all addressed now in the latest push.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
mjtravers: Nice extension of PrimeNg interface

mjtravers (approved):

Elaine-Krauss-TCG (commented):
Searching based on both first and last name together doesn't seem to work.  (both "First Last" & "Last, First")

Being able to search with both at the same time isn't part of the AC, but we might want to cover it.

front-end/src/app/shared/models/contact.model.ts
```diff
@@ -73,42 +73,111 @@ export class Contact extends BaseModel {
...
+        inContacts: true,
       }
     );
   }
 }
 
 export class CommitteeLookupResponse {
-  fec_api_committees: FecCommitteeLookupData[] | null = null;
-  fecfile_committees: FecCommitteeLookupData[] | null = null;
+  fec_api_committees: FecApiCommitteeLookupData[] | null = null;
```
toddlees: I'd recommend changing this to reflect what could come from the server (either an array or undefined).  The server never returns a fec_api_committees of value `null`.
`fec_api_committees: FecApiCommitteeLookupData[] | undefined;`

toddlees (approved):

dheitzer:
Updates are ready for CR.  I think the one sonar code smell should be ignored.

Elaine-Krauss-TCG (approved):
Works great :+1:

dheitzer:
@toddlees @Elaine-Krauss-TCG I pushed up the fix for handling empty results.

Issue #556, opened 2022-09-07
Convert default null values to undefined


mjtravers:
The model properties were assigned a default value of null in most cases, this has turned out to be problematic for a number of reasons including results from the ? operator on objects and the standard return value for many Angular helper functions.

The task is to convert the null default value to undefined in the app models and test (unit and e2e) to make sure the app still functions as expected.

1.  1.  Acceptance Criteria ##

*   The app model properties default to undefined where null
*   unit tests run without error
*   general smoke tests run without error

1.  1.  QA Notes ##

             Testing will consist of screen shot showing that unit tests ran without errors

### QA Notes ###
The task is to convert the null default value to undefined in the app models and test (unit and e2e) to make sure the app still functions as expected.

## Acceptance Criteria ##
* The app model properties default to undefined where null
* unit tests run without error
* general smoke tests run without error

## QA Notes ##
Testing will consist of screen shot showing that unit tests ran without errors

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1019](https://fecgov.atlassian.net/browse/FECFILE-1019)

mjtravers:
Unit test screenshot for QA:
![image.png](https://images.zenhubusercontent.com/61b9fc72d99b6f786e0d798c/f2dff212-048f-4f4e-bd01-145aa2e4a8f6)

WiseQA:
QA Inspection Review completed.   DEV team verified Unit Tests associated with fix passed.

QA Review Completed.   Moved to Stage Ready.

PR #572, opened 2022-09-15
Feature/556 Convert model default property values from null to undefined

mjtravers:
#556


Elaine-Krauss-TCG (approved):
There's million and one changes, but they all look pretty straight-forward

Issue #538, opened 2022-08-26
Allow user to use Lookup within transaction for existing Committee Contact-Part 1


AureliaKhorsand:
Business Reason

The system will employ a Lookup feature when entering transactions in order to allow the user to pull existing contact information from the committee account's Contact Management table. This will also assist the FEC in making more accurate data available to the public.

Acceptance Criteria

_Note: The transaction entry form will _only_ show the contact lookup until a contact is selected. Once a contact is selected, the transaction form will show._

***Given*** the user is entering a transaction with the Entity Type of "COM", whether user-selected or required by transaction type, 

  ***Then*** they will be able to search their committee account's Contact Management table by _name or ID_ using a Lookup feature to find and select an existing contact

  ***And*** the results will including the following information in the dropdown:

*   Committee Name
*   Committee ID
*   "In Contacts"

***And*** the Lookup results will:

*   highlight the matching text from the search box within the results
*   highlight a selection grey if hovered over with the pointer
*   be limited to ***5*** maximum _(Dev/QA note: for now, users need to type more to narrow it down. #542 will address when results are more than 10.)_

***And*** when I select the contact I want, the system will pull in the contact information necessary for the transaction and show the full transaction entry form.

  ***If*** there are no results, the system will indicate as such.

BA notes

FECFile Online basic requirements as follows:

*   Users will be able to search their committee account's Contact Management table using a Lookup feature to find and select an existing contact for a transaction
*   When the user selects the contact, the system will pull in the contact information necessary for the transaction.
*   The transaction will be associated with that contact (used for aggregations/itemization logic)

QA notes

Relevant existing Transactions to test:

*   Joint Fundraising Transfer
*   PAC JF Transfer Memo
*   Other Receipts

NOTE: Other Receipts is Group C therefore Committee ID field is not displayed.

To test, identify an existing committee contact in Contact Management for your committee account and use that ID or name to search.

Wireframes: 

  (Note: down arrow next to "Committee" only if other entity types are valid for selection.)

![image.png]([https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/10bf3724-11d9-40ae-a609-7394adf62bca](https://images.zenhubusercontent.com/61a50e5fdcda4774a0cf2a0e/10bf3724-11d9-40ae-a609-7394adf62bca))

[...]

PR #585, opened 2022-09-22
Added contact lookup to group e transaction screen

mjtravers:
#538


front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -174,6 +174,9 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
             this.form.get('contributor_occupation')?.setValue(value.occupation);
             break;
           case ContactTypes.COMMITTEE:
+            this.form.get('donor_committee_fec_id')?.setValue(value.committee_id);
```
dheitzer, line 177: nice catch

dheitzer (approved):

Issue #460, opened 2022-08-15
Allow user to enter Earmark Memo on Schedule A (Group G)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Earmark Memo”** sub-transaction under **”Earmark Receipt”**  on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

**NOTE:** This is a "child" transaction that will be associated with a parent transaction.

 Acceptance Criteria 
 

**Given** I am a Committee Administrator editing an open Form 3X report
**And** I am editing a Earmark Receipt transaction on an open report
**Then**  I will **be required** to complete step two to add an Earmark Memo transaction
**And** I will be required to enter the following information (unless otherwise noted):

# Conduit 
#
-	**ENTITY TYPE**
Field description in API: ENTITY TYPE
Data field: Dropdown
Dropdown options: Individual, Committee 
*Note - please default to COMMITTEE
---
_if CONTRIBUTOR TYPE = Committee_
-	**CONDUIT NAME** (A/N-200, Required)
Field description in API: CONTRIBUTOR ORGANIZATION
Data field: textbox
-	**CONDUIT FEC ID** (A/N-9, Required)
Field description in API: DONOR COMMITTEE FEC ID
Data field: textbox
With the ID in the following format: "CXXXXXXXX" where X is a numeric value _(C, then 8 numbers)_
**FEC ID IS A STOP-GAP UNTIL THE LOOKUP IS IN PLACE**
---
_if CONTRIBUTOR TYPE = Individual_
-	**LAST NAME** (A/N-30, Required)
Field description in API: CONTRIBUTOR LAST NAME
Data field: textbox
-	**FIRST NAME** (A/N-20, Required)
Field description in API: CONTRIBUTOR FIRST NAME
Data field: textbox
-	**MIDDLE NAME** (A/N-20, Optional)
Field description in API: CONTRIBUTOR MIDDLE NAME
Data field: textbox
-	**PREFIX** (A/N-10, Optional)
Field description in API: CONTRIBUTOR PREFIX
Data field: textbox
-	**SUFFIX** (A/N-10, Optional)
Field description in API: CONTRIBUTOR SUFFIX
Data field: textbox
-	**EMPLOYER** (A/N-38, **Required regardless of aggregate**)
Field description in API: CONTRIBUTOR EMPLOYER
Data field: textbox
[...]

PR #586, opened 2022-09-22
460 Create earmark memo input form

mjtravers:
#460


front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -50,45 +50,41 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
...
+    validateService.formValidatorForm = form;
 
-    // Intialize form on "Individual" entity type
-    if (this.isExisting()) {
-      const txn = { ...this.transaction } as SchATransaction;
-      this.form.patchValue({ ...txn });
-      this.form.get('entity_type')?.disable();
+    // Disable entity type form input field when editing a record.
+    if (this.isExisting(transactionType?.transaction)) {
+      const txn = { ...transactionType?.transaction } as SchATransaction;
```
toddlees, line 63: Do you want to get rid of the schema/transaction/title getters at the beginning of this component since we're not using them anymore?

front-end/src/app/shared/components/transaction-type-x2-base/transaction-type-x2-base.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    ) as SchATransaction;
+
+    if (this.transaction?.transaction_type_identifier) {
+      const fieldsToValidate: string[] = this.getFieldsToValidate(this.validateService, this.transactionType);
+      const childFieldsToValidate: string[] = this.getFieldsToValidate(
+        this.childValidateService,
+        this.transactionType?.childTransactionType
+      );
+
+      if (payload.id) {
```
toddlees: I've been developing the transaction pair save to happen on the backend.  as in, we send one payload with both transactions.  this was the reason we looked into atomics on the django api

toddlees (changes requested):

front-end/src/app/shared/resolvers/transaction-type.resolver.ts
```diff
@@ -64,6 +68,12 @@ export class TransactionTypeResolver implements Resolve<TransactionType | undefi
             transaction.transaction_type_identifier
           ) as TransactionType;
           transactionType.transaction = transaction;
+
+          if (transactionType.childTransactionType) {
+            transactionType.childTransactionType.transaction = undefined; // TODO: Need to populate this with what is coming back from the API, ideally it is sending a list of 2 txns
```
toddlees: https://github.com/fecgov/fecfile-web-api/pull/233
this is going to need to be a subsequent call using the parent_transaction_id filter

toddlees (dismissed):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
mjtravers: The getters/setters were getting to be a bit much when tracing the code. They all have been removed.

front-end/src/app/shared/resolvers/transaction-type.resolver.ts
mjtravers: code updated so that parent transaction is in payload

front-end/src/app/shared/components/transaction-type-x2-base/transaction-type-x2-base.component.ts
mjtravers: Ok, payload sent to api contains both transactions

Elaine-Krauss-TCG (approved):

Issue #459, opened 2022-08-12
Allow user to enter Earmark Receipt on Schedule A (Group A)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Earmark Receipt”** transaction on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

NOTE: This transaction requires a "child" (memo) transaction that will be associated with a parent transaction.

 Acceptance Criteria 
 

**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click(open) on the "Contributions From Individuals/Persons" Accordion link  
**Then** I can enter a transaction with the type "Earmark Receipt"
**Then** I will be required to enter the following information **for step one** (unless otherwise noted):

# Contributor 
#
-	**CONTRIBUTOR TYPE** 
**Note: this can be Individual only**
-	**LAST NAME** ( Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)
---

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)
-	**EMPLOYER** 
-	**OCCUPATION**

# Contribution 
#
-	**DATE RECEIVED** (Required)
-	**MEMO ITEM** (Optional)
-	**CONTRIBUTION AMOUNT** (Required) 
**Note:**CONTRIBUTION AMOUNT Can be a negative value
-	**CONTRIBUTION AGGREGATE** (Required, read-only) 

# Additional information 
#
-	**CONTRIBUTION PURPOSE DESCRIPTION** (Read-only, Optional)
Field description in API: RECEIPT DESCRIPTION
Data field: multi-line textbox that MUST SAY “Earmarked through XX” where XX is auto-populated with the name of the committee entered in filer_committee_id_number
-	**NOTE OR MEMO TEXT** (Optional)

**Then** I will be able to:

- Cancel
- Edit fields (cannot edit fields without saving this function not tested in #459 but will test in #460)

(Note: user will not be able to save at this point. They must save with Earmark Memo, which requires ticket #460)

 Dev Task 

[...]

PR #589, opened 2022-09-23
Feature/413 new cmtee contact from transaction

dheitzer:
This is the code review for app tickets 549, 243, 413


dheitzer:
Note that some of the large diffs are me moving the form portion of contact-details to shared component 'contact-form' to be used in both contact-detail and the 'new contact' dialog for lookup.

Elaine-Krauss-TCG (approved):

front-end/src/app/transactions/transaction-group-ag/transaction-group-ag.component.ts
```diff
@@ -44,7 +45,7 @@ export class TransactionGroupAgComponent extends TransactionTypeBaseComponent im
 
   bForm: FormGroup = this.fb.group({});
 
-  bValidateService: ValidateService;
+  // bValidateService: ValidateService;
```
dheitzer, line 48: This was failing the typescript check of needing undefined added to its type so commented out so author could fix.

mjtravers: This will be corrected in #459

PR #610, opened 2022-10-07
Feature/237 aggregation groups

Elaine-Krauss-TCG:
#237 

[Validator PR](https://github.com/fecgov/fecfile-validate/pull/120)
[API PR](https://github.com/fecgov/fecfile-web-api/pull/240)


front-end/src/app/shared/models/scha-transaction.model.ts
```diff
@@ -255,3 +256,11 @@ export const ScheduleATransactionTypeLabels: LabelList = [
     'Earmark Receipt for Headquarters Account (Contribution)',
   ],
 ];
+
+export type AggregationGroup =
```
dheitzer: Do we want to make this an enum instead (similar to ScheduleATransactionGroups for example)?

front-end/src/app/shared/models/transaction-types/EARMARK_RECEIPT.model.ts
```diff
@@ -20,6 +20,7 @@ export class EARMARK_RECEIPT implements TransactionType {
     return SchATransaction.fromJSON({
       form_type: 'SA11AI',
       transaction_type_identifier: ScheduleATransactionTypes.EARMARK_RECEIPT,
+      aggregation_group: "General",
```
dheitzer: If we make AggregationGroup an enum, we can reference it directly here (and aggregation_group will be typed).  Same comment applies for all the similar .fromJSON( calls in this diff.

mjtravers: I agree with David, an enum has a bit more flexibility. Here it would be referenced by something like AggregationGroup.GENERAL rather than "General". That way if the internal ID ever changes or something of the sort then the enum is changed and the references in the code can stay as is. See what was done with the combination of enums and typing for F3XFormTypes and F3xFormType in f3x-summary.model.ts

mjtravers (changes requested):
See inline comments for requested changes

mjtravers (approved):

Issue #276, opened 2022-05-13
Employer/Occupation is required if Contribution Aggregate > $200 (OP1 Transactions)


AureliaKhorsand:
Business Reason

The system will require Employer and Occupation fields if the Contribution Aggregate is greater than $200.

Acceptance Criteria

For the following transactions:

1\. Individual Receipt

  2\. Other Receipt

***If*** Contribution Aggregate is equal to or greater than $200.01

  ***Then*** "Employer" and "Occupation" will be required to save the transaction.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-1195](https://fecgov.atlassian.net/browse/FECFILE-1195)

Elaine-Krauss-TCG:
While the contribution aggregate field is not *yet* being automatically populated, it can still be tested.  I tested this on DEV, by inspecting the field using my browser's dev panel and removing the "readonly" attribute.  This allowed me to set the value of a new transaction's contribution aggregate by hand.

![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/be4e68ac-cbfb-429e-a6a2-fdbf536d399c)
On setting that value on an Individual transaction, the Employer and Occupation fields became required.

In testing the Other Receipt, the front-end *did not* require the Employer and Occupation fields when the Contribution Aggregate was over $200.  The server did require the fields, though, and it properly threw a 400 error (invalid request) when I attempted to save the transaction.

Elaine-Krauss-TCG:
A fix for this error has been developed and has been tested locally.

![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/3f0bb3d6-ca50-4472-9d4a-7428992c8e6d)

MitchellTCG:
@WiseQA 
Please wait on testing this ticket. We need rework on #472 before you can test this

MitchellTCG:
@WiseQA 
Please wait on testing this ticket. We need rework on #472 before you can test this

MitchellTCG:
@WiseQA @AureliaKhorsand 

I can only test independent receipts and other receipts. 

We can't test Earmark Receipts until App#459 and App#460 are pushed to Stage.

WiseQA:
@MitchellTCG is this ticket being broken out and one made for earmark ?   This ticket as we talked needs to be at least 3 points for this sprint to pass what is working

MitchellTCG:
[...]

PR #615, opened 2022-10-14
Every time the contribution aggregate changes, the validation rules f…

Elaine-Krauss-TCG:
#276

This corrects an error where on receipts where there can be multiple values for entity_type, contributor_employer and contributor_occupation are not set as required unless the value for entity_type is updated *after* contribution_aggregate has been set.

This error is solved by rechecking the validation rules for contributor_employer and contributor_occupation whenever the value of contribution_aggregate updates.


mjtravers (approved):
Great find with updateValueAndValidity()

PR #620, opened 2022-10-18
Patch/159

toddlees:
What do you all think of leveraging this mapping of contact fields to labels in the html templates?


mjtravers (approved):
That's good. Much cleaner.

dheitzer:
Yep looks nice.

Issue #492, opened 2022-08-23
Allow user to enter Transfer on Schedule A (Group F)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Transfers”** transaction on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 

**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click(open) on the "Transfer" Accordion link  
**Then** I can enter a transaction with the type “Transfer"
**Then** I will be required to enter the following information (unless otherwise noted):

---

# ~Contributor~ Contact 
#
-	~CONTRIBUTOR~ CONTACT TYPE (Read-only, **INCLUDES LOOKUP**)
-- Committee Only
-	COMMITTEE NAME (Required)
-	COMMITTEE FEC ID (Required)

# Address 
#
-	STREET ADDRESS (Required)
-	APARTMENT, SUITE, ETC. (Optional)
-	CITY (Required)
-	STATE/PROVINCE (Required)
-	ZIP/POSTAL CODE (Required)

# Contribution 
#
-	DATE RECEIVED (Required)
-	MEMO ITEM (Optional)
-	~CONTRIBUTION~ AMOUNT (Required)
-- Note: AMOUNT can contain a negative value.
-	~CONTRIBUTION~ AGGREGATE (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.
<br>

# Additional information 
#
-	CONTRIBUTION PURPOSE DESCRIPTION (Read-only, Optional)
-	NOTE OR MEMO TEXT (A/N-100, Optional)
---
**Then** I will be able to:

- Save & add another
- Save & view all transactions [will go to "Manage your transactions" Table]
- Edit as needed
- Cancel 

**And** the transaction will be itemized upon saving with it's own line item on the *.fec file (SA12). _(Dev note: All SA12 transactions are itemized regardless of amount)._

 Dev Note 

[...]

PR #640, opened 2022-10-24
Feature/492 transfer

toddlees:
#492


front-end/src/app/shared/models/transaction-types/TRANSFER.model.ts
```diff
@@ -0,0 +1,55 @@
...
+      new NavigationControl(
+        NavigationTypes.SAVE_ADD_ANOTHER,
+        'Save & add another Memo',
+        'p-button-warning',
+        TRANSFER.isTransactionExisting,
+        'pi pi-plus'
+      ),
+    ],
+    [new NavigationControl(NavigationTypes.CANCEL_LIST, 'Cancel')],
+    [new NavigationControl(NavigationTypes.SAVE_LIST, 'Save & view all transactions')]
```
Elaine-Krauss-TCG: The new NavigationControl concept is fantastic.  This should cut down on code duplication pretty significantly

Elaine-Krauss-TCG (approved):

Elaine-Krauss-TCG:
Approved, linting and unit tests notwithstanding, of course.

front-end/src/app/shared/models/transaction-navigation-controls.model.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export enum NavigationDestination {
+  LIST,
+  PARENT,
+  ANOTHER,
+  CHILD,
+}
+
+export class NavigationControl extends BaseModel {
```
mjtravers: Does this class need to extend BaseModel? The BaseModel has methods for the handling of the data sent to/from the API. It's meant for passing data back and forth to Django. We could rename BaseModel to BaseDataModel if that would clarify its purpose.

front-end/src/app/shared/interfaces/transaction-type.interface.ts
```diff
@@ -11,9 +13,18 @@ export interface TransactionType {
   isDependentChild: boolean; // When set to true, the parent transaction is used to generate UI form entry page
   title: string;
   schema: JsonSchema; // FEC validation JSON schema
-  transaction: Transaction | undefined;
-  parentTransaction: Transaction | undefined;
-  childTransactionType: TransactionType | undefined;
+  transaction?: Transaction;
+  parentTransaction?: Transaction;
+  childTransactionType?: TransactionType;
+  subTransactionTypes?: ScheduleATransactionTypes[];
```
mjtravers: Would you add an inline comment describing why the subTransactoinTypes[] is needed? It's not so obvious without digging through the code.

front-end/src/app/shared/interfaces/transaction-type.interface.ts
```diff
@@ -11,9 +13,18 @@ export interface TransactionType {
...
   contributionPurposeDescripReadonly(): string; // Dynamically generates the text in the CPD field
   getNewTransaction(): Transaction; // Factory method to create a new Transaction object with default property values for this transaction type
 }
+
+export function isNewTransaction(transaction?: Transaction): boolean {
+  return !transaction?.id;
+}
+export function hasContact(transaction?: Transaction): boolean {
+  return !transaction?.contact;
+}
```
mjtravers: Rather than having these functions defined as separate in the interface file, would the code be better organized if we crated a BaseTransactionModel that contained these methods and SchATransaction model could inherit the methods? BaseTransactionModel would extend BaseModel.

front-end/src/app/shared/models/transaction-types/TRANSFER.model.ts
```diff
@@ -0,0 +1,57 @@
...
+        'Save & add another Memo',
+        'p-button-warning',
+        hasContact,
+        isNewTransaction,
+        'pi pi-plus'
+      ),
+    ],
+    [CANCEL_CONTROL],
+    [SAVE_LIST_CONTROL]
+  );
```
mjtravers, line 43: This is great! Nice job

mjtravers (changes requested):
See comments for some thoughts about code organization. Great abstraction out of the form navigation. Makes me really want to the the form import controls abstracted out now in a big way

front-end/src/app/shared/models/transaction-navigation-controls.model.ts
toddlees: I definitely wasn't sure about this one.  I can just get rid of the extends

front-end/src/app/shared/interfaces/transaction-type.interface.ts
toddlees: on it!

front-end/src/app/shared/interfaces/transaction-type.interface.ts
toddlees: yeah wow that makes a lot more sense

toddlees: So i tried doing this and it got really complicated with not much gain.  I at least moved the static functions to `Transaction` rather than `TransactionType`

dheitzer (approved):

mjtravers (approved):

Issue #403, opened 2022-07-22
System to save Transaction Memo Text as separate TEXT record


AureliaKhorsand:
Business Reason

As a user, I will be able to enter additional information for a transaction in the Memo Text field of a transaction and it will save as a Text record associated with the transaction so that I can have more space for my documentation.

Acceptance Criteria

***Given*** the user is creating or editing a transaction

  ***Then*** they will have the option to enter Memo Text (A/N, 4000)

  ***And*** when they save the transaction, the Memo Text will be saved as a TEXT record associated with that transactions.

  ***And*** the TEXT record row will be saved in the *.fec file directly beneath the transaction row.

### QA Notes ###
The screenshot of the fec file has been replaced here with a file that can be downloaded and opened with MS Notepad: ~~[The .fec generated in testing.](https://app.zenhub.com/files/306118986/ecbb03c8-1865-4449-b781-bff66827a488/download)~~ (Click on this link [new](https://app.zenhub.com/files/306118986/d4738372-716c-4f04-b6ea-28e23671471a/download) one generated after some last minute revisions to .fec generation)  It was renamed to have a .txt extension because Zenhub does not allow the .fec extension for some reason...

Transactions that need testing (as of 10/13/22)

- Individual Receipt (Group A)
- Tribal Receipt (Group D)
- Earmark Receipt (Group A)
- Earmark Memo (Group G)
- Joint Fundraising Transfer (Group E)
- PAC JF Transfer Memo (Group F)
- Offset to Operating Expenditures (Group B)
- Other Receipts (Group C)

DEV per meeting last week, please add screen shot of .fec file in QA notes what needs verified.   Also, if there is an order to the output in .fec file please denote in QA notes for testing purposes.

### DEV Notes ###
- See specs for Text record [here](https://docs.google.com/spreadsheets/d/1SRS_hfYEHQl13aoC-z1xpxubfbKKNMSr/edit#gid=4932567).
- Backend development should put the memo text from the form input into the memo_text Django model to ease the output to the *.fec file
- Ideally, we'll have a foreign key from the memo_text table to the transaction table so the memo_text is linked to its parent transaction. The wrinkle in this is that the foreign key on memo_text should look to ALL transaction tables not just scha_transactions but also future tables such as schb_transactions and the others
[...]

PR #661, opened 2022-11-01
Feature/403 memo text as separate record

Elaine-Krauss-TCG:
#403


Elaine-Krauss-TCG:
I'll happily take any ideas for correcting the one linting error.  I couldn't find a type that would still compile even after quite a few iterations.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -141,6 +145,33 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     this.contactId$.complete();
   }
 
+  retrieveMemoText(formValues: any) {
+    const memo_text = MemoText.fromJSON({
+      text4000: formValues['memo_text_description'],
+      report_id: this.transactionType?.transaction?.report_id,
+      rec_type: '',
```
toddlees:
```
rec_type: 'TEXT',
back_reference_sched_form_name: this.transactionType?.transaction?.form_type
```

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -141,6 +145,33 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
...
+      filer_committee_id_number: this.transactionType?.transaction?.filer_committee_id_number,
+      transaction_id_number: '',
+      back_reference_sched_form_name: '',
+    });
+
+    if (this.transactionType?.transaction?.id) {
+      memo_text.transaction_id = this.transactionType.transaction.id;
```
toddlees: I don't think you need to keep track of the (internal) transaction id for the memo text

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
Elaine-Krauss-TCG: That's transaction_id, a new field, not SchATransaction_id an internal reference.  It's a schedule-agnostic field for the future where we might have many different transaction models all with their own foreign keys to memo_text.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
Elaine-Krauss-TCG: :+1:

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
Elaine-Krauss-TCG: Renamed the field to "transaction_uuid" to reduce confusion

toddlees (approved):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -141,6 +145,38 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     this.contactId$.complete();
   }
 
+  retrieveMemoText(formValues: any) {
```
mjtravers: @Elaine-Krauss-TCG This method seems to be working but makes me uncomfortable because it calls the transactionType object directly from the "this" object which means it is possibly not calling the childTransactionType when appropriate for screens like the EARMARK RECEIPT page with 2 transactions, a parent and child.

I'm thinking it safer to pass the transactionType in the method parameters: retrieveMemoText(formvalues: any, transactionType: TransactionType) and then replace the this.transactionType. with just transactionType. within the body of the method.

mjtravers (changes requested):
@Elaine-Krauss-TCG Take a look at my comment about the retrieveMemoText() method. It appears to be working in my smoke tests but logically it makes me nervous, especially if we add grandchildren down the road.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
Elaine-Krauss-TCG: That's a really good call.  Done.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -141,6 +145,39 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     this.contactId$.complete();
   }
 
+  // prettier-ignore
+  retrieveMemoText(transactionType: TransactionType, formValues: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+    const text = formValues['memo_text_description'];
```
toddlees: Just had this thought and maybe we can push it off till later, but if we're using a memo_text_description input, won't it not allow >100 characters even if we're packaging it up into the memo record when we send it to the api?  I'm basically suggesting, should we not just have a dedicated memo_text input and alway remove the memo_text_description input on all forms?

Elaine-Krauss-TCG: Done.  I named it `memo_text_input` as it's a bit easier to not have the formControlName to be the same as the resolved memo_text object

Elaine-Krauss-TCG:
In smoke-testing every transaction type after the most recent change, I've discovered that the memos for Earmark Receipts are broken with this build: the first memo text gets saved for both created transactions.  This does not happen on DEV, so I need to look into it before this can be merged...

mjtravers (approved):

PR #666, opened 2022-11-02
feature/448 tribal jf transfer memo

dheitzer:
I had to update the transaction-group-d component to support the readonly/const memo item checkbox as well as the "Save & add another" button in this ticket.  I regression tested tribal receipt and also added unit tests for these.


front-end/src/app/transactions/transaction-group-d/transaction-group-d.component.ts
```diff
@@ -30,10 +30,25 @@ export class TransactionGroupDComponent extends TransactionTypeBaseComponent imp
...
+      this.form.get('memo_code')?.setValue(true);
+    }
+  }
+
+  protected memoCodeMustBeTrue(): boolean {
+    // Look at validation schema to determine if the memo_code must be true in all cases.
+    const memoCodeSchema = this.transactionType?.schema.properties['memo_code'];
+    return !!memoCodeSchema?.const;
+  }
+
```
mjtravers, line 51: Let's move this up into the base class since the group f component uses it as well

front-end/src/app/transactions/transaction-group-d/transaction-group-d.component.ts
```diff
@@ -30,10 +30,25 @@ export class TransactionGroupDComponent extends TransactionTypeBaseComponent imp
     'memo_code',
     'memo_text_description',
   ];
+  readOnlyMemo = false;
```
mjtravers, line 33: i think renaming this to readOnlyMemoCode would be more clear what input it is affecting

mjtravers (changes requested):
A couple of small changes requested in the inline comments.

mjtravers (approved):

Issue #626, opened 2022-10-19
Resolve and refactor TransactionType.parentTransaction


mjtravers:
Questions to resolve:

1. Will the parent transaction be updated if included in the serializer when saving a transaction. We don't want that.
2. Without the depth=1 setting in the scha_transaction serializer, will we go into infinite loops. (I think yes) And will grandchildren work with depth=1

After #460 is merged, the parent transaction (if a transaction has one) will be bundled up in the transaction payload returned from the API. This ticket is to refactor TransactionType.parentTransaction to use the new TransactionType.transaction.parent_transaction field in it's place

**Acceptance Criteria**
Transactions that have a parent transaction, like PAC JF TRANSFER MEMO, create and update as they did before the code change.

**QA Notes**
Test to make sure PAC JF TRANSFER and EARMARK RECEIPT can be created, updated, the menu navigates properly, and the transaction shows up in the transactions table with the correct parent transaction_id showing up in the table.
### QA Notes ### 
 1. Will the parent transaction be updated if included in the serializer when saving a transaction. We don't want that.
2. Without the depth=1 setting in the scha_transaction serializer, will we go into infinite loops. (I think yes) And will grandchildren work with depth=1

After #460 is merged, the parent transaction (if a transaction has one) will be bundled up in the transaction payload returned from the API. This ticket is to refactor TransactionType.parentTransaction to use the new TransactionType.transaction.parent_transaction field in it's place

**Acceptance Criteria**
Transactions that have a parent transaction, like PAC JF TRANSFER MEMO, create and update as they did before the code change.

**QA Notes**
Test to make sure PAC JF TRANSFER and EARMARK RECEIPT can be created, updated, the menu navigates properly, and the transaction shows up in the transactions table with the correct parent transaction_id showing up in the table.
### DEV Notes ### 
 null
### Design ### 
 null

mjtravers:
A code change is requested from CR. See the [PR](https://github.com/fecgov/fecfile-web-app/pull/686) comment. Sending back to In Progress.

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA verified PAC JF TRANSFER and EARMARK RECEIPT can be created per refactor of TransactionType.parent Transaction.

QA Review Completed.  Moved to Stage Ready.

PR #686, opened 2022-11-16
Feature/626

toddlees:
#626

### DIFF
--- a/front-end/package-lock.json
+++ b/front-end/package-lock.json
@@ -22,7 +22,7 @@
                 "@popperjs/core": "^2.10.2",
                 "bootstrap": "5.1.3",
                 "class-transformer": "^0.5.1",
-                "fecfile-validate": "https://github.com/fecgov/fecfile-validate#8474a6caa45f33e89512f1b885eca2cd89615701",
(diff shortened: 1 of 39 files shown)


dheitzer:
Along this vein, do we know if Transaction interface parent_transaction_id is still needed? It seems to be redundant with parent_transaction.id from an initial look, but I could be missing something.

Elaine-Krauss-TCG:
I ran the E2E tests on this branch, and everything looks good.

image

One test failed, but it passed 3/3 times on being re-run individually.

Elaine-Krauss-TCG (approved):

mjtravers (changes requested):
Now that the TransactionType interface is defining the transaction and childTransactionType properties as optional with the "?", can we safely remove those properties from the classes implementing the interface that have those 2 properties set to values such as:

transaction = undefined;
transaction?: SchATransaction = undefined;
transaction: SchATransaction | undefined = undefined;
transaction?: Transaction;
childTransactionType = undefined;

Also, do we need the "parent?: SchATransaction;" and "contact = undefined;" properties in PAC_RECOUNT_RECEIPT.model.ts and TRANSFER.model.ts?

If we can remove those property declarations safely, I'd like to merge develop into this branch to pickup the new transaction type models recently merged and then prune all those superfluous property declarations from the transaction type model class definitions.

mjtravers (approved):

PR #750, opened 2022-11-17
feature/696 Earmark Receipt/Memo should only display full form after contact is selected

dheitzer:


dheitzer:
Only change was to wrap the earmark and memo forms in 

`<ng-container *ngIf="transactionType?.transaction?.contact">`

and

`<ng-container *ngIf="transactionType?.childTransactionType?.transaction?.contact">`

respectively, similar to the other groups.

mjtravers (approved):

Issue #525, opened 2022-08-25
Allow user to enter PAC National Party Recount/Legal Proceedings Account on Schedule A (Group F)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”PAC National Party Recount/Legal Proceedings Account”** transaction on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click(open) on the "Other" Accordion link  
**Then** I can enter a transaction with the type "PAC National Party Recount/Legal Proceedings Account"
**Then** I will be required to enter the following information (unless otherwise noted):

---

# Contact 
#
-	**CONTACT TYPE** (Read-only, **INCLUDES LOOKUP**)
-- Committee Only
-	**COMMITTEE NAME** (Required)
-	**COMMITTEE ID** (Required)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Receipt Information 
#
-	**DATE RECEIVED** (Required)
-	**MEMO ITEM** (Optional)
-	**AMOUNT** (Required)
**Note:** AMOUNT can contain a negative value.
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of all contributions that are:
--- From the person (ENTITY) and within "National Party Recount Account" _(Dev/QA Note: Aggregation Group is "NATIONAL_PARTY_RECOUNT_ACCOUNT")_
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# Additional information 
#
-	**CONTRIBUTION PURPOSE DESCRIPTION** (Read-only, system generated)
Data field: multi-line textbox that MUST SAY “Recount/Legal Proceedings Account”
-	**NOTE OR MEMO TEXT** (Optional)

---
**Then** I will be able to:

- Cancel
- Save & add another
- Save & view all transactions [will go to "Manage your transactions" Table]
- Edit as needed

**And if** the Aggregate amount is >= $200.01, 
**Then** the transaction will be itemized upon saving and saved with it's own line item on the *.fec file.

[...]

PR #768, opened 2022-11-23
Feature/525 - Add new schema PAC_NATIONAL_PARTY_RECOUNT_ACCOUNT

mjtravers:
#525


front-end/angular.json
```diff
@@ -207,8 +207,8 @@
           "builder": "@cypress/schematic:cypress",
           "options": {
             "devServerTarget": "fecfile-web:serve",
-            "watch": true,
-            "headless": false
+            "watch": false,
```
dheitzer, line 210: did you mean to update this?

Elaine-Krauss-TCG (approved):

toddlees (approved):

front-end/angular.json
mjtravers: Good catch. Not at the time, but I do want it updated now along with headless: true

Issue #652, opened 2022-10-25
508 Compliance: Set up Lighthouse with Cypress


mjtravers:
Business Reason

As a the product owner, I would like 508 Compliance scan run with our processes to make sure we are always 508 compliant

Dev Notes

Overview:
[https://developer.chrome.com/docs/lighthouse/overview/](https://developer.chrome.com/docs/lighthouse/overview/)
[https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md](https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md)

Cypress & Lighthouse:
[https://glebbahmutov.com/blog/cypress-lighthouse/](https://glebbahmutov.com/blog/cypress-lighthouse/)
[https://www.lambdatest.com/blog/using-cypress-google-lighthouse-performance-testing/](https://www.lambdatest.com/blog/using-cypress-google-lighthouse-performance-testing/)

Acceptance Criteria

Lighthouse is run in conjunction with our Cypress e2e tests and reports are produced for verification and documentation of accessibility.

### QA Notes ###

This is a visual inspection of the screenshots in the comments below.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-970](https://fecgov.atlassian.net/browse/FECFILE-970)

Elaine-Krauss-TCG:
Cypress integration is done.  Due to the design of Lighthouse, it is not able to test any pop-ups, notably including the contact forms.

Elaine-Krauss-TCG:
Lighthouse generates individual reports for each page tested

![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/9553bcda-0940-463a-b111-9b8c411db702)
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/1b8b2aea-da23-46c7-bd25-9c10366560a5)

Elaine-Krauss-TCG:
An actual report file:
[Uploading... transaction-creation-group-AG.html]

Elaine-Krauss-TCG:
The lighthouse testing runs as a cypress test.  Currently, all pages are passing
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/b6686242-0c14-415b-b3de-aec9d9a9f2c6)

Elaine-Krauss-TCG:
[...]

PR #770, opened 2022-11-28
Feature/508 Cypress Lighthouse integration

Elaine-Krauss-TCG:
#652


Elaine-Krauss-TCG:
IMO, the 2 remaining code smells should stay as they keep typescript happy

front-end/cypress/support/commands.ts
```diff
@@ -204,6 +206,31 @@ export function calendarSetValue(calendar: string, dateObj: Date = new Date()) {
   cy.shortWait();
 }
 
+export function runLighthouse(directory: string, filename: string) {
+  cy.lighthouse(
+    {
+      performance: 0,
+      accessibility: 90,
```
dheitzer, line 213: Might be worth a discussion what we want to set this number to? Maybe we could also discuss if we want to set/track any of the other thresholds.

front-end/cypress/e2e/general/lighthouse.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+  cy.deleteAllReports();
+  cy.medWait();
+  cy.visit('/reports');
+  const report = generateReportObject();
+  cy.createReport(report);
+  cy.navigateToTransactionManagement();
+  cy.shortWait();
+}
+
+describe('Generates lighthouse reports for every page', () => {
```
dheitzer, line 18: This is great

front-end/cypress/support/commands.ts
```diff
@@ -204,6 +206,31 @@ export function calendarSetValue(calendar: string, dateObj: Date = new Date()) {
...
+      screenEmulation: {
+        mobile: false,
+        disable: false,
+        width: Cypress.config('viewportWidth'),
+        height: Cypress.config('viewportHeight'),
+        deviceScaleRatio: 1,
+      },
+    }
+  ).then(() => {
+    cy.exec(`mv lighthouse.html cypress/lighthouse/${directory}/${filename}.html`);
```
dheitzer, line 230: I wonder if we should add a timestamp to the filename or directory

front-end/cypress/support/commands.ts
mjtravers: We can discuss later on after it's been running for a bit and revisit the configuration settings.

front-end/cypress/support/commands.ts
mjtravers: We'll be managing the report dates via the CircleCI process so, at least at this point, we can forgo the timestamps.

mjtravers (approved):

Issue #766, opened 2022-11-23
Create Sched B Angular model and service


mjtravers:
Similarly to the SchATransaction model and SchATransactionService, create a SchBTransaction model and SchBTransactionService in the angular shared directory.

***Tasks***
 1\. Create SchBTransaction model. Use SchATransaction and SchB validation schema as a template.
 2\. Refactor SchATransaction and SchBTransaction classes to factor out common code, if any, to a Transaction parent class. If a Transaction parent class is necessary, incorporate and remove the Transaction interface. Create a transaction directory in the shared/services and shared/models directory to hold the pertinent model and service files.
 3\. Create SchBTransactionService class. If it makes sense, refactor SchATransactionService and SchBTransaction service to inherit a parent abstract TransactionService class. (Would generics make this possible and useful?)

***Acceptance Criteria***
 Passing of Angular unit tests for the SchBTransaction and SchBTransactionService classes.

***QA Notes***
 A screenshot of all the unit tests passing without errors is attached below in the comments for visual verification that the acceptance criteria has been met.

### QA Notes ###
**Tasks**
1. Create SchBTransaction model. Use SchATransaction and SchB validation schema as a template.
2. Refactor SchATransaction and SchBTransaction classes to factor out common code, if any, to a Transaction parent class. If a Transaction parent class is necessary, incorporate and remove the Transaction interface. Create a transaction directory in the shared/services and shared/models directory to hold the pertinent model and service files.
3. Create SchBTransactionService class. If it makes sense, refactor SchATransactionService and SchBTransaction service to inherit a parent abstract TransactionService class. (Would generics make this possible and useful?)

**Acceptance Criteria**
Passing of Angular unit tests for the SchBTransaction and SchBTransactionService classes.

**QA Notes**
A screenshot of all the unit tests passing without errors is attached below in the comments for visual verification that the acceptance criteria has been met.

### DEV Notes ###
null

### Design ###
null

[FECFILE-889](https://fecgov.atlassian.net/browse/FECFILE-889)

mjtravers:
Screenshot of unit test report showing that all 429 of the unit tests passed  with 0 failures:
![image.png](https://images.zenhubusercontent.com/61b9fc72d99b6f786e0d798c/ba3bbe29-4ca4-4963-aa4b-f1054f69bf1e)

mjtravers:
Passes CR. Sending to QA.

WiseQA:
[...]

PR #774, opened 2022-11-30
Feature/766 Create new Schedule B model

mjtravers:
#766


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/services/transaction.service.ts
```diff
@@ -1,18 +1,35 @@
...
+import { TransactionType } from '../interfaces/transaction-type.interface';
+
+/**
+ * Given the API endpoint, return the class of the relevent schedule.
+ * @param key URL of root API endpoint
+ * @returns Transaction subclass
+ */
+function getScheduleClass(apiEndpoint: string) {
+  switch (apiEndpoint) {
+    case '/sch-a-transactions':
```
toddlees, line 20: the new app follows this pattern:
`/transactions/` for general endpoints
`/transactions/schedule-a/` for schedule a
`/transactions/schedule-b/` for schedule b

front-end/src/app/shared/models/transaction.model.ts
```diff
@@ -0,0 +1,46 @@
...
+import { MemoText } from './memo-text.model';
+
+export abstract class Transaction extends BaseModel {
+  id: string | undefined;
+
+  // FECFile spec properties
+
+  form_type: string | undefined;
+  filer_committee_id_number: string | undefined;
+  transaction_id: string | null = null; // This is a required field and must exist
```
toddlees: i changed this to `string | undefined` for a couple reasons
- transaction_id is not a required field in the payload because of new transactions
- there was nothing gained by having this be null while the other fields were undefined
- by sending undefined, i was able to improve the transaction_id generation on the backend.

front-end/src/app/shared/models/transaction.model.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // FECFile Online custom properties
+
+  transaction_type_identifier: string | undefined;
+  itemized: boolean | undefined;
+
+  parent_transaction: Transaction | undefined;
+  parent_transaction_id: string | undefined; // Foreign key to the parent transaction db record
```
toddlees, line 20: `parent_transaction_id` had to change to `parent_transaction_object_id` for the ContentType workaround to work.  I'm seeing if there's a way to get around that

front-end/src/app/shared/models/scha-transaction.model.ts
```diff
@@ -1,16 +1,9 @@
...
-import { MemoText } from './memo-text.model';
 
-export class SchATransaction extends BaseModel implements Transaction {
-  id: string | undefined;
-
-  form_type: string | undefined;
-  filer_committee_id_number: string | undefined;
-  transaction_id: string | null = null; // This is a required field and must exist
+export class SchATransaction extends Transaction {
   back_reference_tran_id_number: string | undefined;
```
toddlees, line 7: If you moved `form_type` `filer_committee_id_number` `transaction_id` into the parent class Transaction, should we move these common fields as well? 
`back_reference_tran_id_number` `back_reference_sched_name` `entity_type` `election_code` `election_other_description` `conduit_name` `conduit_street1` `conduit_street2` `conduit_city` `conduit_state` `conduit_zip` `memo_code` `memo_text_description` `reference_to_si_or_sl_system_code_that_identifies_the_account`

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -115,9 +115,9 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
         switchMap(([contribution_date, contactId]) => {
           const aggregation_group: AggregationGroups | undefined = (transactionType?.transaction as SchATransaction)
```
toddlees: what do you think of moving this aggregation_group assignment into `getPreviousTransaction` since it's derived from `transactionType` which you are now passing in.  We'd have to update service function to not make a network call if it's undefined, but maybe that's where the condition should be anyways right?  let the component ask the service "i'd like the previous transaction" and have the service decide, based on the args, how much work it should do to return a value.  It would make this observable flow easier to read for sure

toddlees (dismissed):

front-end/src/app/shared/models/scha-transaction.model.ts
toddlees: oh i bet you looked ahead at other schedules. nm

front-end/src/app/shared/services/transaction.service.ts
mjtravers: Ok, that works. Nice and clear.

front-end/src/app/shared/models/transaction.model.ts
mjtravers: This change has been made.

front-end/src/app/shared/models/transaction.model.ts
mjtravers: This change will have to wait until the API changes for sched B are merged into the develop branch. Will stay as is for this PR.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
mjtravers: Good suggestion. This refactoring has been made for this PR.

Issue #303, opened 2022-05-24
System to auto-populate coverage dates based on report type selection


AureliaKhorsand:
See [this spreadsheet ]([https://docs.google.com/spreadsheets/d/1c4cJvmAChcKl9o8a_5w0Zi0iizUTyxYa/edit#gid=1409996071)as](https://docs.google.com/spreadsheets/d/1c4cJvmAChcKl9o8a_5w0Zi0iizUTyxYa/edit#gid=1409996071)as) a starting point.

Business Reason

As a Committee Administrator, I will be able to view report coverage dates that are populated by the system so that I can create my report with the standard dates or edit the dates if necessary.

Acceptance Criteria

***If*** I create a new report

 ***When*** I select 'Reporting Frequency', 'Time Period', and 'Report Type'

 ***Then*** The system will set the 'Covering Period' 'From' and 'To' dates as follows:

Report type code: M2

 Coverage start Date: 1/1/YYYY (YYYY = current year)

 Coverage End Date: 1/31/YYYY (YYYY = current year)

Report type code: M3

 Coverage start Date: 2/1/YYYY (YYYY = current year)

 Coverage End Date: 2/28/YYYY (YYYY = current year)

 ( ***Note:*** In leap years, the Coverage End Date is 2/29/YYYY)

Report type code: M4

 Coverage start Date: 3/1/YYYY (YYYY = current year)

 Coverage End Date: 3/31/YYYY (YYYY = current year)

Report type code: M5

 Coverage start Date: 4/1/YYYY (YYYY = current year)

 Coverage End Date: 4/30/YYYY (YYYY = current year)

Report type code: M6

 Coverage start Date: 5/1/YYYY (YYYY = current year)

 Coverage End Date: 5/31/YYYY (YYYY = current year)

Report type code: M7

 Coverage start Date: 6/1/YYYY (YYYY = current year)

 Coverage End Date: 6/30/YYYY (YYYY = current year)

Report type code: M8

 Coverage start Date: 7/1/YYYY (YYYY = current year)

 Coverage End Date: 7/31/YYYY (YYYY = current year)

Report type code: M9

 Coverage start Date: 8/1/YYYY (YYYY = current year)

 Coverage End Date: 8/31/YYYY (YYYY = current year)

Report type code: M10

 Coverage start Date: 9/1/YYYY (YYYY = current year)

 Coverage End Date: 9/30/YYYY (YYYY = current year)

Report type code: M11

[...]

PR #775, opened 2022-12-01
Feature/303 dates

toddlees:
#303


front-end/src/app/shared/utils/report-code.utils.ts
```diff
@@ -0,0 +1,227 @@
...
+    this.label = label;
+    this.coverageDatesFunction = coverageDatesFunction;
+  }
+}
+
+function createCoverageFunction(
+  startMonth: number,
+  startDate: number,
+  endMonth: number,
+  endDate: number
```
mjtravers: A small nitpick: I think this would be more clear if startDate and endDate were renamed startDay and endDay. Initially, I saw the variable and assumed it was a Date() until I traced it back to the parameter in the function and saw it was a number

toddlees: I think it's confusing too.  I tried just matching what language [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) uses.  For Date(), `Day` means 'Saturday', 'Friday', etc.  what if I called it `startDayOfMonth`?

mjtravers: Per our discussion, I will go ahead and update with startDayOfMonth and endDayOfMonth

mjtravers (approved):

PR #776, opened 2022-12-06
Feature/761 nightly cypress e2e tests

dheitzer:


.circleci/config.yml
```diff
@@ -99,19 +99,30 @@ jobs:
...
+            docker-compose down
+            DB_DOCKERFILE="Dockerfile-e2e" WORKER_DOCKERFILE="Worker_Dockerfile-e2e" API_DOCKERFILE="Dockerfile-e2e" FECFILE_TEST_DB_NAME="postgres"  DJANGO_SECRET_KEY=${E2E_DJANGO_SECRET_KEY} DATABASE_URL=${E2E_DATABASE_URL} docker-compose up -d
+            docker container run --network container:fecfile-api \
+              docker.io/jwilder/dockerize \
+              -wait http://localhost:8080/api/v1/ \
+              -wait-retry-interval 2s \
+              -timeout 60s; \
+            docker container run -e E2E_BRANCH_NAME=${E2E_BRANCH_NAME} -e CYPRESS_EMAIL=${CYPRESS_EMAIL} -e CYPRESS_COMMITTEE_ID=${CYPRESS_COMMITTEE_ID} -e CYPRESS_PASSWORD=${CYPRESS_PASSWORD} --network container:fecfile-api cypress/browsers:node18.12.0-chrome107 /bin/bash -c '\
+              cd ~; \
+              git clone https://github.com/fecgov/fecfile-web-app.git; \
```
toddlees, line 131: I think line 121-123 can be replaced with a `-checkout`
less inline bash scripting, and it will be it's own circleci step in the job

Elaine-Krauss-TCG (approved):

.circleci/config.yml
```diff
@@ -99,19 +99,30 @@ jobs:
...
+              cd fecfile-web-app/front-end; \
+              git checkout feature/${E2E_BRANCH_NAME}; \
+              npm install; \
```
toddlees, line 134: I'd recommend making this it's own step as well.
```
      - node/install-packages:
          override-ci-command: npm install
          app-dir: ~/project/front-end/
```

.circleci/config.yml
```diff
@@ -99,19 +99,30 @@ jobs:
...
+              node --max_old_space_size=4000 ./node_modules/@angular/cli/bin/ng e2e --headless --watch=false --browser chrome;'\
```
toddlees: running the e2e tasks as it's own step seems good too.

.circleci/config.yml
```diff
@@ -99,19 +99,30 @@ jobs:
...
-          override-ci-command: npm install
-          app-dir: ~/project/front-end/
-
+      - setup_remote_docker:
+          docker_layer_caching: true
       - run:
-          name: run e2e tests
-          command: node --max_old_space_size=4000 ./node_modules/@angular/cli/bin/ng e2e
-          working_directory: ~/project/front-end/
+          name: init fecfile-web-api
```
toddlees: This step does a lot more than init fecfile-web-api.  In a lot of cases (obviously not always), I think more steps means more clarity when we look at what failed or what's slow.  plus, circle ci has a lot of steps already made like `-checkout`

It might be worth splitting these up into `initialize api`, `-checkout`(for the front end), `- node/install-packages` (again for the front end), and `run e2e tests`

toddlees: like here, the api was initialized fine and didn't take an hour, but it looks like it failed and took forever to build
image

toddlees (dismissed):

.circleci/config.yml
dheitzer: @toddlees the trouble is that circleci commands are not available inside the container context (only bash commands)  and I can't really seem to find any workaround for that (if you know of any I am all ears though).  I was still however able to refactor into multiple run steps and will should give us more a more granular view of the run.

mjtravers (approved):

dheitzer:
Comments all addressed.  Some of @toddlees comments couldn't be addressed due to the circleci commands not being available inside the docker containers.

Issue #69, opened 2022-01-18
Allow user to delete a contact ONLY if no associated transactions


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to delete a contact so that I can maintain an up-to-date contact list without any duplicates or contacts created in error.

 Acceptance Criteria 
 

**If** I am a Committee Administrator
**When** I go to Contact Management and/or view a contact
**Then** I will be able to delete that contact _ONLY IF_ there are no transactions associated with it
**And** if there are transactions associated with the contact, the system will inform the user that they cannot delete the contact due to associated transactions.

**Dev Tasks:**
1. refactor old code
2. unit tests
3. documentation
4. logging in the back-end
5. accessible in the UI

![Delete Contacts - Tooltip.png](https://images.zenhubusercontent.com/621cecaf32b9700fedb7b3e6/76fff808-623a-4f56-b64f-a999848dccd7)

### QA Notes ### 
 For this ticket, deleted contacts are permanently deleted, and the user cannot restore them from the front-end. (In the future, deleted contacts will be recoverable. See #88)

Confirmed that there is NO confirmation pop-up.
### DEV Notes ### 
 null
### Design ### 
 null

AureliaKhorsand:
Question for FEC: Should these be recoverable? What happens if the contact is associated with a transaction? Should the user still be allowed to delete it?

lbeaufort:
> Question for FEC: Should these be recoverable? What happens if the contact is associated with a transaction? Should the user still be allowed to delete it?

@AureliaKhorsand I would ultimately defer to usability research on this, but because contacts hold valuable contact information, it would be ideal to make them recoverable. Committees should have additional records, but if a contact has 20 years of contribution data (or 10 different addresses over time) associated with it, it seems more user friendly to make them recoverable.

AureliaKhorsand:
DEV NOTE: There is functionality in the system to delete, but it's unclear on if it's working or not. Need to ensure users can view all transactions for a contact before this can be fully tested.

MitchellTCG:
@mary-tcg 
[...]

PR #777, opened 2022-12-07
Feature/69 Prohibit contact delete if linked transactions

mjtravers:
#69


front-end/src/app/contacts/contact-list/contact-list.component.html
```diff
@@ -77,33 +77,32 @@ <h5 class="m-0">Manage Contacts</h5>
         <th pSortableColumn="occupation" id="occupation-column" role="columnheader">
           Occupation <p-sortIcon field="occupation"></p-sortIcon>
         </th>
-        <th id="buttons" role="columnheader"></th>
+        <th id="buttons" role="columnheader" style="text-align: center">Delete</th>
```
Elaine-Krauss-TCG, line 80: It _might_ be worth using the `.text-center` class from shared/scss/classes/_text.scss instead.  It's a very small thing, though

Elaine-Krauss-TCG (approved):

toddlees (approved):

dheitzer (approved):

front-end/src/app/contacts/contact-list/contact-list.component.html
mjtravers: FYI: That scss directory in shared is legacy styling that is going to be eliminated as soon as the current login page goes away. Also, unfortunately, the style is being used because that was the only way to enforce the centering due to aggressive and contrary parent element styling.

PR #802, opened 2022-12-20
Feature/371 dynamic purpose description updates

Elaine-Krauss-TCG:
#371


front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -372,6 +377,53 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
...
+   *
+   * @param payload
+   *    Takes a SchATransaction because a Transaction does not have the
+   *    contributor_organization_name or contribution_purpose_description
+   *    fields.
+   * @returns
+   *    An array of Transaction objects whose contribution_purpose_descriptions
+   *    have been re-generated to account for changes to their parent
+   */
+  updateChildPurposeDescriptions(payload: SchATransaction): Transaction[] {
```
toddlees: I'm finding that this doesn't work when i change the name of the committee name of the parent transaction, because the user hasn't accepted the contact change yet at this point

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
toddlees: Disregard! this is working

toddlees (approved):

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -195,6 +196,10 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     });
     payload.contact_id = payload.contact?.id;
 
+    if (payload.children) {
+      payload.children = this.updateChildPurposeDescriptions(payload as SchATransaction);
```
mjtravers: @Elaine-Krauss-TCG  Since this functionality is exclusive the the contribution_purpose_description of the Schedule A transaction, the updateChildPurposeDescription method (and renamed "updateChildren()") should be moved to the SchATransaction model and called from there. To call it, we could generalize a new method on the Transaction class called "updateChildren()" that does nothing in the Transaction class but then is overridden in the ScheduleATransaction class with the method below.

mjtravers (changes requested):
@Elaine-Krauss-TCG See comment in code for renaming and relocating the updateChildPurposeDescriptions() method.

mjtravers (approved):

PR #806, opened 2022-12-21
feature/488 Unregistered Receipt from Person

dheitzer:
I don't currently love how I'm toggling (OPTIONAL) after the contribution purpose description using:

`purposeDescriptionUserInputRequired`

I feel like there could be a cleaner way to integrate this with (SYSTEM GENERATED).  If anyone has a better idea, I'm all ears.


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/interfaces/transaction-type.interface.ts
```diff
@@ -19,5 +19,6 @@ export interface TransactionType {
   subTransactionTypes?: ScheduleATransactionTypes[] | ScheduleBTransactionTypes[]; // TransactionTypes to choose from when creating a sub transaction
   navigationControls?: TransactionNavigationControls;
   generatePurposeDescription?(): string; // Dynamically generates the text in the CPD or EPD field
+  purposeDescriptionUserInputRequired?: boolean;
```
mjtravers: @dheitzer Is this flag somewhat redundant with the isDescriptionSystemGenerated() method? Is there a nuance I'm missing as why that method can't me used instead to determine if a user can edit the Contribution Purpose Description field?

mjtravers (changes requested):
See comment about flag

front-end/src/app/shared/interfaces/transaction-type.interface.ts
toddlees:
```
     system generated
           /      \
         yes       no
                  /   \
          optional      required
        and editable.  and editable
```

It does need to be separate, but rather than a property, we could just have a common function that determines it's value from the spec rather than the class.  that means one less configuration to implement in the typescript

toddlees: this could be implemented in transactionbasecomponent.ts
```
  isDescriptionRequired(transactionType?: TransactionType): boolean {
    return transactionType?.schema.required.includes('contribution_purpose_descrip') || false;
  }

Comment (dheitzer) [reply front-end/src/app/shared/interfaces/transaction-type.interface.ts]: @mjtravers the key usage in the code is 

```
{{ descriptionIsSystemGenerated ? '(SYSTEM-GENERATED)' : 
          descriptionRequiresUserInput ? '' : '(OPTIONAL)' }}</label
```

Comment (dheitzer) [reply front-end/src/app/shared/interfaces/transaction-type.interface.ts]: @toddlees some suggested avoiding [functions in a templates](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496) so we might want to make this an initialized property?

File: front-end/src/app/shared/models/transaction-types/SchaTransactionType.model.ts [hunk 2]
```diff
@@ -0,0 +1,12 @@
+import { TransactionType } from '../../interfaces/transaction-type.interface';
+
+export abstract class SchaTransactionType extends TransactionType {
+  override generatePurposeDescriptionLabel?(): string {
+    if (this.generatePurposeDescription !== undefined) {
+      return '(SYSTEM-GENERATED)';
+    } else if (!this.purposeDescriptionUserInputRequired) {
```
toddlees: I thought the point of adding this class was to check the schema for whether or not the description is required
If we don't do that, we still have to re-configure `purposeDescriptionUserInputRequired` in every model
```
if (this.generatePurposeDescription) {
  return '(SYSTEM-GENERATED)';
} else if (transactionType?.schema.required.includes('contribution_purpose_descrip')) {
  return '';
}
return '(OPTIONAL)';

toddlees (changes requested):

Elaine-Krauss-TCG (approved):

front-end/src/app/shared/models/transaction-types/SchaTransactionType.model.ts
dheitzer: @toddlees yes I forgot to reach into the schema while focused on the x2-base fix.  Went ahead and added that and pushed it.  Thanks for the snippet.

toddlees (approved):

mjtravers (approved):

Issue #623, opened 2022-10-19
System to default to "Organization" contact type for Offsets to Operating Expenditures


AureliaKhorsand:
Requirement from Ryan Lanz during stakeholder interviews in late August 2022.

Business Reason

As a user, I will be able to create new Offsets to Operating Expenditure transactions that are defaulted to "Organization" contact type so that I can reduce the number of clicks needed to add this transaction type.

Acceptance Criteria

***Given*** the transaction type "Offsets to Operating Expenditures" (Group B)

  ***When*** the user creates a new transaction

  ***Then*** the contact type will default to "Organization".

_Note: Other contact types available are Individual and Committee._

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-993](https://fecgov.atlassian.net/browse/FECFILE-993)

Elaine-Krauss-TCG:
Offset to Opex presents "Organization" as the default contact type
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/2235edfd-bea6-4ff1-aa23-b20fec13b1b7)

Elaine-Krauss-TCG:
Devs can now override the contact type options for a specific transaction type.  (See line #24)  The first type in the list is the default option presented to the user when a new transaction form is opened.
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/94fbee82-87c7-4ce8-90c3-ba7fc8688ef0)

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified "Organization" is the default contact type for Offsets to Operating Expenditures (see screenshot below).

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/7e82cfa8-62ad-40d1-bd7b-8a3d2f0c09be)

QA Review Completed.   Moved to Stage Ready.

Changed naming from **Offset** to Operating Expenditure to **Offsets** to Operating Expenditure in acceptance criteria to be consistent what transaction page is showing.

PR #813, opened 2023-01-03
Feature/623 offset to opex defaults to organization

Elaine-Krauss-TCG:
#623


front-end/src/app/shared/models/transaction-types/OFFSET_TO_OPERATING_EXPENDITURES.model.ts
```diff
@@ -13,13 +13,15 @@ import {
...
   TransactionNavigationControls,
 } from '../transaction-navigation-controls.model';
+import { ContactTypes } from '../contact.model';
 
 export class OFFSET_TO_OPERATING_EXPENDITURES implements TransactionType {
   scheduleId = 'A';
   componentGroupId = 'B';
   isDependentChild = false;
   title = LabelUtils.get(ScheduleATransactionTypeLabels, ScheduleATransactionTypes.OFFSET_TO_OPERATING_EXPENDITURES);
   schema = schema;
```
toddlees, line 23: This is a great direction to go!  I'm a big fan of having transactionType-level configurations in the transactionType classes

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -63,6 +63,11 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
     validateService.formValidatorSchema = transactionType?.schema;
     validateService.formValidatorForm = form;
 
+    // Override contact type options if present in transactionType
+    if (transactionType && transactionType.contactTypeOptions) {
+      this.contactTypeOptions = this.getContactTypeOptions(transactionType.contactTypeOptions);
```
toddlees: is this essentially equivalent to:
```
    if (transactionType && transactionType.contactTypeOptions) {
      this.contactTypeOptions = transactionType.contactTypeOptions.map((type: ContactTypes) => ({
        name: LabelUtils.get(ContactTypeLabels, type),
        code: type,
      }));
    }

Comment (toddlees) [reply front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts]: I could see us using this pattern elsewhere when we want a PrimeOptions list given a list of 'code's and one of our homebrew LabelLists

maybe in LabelUtils we could add

```

  public static getPrimeOptions(labelArrays: LabelList, options: string[]) {
    return options.map((type: string) => ({
      name: LabelUtils.get(labelArrays, type),
      code: type,
    }));
  }

mjtravers: I agree that the getContactTypeOptions() code would be good to generalize out into the LabelUtils class.

Elaine-Krauss-TCG: I like this idea a lot.  I'll take a look at it!

Elaine-Krauss-TCG: https://github.com/fecgov/fecfile-web-app/pull/813/files#diff-5bf0fd4e8831d3a8ff3905e7c9a0a450bbeb72fb77c6cd9c7438a604a6925517R55

LabelUtil's getPrimeOptions was updated to allow you to optionally specify the values to use.

toddlees (approved):

dheitzer (approved):

mjtravers (approved):

PR #844, opened 2023-02-01
 e2e patch

toddlees:
leaning on https://docs.cypress.io/guides/core-concepts/retry-ability
use a get that will have to retry until conditions are met so we aren't guessing with wait times
also, narrows tr query to not match thead tr's.


toddlees:
image

mjtravers (approved):

PR #850, opened 2023-02-03
remove save and add another memo on edit

toddlees:


Elaine-Krauss-TCG (approved):

mjtravers (approved):
This is good. Handling navigation by an event is more flexible.

Issue #836, opened 2023-01-25
Update Contribution Purpose Description label for Schedule A Transactions


AureliaKhorsand:
***Task:*** "Contribution Purpose Description" should read "Purpose of Receipt" for Schedule A Transactions.

### QA Notes ###
List of Schedule A transaction groups for testing:
- Individual Receipt (Group A) Note:  Optional
- Offsets to Operating Expenditures (Group B) Note:  Optional
- Other Receipts (Group C) Note:  Optional (Updated from Other Receipt to Other Receipts in QA Notes)
- Tribal Receipt (Group D) Note:  System Generated
- Joint Fundraising Transfer (Group E) Note:  System Generated
- PAC Joint Fundraising Transfer Memo (Group F) Note:  System Generated
- Earmark Receipt/Memo (Group G) Note:  System Generated

Updated Note to QA this field is "Optional" for all above transactions

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-871](https://fecgov.atlassian.net/browse/FECFILE-871)

Elaine-Krauss-TCG:
A new system has been put in place where the purpose description field's title takes on different values depending on the schedule.  Alternatively, this can be overridden by providing a specific title to use.

Example from an Individual Receipt:
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/11334410-53d8-497d-97b8-b425c85885d2)

Elaine-Krauss-TCG:
This has been implemented for the following transaction groups:

- [x] Individual Receipt (Group A)
- [x] Offsets to Operating Expenditures (Group B)
- [x] Other Receipt (Group C)
- [x] Tribal Receipt (Group D)
- [x] Joint Fundraising Transfer (Group E)
- [x] PAC Joint Fundraising Transfer Memo (Group F)
- [x] Earmark Receipt/Memo (Group AG)

Elaine-Krauss-TCG:
Reimplemented to work with the Schedule B refactor
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/12aec081-f1e9-4009-8028-e348fb27dcda)

mjtravers:
Passes CR. Ready for QA.

WiseQA:
QA Review Verified the following transactions Contribution Purpose Description field label has been changed to "Purpose of Receipt (Optional)".

[...]

PR #851, opened 2023-02-03
Revised title for Purpose Description field

Elaine-Krauss-TCG:
A new system has been put in place where the purpose description field's title takes on different values depending on the schedule. Alternatively, this can be overridden by providing a specific title to use.

#836

### DIFF
--- a/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
+++ b/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from '@angular/core';
+import { Component, OnInit, Input } from '@angular/core';
 import { BaseInputComponent } from '../base-input.component';
 
 @Component({
@@ -6,7 +6,7 @@ import { BaseInputComponent } from '../base-input.component';
   templateUrl: './additional-info-input.component.html',
   styleUrls: ['./additional-info-input.component.scss'],
 })
-export class AdditionalInfoInputComponent extends BaseInputComponent {
+export class AdditionalInfoInputComponent extends BaseInputComponent implements OnInit {
   @Input() descriptionIsSystemGenerated = false;
   @Input() contributionPurposeDescriptionLabel = '';
   @Input() purposeDescriptionLabelNotice?: string;
--- a/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
+++ b/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
@@ -20,6 +20,7 @@ export class AdditionalInfoInputComponent extends BaseInputComponent implements 
   }
 
   public getPurposeFieldTitle() {
+    //This is a switch because there will *eventually* be many more cases
     switch (this.schedule) {
       case 'A':
         return 'Purpose of Receipt';
--- a/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
+++ b/front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
@@ -11,7 +11,7 @@ export class AdditionalInfoInputComponent extends BaseInputComponent implements 
   @Input() contributionPurposeDescriptionLabel = '';
   @Input() purposeDescriptionLabelNotice?: string;
   @Input() purposeFieldTitle?: string; //Specifies a specific title for the CPD field
-  @Input() schedule?: 'A' | 'B'; //Specifies a schedule so that a default CPD title can be used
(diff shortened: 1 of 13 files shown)


front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
```diff
@@ -1,13 +1,30 @@
...
   templateUrl: './additional-info-input.component.html',
   styleUrls: ['./additional-info-input.component.scss'],
 })
-export class AdditionalInfoInputComponent extends BaseInputComponent {
+export class AdditionalInfoInputComponent extends BaseInputComponent implements OnInit {
   @Input() descriptionIsSystemGenerated = false;
   @Input() contributionPurposeDescriptionLabel = '';
   @Input() purposeDescriptionLabelNotice?: string;
+  @Input() purposeFieldTitle?: string; //Specifies a specific title for the CPD field
+  @Input() schedule?: 'A' | 'B'; //Specifies a schedule so that a default CPD title can be used
```
toddlees: This will depend heavily on how @mjtravers decides to implement schedule B.  consider that the form field for the purpose description is `contribution_purpose_descrip`, a ScheduleA-only field.  Matt might go one route, which is to use a different component for `expenditure_purpose_descrip`, in that scenario, this component would never be used for ScheduleB.  More likely, he'll make the component use a different form field depending on the schedule.  If that's the case, you'll want your condition to depend on how he signals the schedule, so we don't have two different data points describing what schedule the component is used for.

front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
```diff
@@ -1,13 +1,30 @@
...
+    }
+  }
+
+  public getPurposeFieldTitle() {
+    //This is a switch because there will *eventually* be many more cases
+    switch (this.schedule) {
+      case 'A':
+        return 'Purpose of Receipt';
+    }
+    return 'Contribution Purpose Description';
```
toddlees, line 28: This return statement will never be hit.

front-end/src/app/transactions/transaction-group-a/transaction-group-a.component.html
```diff
@@ -39,6 +39,7 @@ <h3>Additional Information</h3>
         [form]="form"
         [formSubmitted]="formSubmitted"
         [descriptionIsSystemGenerated]="isDescriptionSystemGenerated(transactionType)"
+        [schedule]="'A'"
```
toddlees: maybe you could set this with `transactionType.scheduleId`

front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
```diff
@@ -1,13 +1,30 @@
...
+    if (!this.purposeFieldTitle) {
+      this.purposeFieldTitle = this.getPurposeFieldTitle();
+    }
+  }
+
+  public getPurposeFieldTitle() {
+    //This is a switch because there will *eventually* be many more cases
+    switch (this.schedule) {
+      case 'A':
+        return 'Purpose of Receipt';
```
toddlees, line 26: I think this needs to be all-caps

toddlees (changes requested):

front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
Elaine-Krauss-TCG: It's not listed as all-caps in the ticket, but it's well worth double-checking

front-end/src/app/transactions/transaction-group-a/transaction-group-a.component.html
Elaine-Krauss-TCG: :+1:

front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
toddlees: ah you did the right thing in following it to the tee!  good to ask too.

Elaine-Krauss-TCG: With so many people out, this'll have to wait for Monday so that I can double check this, but that's no big loss.

mjtravers (changes requested):
@Elaine-Krauss-TCG I believe this ticket will be OBE with how I am integrating the management of the Schedules into the group form templates. Don't merge this branch into the develop branch. Let's sit on this ticket and PR for the moment.

mjtravers:
This PR was replaced by https://github.com/fecgov/fecfile-web-app/pull/882

PR #858, opened 2023-02-09
Feature/206 test committee accounts for filing

dheitzer:
see api pr description.


front-end/src/app/shared/services/fec-api.service.ts
```diff
@@ -36,14 +39,10 @@ export class FecApiService {
     if (!committeeId) {
       throw new Error('No Committee Id provided.');
     }
-    const headers = this.getHeaders();
```
toddlees, line 60: Weren't we planning on making this configurable.  We don't want to eliminate the actual code that works, but point to something else when it's configured to mock right?

front-end/src/app/shared/services/fec-api.service.ts
```diff
@@ -60,48 +59,9 @@ export class FecApiService {
...
```
toddlees, line 84: We still need this code right?  we just need to configure the endpoints it hits to be mocks when the system is configured to test

toddlees (dismissed):

dheitzer:
Made the changes discussed in our dev meeting for api:
1) using an environment variable to determine committee id data to override
2) using a local json file to override the committee data for specified committee ids

front-end/src/app/shared/services/fec-api.service.ts
dheitzer: made changes per our dev meeting for the new approach to overriding committee data in the api.  I believe this is OBE now.

front-end/src/app/shared/services/fec-api.service.ts
dheitzer: made changes per our dev meeting for the new approach to overriding committee data in the api.  I believe this is OBE now.

mjtravers (approved):

PR #884, opened 2023-02-22
Feature/402 report action

toddlees:


mjtravers (approved):
Very nice generalization of the actions into the RowAction class

PR #901, opened 2023-03-03
Feature/684 payroll memo for operating expenditure

dheitzer:


toddlees (approved):
added sub transaction to parent transaction and added transaction to transactionTypeClasses

Issue #583, opened 2022-09-22
Remove extraneous left-hand nav bar


MitchellTCG:
Businuess Case

As a UX designer, I would like the extraneous 300px wide blue nav bar from pages that have a workflow nav so the user can regain some screen real estate.

Acceptance Criteria

Please remove the 300px Blue Nav from the following pages:

1.  1\. Dashboard

       #

       A. Dashboard Page

*   [https://fecfile-web-app-prod.app.cloud.gov/dashboard](https://fecfile-web-app-prod.app.cloud.gov/dashboard)

       ***Change from:***

       ![Screen Shot 2022-10-31 at 8.31.44 PM.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/145f7dd6-52dd-423c-b62a-37978006eb83](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/145f7dd6-52dd-423c-b62a-37978006eb83)) 

       ***Change to:***

       ![Screen Shot 2022-10-31 at 8.31.30 PM.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7e5a2499-d4e9-47c5-83cc-b39f8f8eefaa](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7e5a2499-d4e9-47c5-83cc-b39f8f8eefaa))

1.  2\. Reports

       #

       A. Manage Reports Page

*   [https://fecfile-web-app-prod.app.cloud.gov/reports](https://fecfile-web-app-prod.app.cloud.gov/reports)

       ***Change from:***

       ![Screen Shot 2022-10-31 at 8.33.11 PM.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ba8cbb7f-e642-4695-bf37-178cfc1cd5f9](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ba8cbb7f-e642-4695-bf37-178cfc1cd5f9))

       ***Change to:***

       ![Screen Shot 2022-10-31 at 8.33.02 PM.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/27fe6a3e-26f1-4ff1-9d89-17485644b27e](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/27fe6a3e-26f1-4ff1-9d89-17485644b27e))

B. Create a new report

[...]

PR #909, opened 2023-03-09
Feature/583 remove extraneous navbar

Elaine-Krauss-TCG:
#583

### DIFF
--- a/front-end/src/app/layout/layout.component.ts
+++ b/front-end/src/app/layout/layout.component.ts
@@ -1,23 +1,74 @@
-import { Component, OnInit } from '@angular/core';
-import { Observable } from 'rxjs';
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Observable, Subject, takeUntil } from 'rxjs';
 import { Store } from '@ngrx/store';
 import { selectCommitteeAccount } from '../store/committee-account.selectors';
 import { selectSpinnerStatus } from '../store/spinner.selectors';
 import { CommitteeAccount } from 'app/shared/models/committee-account.model';
+import { NavigationEnd, Router, Event } from '@angular/router';
 
 @Component({
   selector: 'app-layout',
   templateUrl: './layout.component.html',
-  styleUrls: ['./layout.component.scss']
+  styleUrls: ['./layout.component.scss'],
 })
-export class LayoutComponent implements OnInit {
+export class LayoutComponent implements OnInit, OnDestroy {
   committeeAccount$: Observable<CommitteeAccount> | undefined;
   progressBarVisible$: Observable<{ spinnerOn: boolean }> | undefined;
+  sidebarVisible = false;
+  private destroy$ = new Subject<boolean>();
 
-  constructor(private store: Store) {}
+  // The order of the url regular expressions listed inthe urlMatch array is important
+  // because the order determines the expanded menu item group in the panal menu:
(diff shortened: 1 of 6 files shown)


front-end/src/app/layout/layout.component.ts
```diff
@@ -1,23 +1,73 @@
...
+  styleUrls: ['./layout.component.scss'],
 })
-export class LayoutComponent implements OnInit {
+export class LayoutComponent implements OnInit, OnDestroy {
   committeeAccount$: Observable<CommitteeAccount> | undefined;
   progressBarVisible$: Observable<{ spinnerOn: boolean }> | undefined;
+  sidebarVisible = false;
+  private destroy$ = new Subject<boolean>();
 
-  constructor(private store: Store) {}
```
toddlees, line 17: For this component and for the menu component, it would be cool if we assigned `sidebarState` with nested `expandedSection` in with a resolver.  The presence/definition of the state would indicate it's visible. We could apply the resolver in *-routing.module.ts like so:
```

  {
    path: '',
    title: 'Manage Reports',
    component: ReportListComponent,
    pathMatch: 'full',
  }, // No sidebar resolver, no sidebar
  ...
  {
    path: 'f3x/summary/:reportId',
    title: 'View summary page',
    component: ReportSummaryComponent,
    resolve: { report: ReportResolver, sidebarState: SidebarResolver(MenuSections.REVIEW) },
  }, // sidebar resolver means it shows, MenuSection enum tells it what section to open
```

That way, there'd be no regexes and the configurations would be applied once per route, at the route config, instead again every time and everywhere we want to check.  Less router subscriptions and event handlers too

Not a blocker, just a thought seeing as we had to duplicate the url regexes

front-end/src/app/layout/sidebar/menu-report/menu-report.component.ts
```diff
@@ -20,7 +20,6 @@ export class MenuReportComponent implements OnInit, OnDestroy {
   currentReportId: string | undefined;
   currentReportTimestamp: number | undefined;
   items: MenuItem[] = [];
-  showMenu = false;
```
toddlees, line 23: I'm a big fan of moving the visibility of the menu OUT of the menu. 👍 👍

toddlees (approved):

front-end/src/app/layout/layout.component.ts
toddlees: @Elaine-Krauss-TCG @mjtravers Thoughts?

dheitzer (approved):

front-end/src/app/layout/layout.component.ts
mjtravers: I really like Todd's idea of moving the logic into a resolver that then is attached (or not) to the routing. One thing to take into account is that there are 2 things to hide/show here: (1) is the left had blue area which is basically the FECFile Online header and blue area (aka the sidebar component) and (2) the Report Progress menu (aka menu-report component). It could be in the future that the sidebar is visible but a menu component other than the menu-report component will be visible in the sidebar. The resolver solution should be flexible enough to account for different components being inserted into the sidebar component (we do something similar in the transaction-container component)

mjtravers (changes requested):
Sending back to refactor using a resolver to hide/show the sidebar per Todd's suggestion

PR #928, opened 2023-03-17
Feature/583b remove extraneous navbar with resolver

Elaine-Krauss-TCG:
#583


Elaine-Krauss-TCG:
I don't have a good idea for how to get at expanding unit testing here.  Any ideas are welcome

front-end/src/app/layout/sidebar/menu-report/menu-report.component.ts
```diff
@@ -19,34 +20,18 @@ export class MenuReportComponent implements OnInit, OnDestroy {
...
+  expandedSection: 'Transactions' | 'Review' | 'Submission' | 'None' = 'None';
   items: MenuItem[] = [];
-  showMenu = false;
   f3xFormTypeLabels: LabelList = F3xFormTypeLabels;
   reportIsEditableFlag = false;
   cashOnHand: CashOnHand = {
     report_id: undefined,
     value: undefined,
   };
+  sidebarState: ReportSidebarState | undefined = undefined;
```
dheitzer: minor: no need for `= undefined` here

front-end/src/app/layout/sidebar/menu-report/menu-report.component.ts
```diff
@@ -161,23 +146,13 @@ export class MenuReportComponent implements OnInit, OnDestroy {
...
-   * group of menu items is selected.
-   *
-   * @param urlMatch {RegExp{}} - List of url regular expressions to compare to current router url
-   * @param url {string} - Current url in browser address bar
-   * @returns {boolean} - True is url matches one of the matchUrl regular expressions
-   */
-  isActive(urlMatch: RegExp[], url: string): boolean {
-    return urlMatch.reduce((prev: boolean, regex: RegExp) => prev || regex.test(url), false);
+      // This fixes a bug where the sidebarState is undefined when first navigating to a report
+      if (this.sidebarState == undefined) this.items[0].expanded = true;
```
dheitzer, line 154: could you just initialize `items[0].expanded = true` at the top of the class?

Elaine-Krauss-TCG: the items are defined immediately prior to this, so they can't be set at the top of the class, unfortunately.  The first item also can't be initialized to true without needing to be set to false again if another section is supposed to be expanded.  Now that the component will only be loaded when there's a relevant report, there's room to rewrite the whole component, but I think that it would be better to wait on that until there's a second sidebar to implement that could share code with it.

mjtravers (approved):

Issue #88, opened 2022-01-31
Allow user to recover a deleted contact


AureliaKhorsand:
Business Reason 
 

As a Committee Administrator, I will be able to recover a deleted contact so that I can restore it to Contact Management if it was deleted in error, or it is needed again.

 Acceptance Criteria 
 

**If** I am a user
**And** I go to Contact Management
**Then** I can see an option to "Restore deleted contact(s)" _(and this link will be disabled if there are no deleted contacts)_
**And when** I open to "Restore deleted contact(s)", I will view deleted contacts (including pagination)
**Then** I will be able to restore one or multiple deleted contact(s) to the Contact Management list.

 Dev Tasks:
1. refactor old code
2. unit tests
3. documentation
4. logging in the back-end
5. accessible in the UI

# Wireframes 
#
![88  - Restore deleted contacts link.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/aac63eff-6595-44f2-adef-f37abbc09052)

![88  - Restore deleted contacts modal.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/673e088c-7d9d-40d7-b324-2cd1f8e0532c)

### QA Notes ### 
 Restored contacts should appear back in the Contact Management list in the default Contact sort order (alphabetical).
### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
@AureliaKhorsand @mary-tcg @JonellaCulmer 
Updated

MitchellTCG:
@AureliaKhorsand @WiseQA  added wireframe with pagination

toddlees:
@MitchellTCG @AureliaKhorsand Should the "Restore selected" button be disable until at least one contact is selected?

Do we need the parenthetical plural for contacts if "Restore deleted contacts" is still pretty clearly the place to restore a single contact?

AureliaKhorsand:
@toddlees executive decisions while Mitchell is out: Yes, disable the restore button unless one contact is selected (AC is updated). Keep the parenthetical plural as is in the wireframe and we can change it later if needed.

toddlees:
Thanks!

mjtravers:
Passes CR. Sending to QA.

WiseQA:
@toddlees @mjtravers so all but my recent contacts the "delete" trash can is not accessible.   Is there a reason ?

[...]

PR #962, opened 2023-04-06
deleted contact service and

toddlees:
#88


front-end/src/app/contacts/deleted-contact-dialog/deleted-contact-dialog.component.scss
Elaine-Krauss-TCG, line 1: This file being blank is causing a code smell.  It's a really minor thing, but it might be worthwhile to just get rid of the file

Elaine-Krauss-TCG (approved):

Issue #945, opened 2023-03-27
Create Angular model for Schedule C


mjtravers:
Following the same pattern as was created for the Schedule A and Schedule B, create a Schedule C class to manage the data for a Schedule C record.

DEV Tasks:

*   Create a SchCTransaction class that extends the Transaction class
*   Create a ScheduleCTransactionGroups enum containing groupings for the "Add transacation" expand/collapse page
*   Create a ScheduleCTransactionsGroupsType type
*   Create a ScheduleCTransactionTypes enum containing the transaction unique identifiers
*   Create a ScheduleCTransactionTypeLabels LabelList containing the UI title for each transaction type
*   Update the ScheduleTransaction, TransactionTypes, and TransactionGroupTypes type definitions in the transactions.model.ts file

### QA Notes ###

There will be a unit test that can be reviewed verifying the correctness of the Schedule C model code. A screenshot can be provided.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-831](https://fecgov.atlassian.net/browse/FECFILE-831)

mjtravers:
Screenshot of unit tests running successfully:
![image.png](https://images.zenhubusercontent.com/61b9fc72d99b6f786e0d798c/1abde8e1-a12d-4096-9a8e-6318270fbfbc)

mjtravers:
Passes CR. Ready for QA.

WiseQA:
QA review of visual inspection of Unit Tests as provided above.

QA Review Completed.    Moved to Stage Ready.

PR #965, opened 2023-04-06
Feature/945 Add schedule C model

mjtravers:
#945

### DIFF
--- a/front-end/src/app/shared/models/schc-transaction.model.ts
+++ b/front-end/src/app/shared/models/schc-transaction.model.ts
@@ -73,13 +73,9 @@ export type ScheduleCTransactionGroupsType = ScheduleCTransactionGroups.SCHEDULE
 export enum ScheduleCTransactionTypes {
   LOANS_RECEIVED_FROM_INDIVIDUAL = 'LOANS_RECEIVED_FROM_INDIVIDUAL',
   LOANS_RECEIVED_FROM_BANK = 'LOANS_RECEIVED_FROM_BANK',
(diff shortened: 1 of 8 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/models/schc-transaction.model.ts
```diff
@@ -0,0 +1,85 @@
...
+export enum ScheduleCTransactionGroups {
+  SCHEDULE_C_AND_C1 = 'Schedule C and C-1',
+}
+
+export type ScheduleCTransactionGroupsType = ScheduleCTransactionGroups.SCHEDULE_C_AND_C1;
+
+export enum ScheduleCTransactionTypes {
+  LOANS_RECEIVED_FROM_INDIVIDUAL = 'LOANS_RECEIVED_FROM_INDIVIDUAL',
+  LOANS_RECEIVED_FROM_BANK = 'LOANS_RECEIVED_FROM_BANK',
+  C1_LOAN_AGREEMENT = 'C1_LOAN_AGREEMENT',
```
toddlees: C1 and C2 are different specs/models than C https://docs.google.com/spreadsheets/d/18TtYCHWjj0gbsubJXQTbALEkbK_t93t0/edit#gid=1726907415

toddlees (dismissed):

front-end/src/app/shared/models/schc-transaction.model.ts
mjtravers: I went back and forth as to whether to add them into a generic Schedule C type as the form fields are quite different but they might be on the same picker. Not having them here makes sense.

Issue #721, opened 2022-11-16
Allow user to enter Other Committee Refund - National Party Recount/Legal Proceedings Account on Schedule B (Group E)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Other Committee Refund - National Party Recount/Legal Proceedings Account”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click (open) on the "Other Expenditures" Accordion link  
**Then** I can enter a transaction with the type “Other Committee Refund - National Party Recount/Legal Proceedings Account"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE**
-Committee
-	**COMMITTEE NAME** (Required)
-	**COMMITTEE ID** (Required)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional)
-	**AMOUNT** (Required, can be a negative value)

# Additional information 
#
-	**PURPOSE OF DISBURSEMENT** (Required)
-Text: 'Recount/Legal Proceedings Account: Refund'
-	**NOTE OR MEMO TEXT** (Optional)
-	**CATEGORY CODE** (Optional)

---
**Then** I will be able to:
- Cancel
- Save 
- Edit

**And** upon saving, the amount will **BE SUBTRACTED** from the aggregates of the same entity for Schedule A transactions with the Aggregation Group "NATIONAL_PARTY_RECOUNT_ACCOUNT", including:
- Joint Fundraising Transfer - National Party Recount/Legal Proceedings Account
- PAC National Party Recount/Legal Proceedings Account JF Transfer Memo
- Party National Party Recount/Legal Proceedings Account
- PAC National Party Recount/Legal Proceedings Account
- Partnership National Party Recount/Legal Proceedings Account 
- Partnership National Party Recount/Legal Proceedings Account Memo
- Earmark Receipt for National Party Recount/Legal Proceedings Account (Contribution) & Memo

[...]

Issue #718, opened 2022-11-16
Allow user to enter Other Committee Refund - National Party Pres. Nominating Convention Account on Schedule B (Group E)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Other Committee Refund - National Party Pres. Nominating Convention Account”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click(open) on the "Other Expenditures" Accordion link  
**Then** I can enter a transaction with the type “Other Committee Refund - National Party Pres. Nominating Convention Account"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE**
-Committee
-	**COMMITTEE NAME** (Required)
-	**COMMITTEE ID** (Required)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional)
-	**AMOUNT** (Required, can be a negative value)

# Additional information 
#
-	**PURPOSE OF DISBURSEMENT** (System-generated)
-Text: 'Pres. Nominating Convention Account: Refund'
-	**NOTE OR MEMO TEXT** (Optional)
-	**CATEGORY CODE** (Optional)
---
**Then** I will be able to:
- Cancel
- Save
- Edit

**And** upon saving, the amount will **BE SUBTRACTED** from the aggregates of the same entity for Schedule A transactions with the Aggregation Group  "NATIONAL_PARTY_CONVENTION_ACCOUNT", including:
- Joint Fundraising Transfer - National Party Pres. Nominating Convention Account
- PAC National Party Pres. Nominating Convention Account JF Transfer Memo
- Party National Party Pres. Nominating Convention Account
- PAC National Party Pres. Nominating Convention Account
- Partnership National Party Pres. Nominating Convention Account
- Partnership National Party Pres. Nominating Convention Account Memo
- Earmark Receipt for National Party Pres. Nominating Convention Account (Contribution) & Memo

[...]

Issue #715, opened 2022-11-16
Allow user to enter Other Committee Refund - National Party Headquarters Buildings Account on Schedule B (Group E)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Other Committee Refund - National Party Headquarters Buildings Account”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a new transaction 
**Then** Click (open) on the "Other Expenditures" Accordion link  
**Then** I can enter a transaction with the type “Other Committee Refund - National Party Headquarters Buildings Account"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE** (required, includes LOOKUP)
-Committe
-	**COMMITTEE NAME** (Required)
-	**COMMITTEE ID** (Required)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional)
-	**AMOUNT** (Required, can be negative)

# Additional information 
#
-	**PURPOSE OF DISBURSEMENT** (System-generated)
-Text: 'Headquarters Buildings Account: Refund'
-	**NOTE OR MEMO TEXT** (Optional)
-	**CATEGORY CODE** (Optional)

---
**Then** I will be able to:
- Cancel
- Save
- Edit

**And** upon saving, the amount will **BE SUBTRACTED** from the aggregates of the same entity for Schedule A transactions with the Aggregation Group "NATIONAL_PARTY_HEADQUARTERS_ACCOUNT", including:
- Joint Fundraising Transfer - National Party Headquarters Buildings Account
- PAC National Party Headquarters Buildings Account JF Transfer Memo
- Party National Party Headquarters Buildings Account
- PAC National Party Headquarters Buildings Account
- Partnership National Party Headquarters Buildings Account
- Partnership National Party Headquarters Buildings Account Memo
- Earmark Receipt for National Party Headquarters Buildings Account (Contribution) & Memo

[...]

PR #993, opened 2023-04-20
Feature/715,718,721 Add national party other committee refunds schema

mjtravers:
#715 
#718 
#721


front-end/src/app/shared/models/schb-transaction.model.ts
```diff
@@ -52,7 +52,12 @@ export class SchBTransaction extends Transaction {
   override apiEndpoint = '/transactions/schedule-b';
 
   override getFieldsNotToValidate(): string[] {
-    return ['back_reference_tran_id_number', 'back_reference_sched_name', ...super.getFieldsNotToValidate()];
+    return [
+      'beneficiary_committee_name',
```
toddlees: Is this a required field that the user doesn't provide?

toddlees (changes requested):

front-end/src/app/shared/models/schb-transaction.model.ts
mjtravers: Yes, it is similar to the donor_committee_fec_id and donor_committee_name. The plan was to populate the name when the fec file is created.

Elaine-Krauss-TCG (approved):

dheitzer (approved):

toddlees (approved):

PR #1013, opened 2023-04-26
Feature/591 force itemize/unitemize transactions

dheitzer:
updated the "itemized" flag logic to use the new "force_itemized" flag so that we can avoid checking two flags everywhere (front-end, .fec construction, etc) and I thought it made more sense this way. Open to other options though.

### DIFF
--- a/front-end/src/app/transactions/transaction-list/transaction-list.component.html
+++ b/front-end/src/app/transactions/transaction-list/transaction-list.component.html
@@ -3,34 +3,15 @@
     <ng-template pTemplate="left">Transactions</ng-template>
 
     <ng-template pTemplate="right">
-      <p-overlayPanel #op [styleClass]="'action-options'">
-        <ng-template pTemplate>
-          <ul>
-            <li *ngFor="let action of tableActions">
-              <button
-                *ngIf="!action.isAvailable || action.isAvailable(report)"
-                [disabled]="!action.isEnabled(report)"
-                pButton
-                pRipple
-                type="button"
-                class="p-button-secondary p-button-text"
-                id="one"
-                (click)="onTableActionClick(action, report)"
-              >
-                {{ action.label }}
-              </button>
-            </li>
-          </ul>
-        </ng-template>
-      </p-overlayPanel>
-      <p-button
-        pRipple
-        icon="pi pi-plus"
-        label="Add new transaction"
-        styleClass="p-button-success mr-2"
-        (onClick)="op.toggle($event)"
-        ariaLabel="Add a Transaction"
-      ></p-button>
+      <app-table-actions-button
+        [tableActions]="tableActions"
+        [actionItem]="report"
+        [onTableActionClick]="onTableActionClick"
+        buttonIcon="pi pi-plus"
+        buttonLabel="Add new transaction"
+        buttonStyleClass="p-button-success mr-2"
+        buttonAriaLabel="Add a Transaction"
+      ></app-table-actions-button>
     </ng-template>
   </p-toolbar>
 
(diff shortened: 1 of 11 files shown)


Elaine-Krauss-TCG (approved):
Works great, looks great!

front-end/src/app/transactions/transaction-list/transaction-list.component.ts
```diff
@@ -89,6 +109,27 @@ export class TransactionListComponent extends TableListBaseComponent<Transaction
...
+    this.updateItem(transaction);
+  }
+
+  public forceUnitemize(transaction: Transaction): void {
+    transaction.force_itemized = false;
+    this.updateItem(transaction);
+  }
+
+  public updateItem(item: Transaction) {
+    this.itemService.update(item).subscribe(() => {
```
toddlees: probably want a takeUntil(this.destroy$)

front-end/src/app/transactions/transaction-list/transaction-list.component.html
```diff
@@ -111,7 +111,32 @@ <h5 class="m-0">Transactions during coverage dates</h5>
           {{ item.parent_transaction?.transaction_id }}
         </td>
         <td>
-          <p-button pRipple icon="pi pi-ellipsis-v" styleClass="p-button-text" ariaLabel="action"></p-button>
+          <p-overlayPanel #op [styleClass]="'action-options'">
```
toddlees: I wonder if there's a clean way to add this into the generic table component

toddlees (dismissed):

front-end/src/app/transactions/transaction-list/transaction-list.component.html
dheitzer: @toddlees since the generic table class (assuming you meant table-list-base) has no template, I'm trying to figure a way to implement any of this inside the .ts.  Do you have any suggestions for which bits? My initial thought is perhaps we might consider adding a template (we would need to generalize it for transaction-list, report-list, contact-list, user-list and deleted-contact-list).  I think it could be done, but it may be quite complex and could take a lot of time.  This actions kebab/dropdown could then be added to that template.

toddlees: Yeah i think that would get pretty big.  Maybe a compromise would be to turn the ellipsis and dropdown into a component.  The callbacks would still work because we bound them to the list component. No blocker though

mjtravers: Yes, let's pull the p-overlay with button HTML out into a component. At the very least it will reduce the HTML code duplication in the transaction-list component and I'm sure will be used elsewhere.

`<p-overlayPanel #op [styleClass]="'action-options'">
            <ng-template pTemplate>
              <ul>
                <li *ngFor="let action of rowActions">
                  <button
                    *ngIf="!action.isAvailable || action.isAvailable(item)"
                    pButton
                    pRipple
                    type="button"
                    class="p-button-secondary p-button-text"
                    id="one"
                    (click)="onRowActionClick(action, item)"
                  >
                    {{ action.label }}
                  </button>
                </li>
              </ul>
            </ng-template>
          </p-overlayPanel>
          <p-button 
            pRipple 
            icon="pi pi-ellipsis-v" 
            styleClass="p-button-text"
            (onClick)="op.toggle($event)"
            ariaLabel="action"
          ></p-button>
          
        </td>`

mjtravers (changes requested):
Look at making the component discussed in the comments. Also, the "Edit"  option in the transaction action drop down does not seem to work.

dheitzer:
Changes have been made to the action button/dropdown into a shared component and to fix the edit link on the transactions action menu.

front-end/src/app/transactions/transaction-list/transaction-list.component.ts
dheitzer: I opted for 

```
take(1),
takeUntil(this.destroy$)
```

Which will resolved either when the first is received or component is destroyed

mjtravers (approved):

mjtravers (approved):

PR #1021, opened 2023-05-01
Feature/977 cleanup unit test errors

Elaine-Krauss-TCG:
Cleaned up 147 errors and 3 warnings, comprising about 75% of the length of warnings/errors (see size of scrollbar) in the browser dev console when unit tests are run.

Of the remaining issues
 - 1 type of errors remain
    - 2 (NG04002) invalid route errors (these are likely the same error)
 - 1 type of warnings remain
    - 118 [disabled]="X" style warnings (the cure might be worse than the disease)

### DIFF
--- a/front-end/src/app/shared/resolvers/report.resolver.spec.ts
+++ b/front-end/src/app/shared/resolvers/report.resolver.spec.ts
@@ -42,7 +42,7 @@ describe('ReportResolver', () => {
     httpTestingController.verify();
   });
 
-  xit('should return undefined', () => {
+  it('should return undefined', () => {
     const route = {
       paramMap: convertToParamMap({ reportId: undefined }),
     };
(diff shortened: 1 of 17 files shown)


front-end/src/app/shared/resolvers/report.resolver.spec.ts
```diff
@@ -42,7 +42,7 @@ describe('ReportResolver', () => {
     httpTestingController.verify();
   });
 
-  it('should return undefined', () => {
+  xit('should return undefined', () => {
```
Elaine-Krauss-TCG, line 45: This unit test is broken.  The expect() call never registers with Karma, and attempts to make it asynchronous did not work

front-end/src/app/shared/resolvers/transaction.resolver.spec.ts
```diff
@@ -106,7 +106,7 @@ describe('TransactionResolver', () => {
     });
   });
 
-  it('should attach child for transaction with dependent child transaction type', () => {
+  xit('should attach child for transaction with dependent child transaction type', () => {
```
Elaine-Krauss-TCG, line 109: This unit test is broken.  The expect() call never registers with Karma, and attempts to make it asynchronous did not work

mjtravers (approved):

PR #1045, opened 2023-05-04
Feature/974 consolidate angular group templates

dheitzer:
consolidated templates into transaction-detail and double-transaction-detail components and added and replaced componentId with component group classes in models.


front-end/src/app/shared/models/transaction-groups/transaction-group-a.ts
```diff
@@ -0,0 +1,80 @@
...
+      templateMap.aggregate,
+      templateMap.purpose_description,
+      templateMap.employer,
+      templateMap.occupation,
+      templateMap.memo_code,
+      templateMap.memo_text_input,
+      'subTransaction',
+    ].filter(val => !!val);
+  }
+
```
mjtravers: The hope was that with the use of the templateMap we would not have to discern between the schedule A and schedule B when getting the form properties. For this case, schedule A, the only difference should be the category_code as the Group A has the employer and occupation by definition. Do we have Schedule B transaction types assigned to Group A that are individuals that do not collect the employer and occupation data?

front-end/src/app/shared/models/transaction-groups/transaction-group-a.ts
```diff
@@ -0,0 +1,80 @@
...
+    return false;
+  }
+
+  hasElectionInformationInput(): boolean {
+    return false;
+  }
+
+  getAmountInputTitle(): string {
+    return 'Receipt Information';
+  }
```
mjtravers: Is amountInputTitle something standard across schedule types and can be put in the templateMap?

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.spec.ts
```diff
@@ -304,7 +304,20 @@ describe('TransactionTypeBaseComponent', () => {
       life: 3000,
     };
     const messageServiceAddSpy = spyOn(testMessageService, 'add');
-    component.navigateTo(new NavigationEvent(NavigationAction.SAVE, NavigationDestination.ANOTHER));
+    spyOn(testRouter, 'navigateByUrl').and.callFake(
+      () =>
+        new Promise((res) => {
+          res(true);
+        })
+    );
```
mjtravers: This promise is being reported as a code smell by SonarCloud. Let's look at SonarCloud's recommendation for changing the code.

front-end/src/app/shared/models/transaction-groups/transaction-group-a.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      memo_code: '',
+      amount: '',
+      aggregate: '',
+      purpose_description: '',
+      purposeDescripLabel: '',
+      memo_text_input: '',
+      category_code: testCategoryCode,
+      election_code: '',
+      election_other_description: ''
+    }
```
mjtravers: There is a lot of code duplication defining these testTemplateMaps in each spec file and across all the transaction-group spec files. Can we migrate all of them into 1 or 2 definitions in the unit-test.utils.ts library?

front-end/src/app/shared/models/transaction-groups/transaction-group-a.ts
dheitzer: After discussing with you and Gregg yesterday, I went ahead and made this change to single templateMap.

front-end/src/app/shared/models/transaction-groups/transaction-group-a.ts
dheitzer: nice catch.  Went ahead and added this to templateMap.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.spec.ts
dheitzer: pulled in develop changes to fix this.

front-end/src/app/shared/models/transaction-groups/transaction-group-a.spec.ts
dheitzer: used unit-test.utils as suggested and removed a lot of template def code.

mjtravers (approved):

Issue #469, opened 2022-08-16
Allow user to enter Partnership Individual Recount/Legal Proceedings Account JF Transfer Memo on Schedule A (Group A)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Partnership Individual Recount/Legal Proceedings Account JF Transfer Memo”** sub-transaction under a **”Partnership Receipt Recount/Legal Proceedings Account JF Transfer Memo”**  on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

**NOTE:** This is a Tier 3 transaction that will be associated with a Tier 2 AND Tier 1 transaction.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**And** I am editing a Partnership Receipt Recount/Legal Proceedings Account JF Transfer Memo on an open report  
**Then** I will have the option to enter the following sub-transaction: 

- Partnership Individual Recount/Legal Proceedings Account JF Transfer Memo

**And** I will be required to enter the following information (unless otherwise noted):

---

# Contact 
#
-	**CONTACT TYPE** (Read-only, includes LOOKUP)
-- Individual 
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)
-	**EMPLOYER** (Required if Aggregate > $200)
-	**OCCUPATION** (Required if Aggregate > $200)

# Receipt Information 
#
-	**DATE RECEIVED** (Required)
-	**MEMO ITEM** (Required, system checked and read-only)
-	**AMOUNT** (Required, can be negative)
-	**AGGREGATE** (Required) 
-- Amount Displayed will be the total of all contributions that are:
--- From the person (ENTITY) and within "National Party Recount Account" _(Dev/QA Note: Aggregation Group is "NATIONAL_PARTY_RECOUNT_ACCOUNT")_
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# Additional information 
#
[...]

Issue #468, opened 2022-08-16
Allow user to enter Partnership Receipt Recount/Legal Proceedings Account JF Transfer Memo on Schedule A (Group D)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Partnership Receipt Recount/Legal Proceedings Account JF Transfer Memo”** sub-transaction under a **”Joint Fundraising Transfer - National Party Recount/Legal Proceedings Account”**  on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

**NOTE:** This is a Tier 2 transaction that will be associated with a Tier 1 transaction.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**And** I am editing a Joint Fundraising Transfer - National Party Recount/Legal Proceedings Account on an open report  
**Then** I will have the option to enter the following sub-transaction: 
- Partnership Receipt Recount/Legal Proceedings Account JF Transfer Memo

**And** I will be required to enter the following information (unless otherwise noted):

---

# Contact 
#
-	**CONTACT TYPE** (Read-only, includes LOOKUP)
-- Organization 
-	**ORGANIZATION NAME** (Required)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Receipt Information 
#
-	**DATE RECEIVED** (Required)
-	**MEMO ITEM** (Required, system checked and read-only)
-	**AMOUNT** (Required, can be negative)
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of all contributions that are:
--- From the person (ENTITY) and within "National Party Recount Account" _(Dev/QA Note: Aggregation Group is "NATIONAL_PARTY_RECOUNT_ACCOUNT")_
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# Additional information 
#
-	**PURPOSE OF RECEIPT** (Read-only, system generated)
-- If transaction has no associated Partnership memos, reads "Recount/Legal Proceedings Account JF Memo: XX (Partnership attributions do not require itemization)". 
[...]

PR #1046, opened 2023-05-04
Feature/468,469 partnership receipt memos

mjtravers:
#468 
#469


front-end/src/app/shared/models/transaction-types/PARTNERSHIP_NATIONAL_PARTY_RECOUNT_JF_TRANSFER_MEMO.model.ts
```diff
@@ -0,0 +1,50 @@
...
+  override generatePurposeDescription(transaction: SchATransaction): string {
+    let committeeClause = `Recount/Legal Proceedings Account JF Memo: ${
+      (transaction.parent_transaction as SchATransaction).contributor_organization_name
+    }`;
+    const hasChildren = transaction.children && transaction.children.length > 0;
+    const parenthetical = hasChildren
+      ? ' (See Partnership Attribution(s) below)'
+      : ' (Partnership attributions do not require itemization)';
+    if ((committeeClause + parenthetical).length > 100) {
+      committeeClause = committeeClause.slice(0, 97 - parenthetical.length) + '...';
```
dheitzer, line 44: Looks like you'll run into the same issue I ran into on 504 here where the CPD won't validate:  

Recount/Legal Proceedings Account JF Memo:  (Partnership attributions do not require itemization)

is already 97 chars so adding ... will cause this slice to remove the ':' thus invalidating the pattern.  

The solution I recommend (that I made on 504) was to add this special case:

```
const base = 'Recount/Legal Proceedings Account JF Memo: ';
if (hasChildren) {
  return base + ' (Partnership attributions do not require itemization)'
}
```

@AureliaKhorsand said this is acceptable until we increase the CPD field length.

dheitzer (changes requested):
just one find

front-end/src/app/shared/models/transaction-types/PARTNERSHIP_NATIONAL_PARTY_RECOUNT_JF_TRANSFER_MEMO.model.ts
mjtravers: @dheitzer I am not running into this problem and am able to save transactions without triggering the validation error. Looks like that description plus the "..." is coming in at exactly 100 characters. Are you hitting this validation issue in your testing of the ticket?

dheitzer (approved):

Issue #930, opened 2023-03-20
UI: Update Account Info Page Look and feel


MitchellTCG:
***Task:*** Update "Account Info" page to look like the rest of the application.

Acceptance Criteria:

1\. The style of Account Info page should match the style of the Transaction entry pages.

  2\. The fields on the Account Info page area read-only.

  2\. There is a message indicating to the user "You can only view your committee's information on this page". 

  4\. Text in read only boxes should be Karla Bold at 14px

QA notes

*   No functionality of the page is changing, this is just the layout/visual of how the screen should appear.

Here is an updated wireframe for that page:

![60 - Updated Accouint Info.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/e2a7cdea-40d4-401e-a2d4-1c72dd82c200](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/e2a7cdea-40d4-401e-a2d4-1c72dd82c200))

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-842](https://fecgov.atlassian.net/browse/FECFILE-842)

MitchellTCG:
@JonellaCulmer @mary-tcg @AureliaKhorsand 

Please review

MitchellTCG:
@AureliaKhorsand 
Please review and place into ADTR if you feel it is ready for pointing.

Elaine-Krauss-TCG:
Everything has been switched over to field, and the fields auto-populate as expected.  The fields need to be disabled and the details of styling need an overhaul, but the lion's share of the ticket is done.

![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/02799547-e528-47f8-802f-6df47a64fa10)

mjtravers:
In CR, the TELEPHONE field is editable. Sending back to In Progress for a fix.

Also, the AccountInfoComponent has been extended with the DestroyerComponent so a pull will be needed to get the updates before working on the readonly telephone form input.

mjtravers:
There was a mix up in the branch I CR'ed. No code changes were needed when sent back to In Progress. Passes CR. Sending to QA.

WiseQA:
QA review Account information is title case needs changed to sentence case match the style of the Transaction entry pages.

Displaying as:

[...]

PR #1081, opened 2023-05-16
Feature/830 Transaction read-only screens

toddlees:
#930


mjtravers (approved):
Nice, you removed the "edit" text from the URLs, too. Passes CR.

PR #1097, opened 2023-05-17
Feature/1061 Memo Text Validation Fix

### LINKED ISSUE #1061: Note or Memo Text (Optional) field not recognizing characters not allowed
During testing when entering a non allowed character in NOTE OR MEMO TEXT (OPTIONAL) field for both Receipts and Disbursement the field does not display the error message "This field contains characters that are not allowed." Upon saving populates error message (tested on both STAGE and DEV environments).

Receipt:

)

Disbursement:

)

Error message when selecting "Save"

)

Per review with @MitchellTCG @mjtravers fix characters not allowed in NOTE OR MEMO TEXT (OPTIONAL) field for both receipts and disbursements. 2 point bug fix.

***DEV NOTES***

  This bug is related to the mapping of the NOTE OR MEMO TEXT field to the MemoText object which has a separate validation schema from the transaction type schemas used to validate all the other form fields on the transaction input page.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-778](https://fecgov.atlassian.net/browse/FECFILE-778)

Elaine-Krauss-TCG:
#1061 

`memo_text_input` fields now validate against the Text.json schema file, and this required changing the form control name to `text4000` to avoid some really awkward workarounds.


toddlees (approved):

front-end/src/app/shared/components/inputs/additional-info-input/additional-info-input.component.ts
```diff
@@ -19,6 +21,8 @@ export class AdditionalInfoInputComponent extends BaseInputComponent implements
     //Add 'implements OnInit' to the class.
 
     this.patchPurposeDescriptionPrefix();
+
+    ValidateUtils.addJsonSchemaValidators(this.form, memoTextSchema, false);
```
toddlees: I noticed that the memo text spec has 
`
            "type": [
                "string",
                "null"
            ],`
  for all of its fields, meaning though they're in the required fields list, we allow null (meaning, it's effectively not required).  If we fix that in the schema, this might stop working because the form will be missing the other fields from the schema.  Not an immediate issue, but we'll want to keep an eye on it

toddlees: nm we are good to go here.  first off, other fields in the text spec are limited to 'string' already.  second it's already been set up on the front end to ignore errors from fields that aren't on the form.  So a field like `rec_type` in the text spec would never impact the transaction form because it's not part of the form.

toddlees (approved):

PR #1113, opened 2023-05-22
feature/1106 just checking box showing when typing date

dheitzer:
date is null until full string is types (ie '05/22/2023') but still is sent to the Observable.  Need to add null check on date to updateMemoItemWithDate so that this method doesn't trigger the dialog before they have finished typing.

### DIFF
--- a/front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
+++ b/front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
@@ -74,8 +74,8 @@ export class AmountInputComponent extends BaseInputComponent implements OnInit, 
   }
 
   updateMemoItemWithDate(date: Date) {
-    if (date && this.report?.coverage_from_date && this.report?.coverage_through_date) {
-      if (date < this.report.coverage_from_date || date > this.report.coverage_through_date) {
+    if (this.report?.coverage_from_date && this.report?.coverage_through_date) {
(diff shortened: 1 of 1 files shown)


Elaine-Krauss-TCG (approved):

toddlees (approved):

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
```diff
@@ -74,7 +74,7 @@ export class AmountInputComponent extends BaseInputComponent implements OnInit,
   }
 
   updateMemoItemWithDate(date: Date) {
-    if (this.report?.coverage_from_date && this.report?.coverage_through_date) {
+    if (date && this.report?.coverage_from_date && this.report?.coverage_through_date) {
```
toddlees: By adding this, the else below  doesn't get hit .  If we set a date outside and then erase it, this.dateIsOutsideReport stays true. So clicking on the memo box throws the alert

image

toddlees (changes requested):

toddlees (approved):

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
dheitzer: Added suggested patch by @toddlees

PR #1116, opened 2023-05-22
Feature/937

toddlees:


front-end/src/app/reports/styles.scss
```diff
@@ -54,6 +54,11 @@ td.justify-left {
   text-align: left;
 }
 
+.cell-with-overlay {
+  display: flex;
+  justify-content: space-between;
```
dheitzer, line 59: Cool - I've not seen this usage of justify-content before

dheitzer (changes requested):
Looks nice.  Only concern - the tooltip overlays don't seen to be accessible via the keyboard.  I'm wondering if these need to be? Maybe a requirements question.

dheitzer (approved):

Issue #739, opened 2022-11-16
Allow system to create In-kind Out on Schedule B (Group A)


GreggMoreland:
Business Reason

As a Committee Administrator, I will be able to view an ***In-kind Out*** transaction on Schedule B of a Form 3x report when an ***In-kind Receipt*** Transaction on Schedule A of a Form 3x report is saved, so that I can accurately report on my committee's transactions.

Acceptance Criteria

***Given*** I am a Committee Administrator editing an open Form 3X report

 ***If*** an ***"In-kind Receipt"*** (Schedule A) transaction is saved

 ***Then*** an ***In-kind Out*** (Schedule B) transaction will be created with the following information as read-only unless otherwise indicated (users can edit through the transaction on Schedule A):

—

Contact

 #

*   ***CONTACT TYPE***

      -Individual Only

*   ***LAST NAME*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***FIRST NAME*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***MIDDLE NAME*** (Optional, auto-generated from associated IN_KIND_RECEIPT)
*   ***PREFIX*** (Optional, auto-generated from associated IN_KIND_RECEIPT)
*   ***SUFFIX*** (Optional, auto-generated from associated IN_KIND_RECEIPT)

1.  Address

      #

*   ***STREET ADDRESS*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***APARTMENT, SUITE, ETC.*** (Optional, auto-generated from associated IN_KIND_RECEIPT)
*   ***CITY*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***STATE/PROVINCE*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***ZIP/POSTAL CODE*** (Required, auto-generated from associated IN_KIND_RECEIPT)

1.  Expenditure Information

      #

*   ***DATE*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   ***MEMO ITEM*** (auto-generated from associated IN_KIND_RECEIPT)
*   ***AMOUNT*** (Required, auto-generated from associated IN_KIND_RECEIPT)
*   AGGREGATE (Required)
      *   Amount Displayed will be the total of contributions that are:
            *   From the person (ENTITY)
    *   within Aggregation Group of "GENERAL_DISBURSEMENT"
[...]

Issue #642, opened 2022-10-25
Allow user to enter In-kind Receipt on Schedule A (Group A)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”In-kind Receipt”** transaction on Schedule A of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 

**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a receipt
**Then** Click(open) on the "Contributions from Individuals/Persons" Accordion link  
**Then** I can enter a transaction with the type "In-kind Receipt"
**Then** I will be required to enter the following information (unless otherwise noted):

---

# Contact 
#
-	CONTACT TYPE (required, read-only)
-- Individual only
-	LAST NAME (Required)
-	FIRST NAME (Required)
-	MIDDLE NAME (Optional)
-	PREFIX (Optional)
-	SUFFIX (Optional)

# Address 
#
-	STREET ADDRESS (Required)
-	APARTMENT, SUITE, ETC. (Optional)
-	CITY (Required)
-	STATE/PROVINCE (Required)
-	ZIP/POSTAL CODE (Required)
-	EMPLOYER (Required if Aggregate > $200)
-	OCCUPATION (Required if Aggregate > $200)

# Receipt Information 
#
-	DATE RECEIVED (Required)
-	MEMO ITEM (Optional, will be required if CONTRIBUTION DATE falls outside of report dates)
-	AMOUNT (Required, can be negative)
-	AGGREGATE (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# Additional information 
#
-	PURPOSE OF RECEIPT (Required, system generated and user updated)
-- Text should read: In Kind: XX
-- And 'XX' is user entered and concatenated with 'In Kind:' within the CPD
**NOTE: System will automatically prefix the entered text with 'In Kind: '**
-	NOTE OR MEMO TEXT (A/N-100, Optional)

---
**Then** I will be able to:

- Cancel
- Edit
- Save both transactions (When this is Saved, an 'In-kind Out' transaction is created on Schedule B as defined in [#739])

**And if** the Aggregate amount is >= $200.01, 
[...]

PR #1127, opened 2023-06-05
Feature/642,739 Create in-kind receipt and In-kind out transactions

mjtravers:
#642 
#739


front-end/src/app/shared/models/transaction-groups/transaction-group-aa.model.ts
```diff
@@ -103,7 +103,7 @@ export class TransactionGroupAA extends DoubleTransactionGroup {
 
   getAutoGeneratedChildFields(childTemplateMap: TransactionTemplateMapType): string[] {
     return this.getChildFormProperties(childTemplateMap).filter(
-      (field) => ![childTemplateMap.text4000].includes(field)
+      (field) => ![childTemplateMap.text4000, childTemplateMap.category_code].includes(field)
```
toddlees, line 112: Did you mean to add memo_code rather than category_code?

toddlees (changes requested):

front-end/src/app/shared/models/transaction-groups/transaction-group-aa.model.ts
mjtravers: I added category code here so that it was not set read-only in the form. This field is not pulling values from the parent schedule A but should allow the user to select a category code for the schedule B transaction.

toddlees: image

~~I think the user is allowed to edit that part.  The schema shows it as variable, like the memo~~

toddlees: ~~If it's not editable, should it be in the interface/schema at all?  it will never be populated~~

toddlees: I just figured out what i'm saying and what you did after testing it locally 🤦

toddlees (approved):

Elaine-Krauss-TCG (approved):

Issue #1111, opened 2023-05-19
Create Angular model for Schedule C2


mjtravers:
Following the same pattern as was created for the Schedule A and Schedule B, create a Schedule C2 class to manage the data for a Schedule C2 record.

DEV Tasks:

*   [x] Create a SchC2Transaction class that extends the Transaction class
*   [x] Create a ScheduleC2TransactionGroups enum containing groupings for the "Add transacation" expand/collapse page
*   [x] Create a ScheduleC2TransactionsGroupsType type
*   [x] Create a ScheduleC2TransactionTypes enum containing the transaction unique identifiers
*   [x] Create a ScheduleC2TransactionTypeLabels LabelList containing the UI title for each transaction type
*   [x] Update the ScheduleTransaction, TransactionTypes, and TransactionGroupTypes type definitions in the transactions.model.ts file

### QA Notes ###

There will be a unit test that can be reviewed verifying the correctness of the Schedule C2 model code. A screenshot can be provided.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-756](https://fecgov.atlassian.net/browse/FECFILE-756)

mjtravers:
Passes CR. Sending to QA.

mjtravers:
Unit test results for QA: https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/2994/workflows/76523d30-baf1-4b44-a13e-0642521f4339/jobs/11170

WiseQA:
QA review consisted of visual inspection of code passing as denoted in app.circleci per DEV.

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/92c764e2-83b9-44b0-b5da-f9b3597daa0b)

QA Review Completed. Moved to Stage Ready,

Issue #1110, opened 2023-05-19
Create Angular model for Schedule C1


mjtravers:
Following the same pattern as was created for the Schedule A and Schedule B, create a Schedule C1 class to manage the data for a Schedule C1 record.

DEV Tasks:

*   [x] Create a SchC1Transaction class that extends the Transaction class
*   [x] Create a ScheduleC1TransactionGroups enum containing groupings for the "Add transacation" expand/collapse page
*   [x] Create a ScheduleC1TransactionsGroupsType type
*   [x] Create a ScheduleC1TransactionTypes enum containing the transaction unique identifiers
*   [x] Create a ScheduleC1TransactionTypeLabels LabelList containing the UI title for each transaction type
*   [x] Update the ScheduleTransaction, TransactionTypes, and TransactionGroupTypes type definitions in the transactions.model.ts file

### QA Notes ###

There will be a unit test that can be reviewed verifying the correctness of the Schedule C1 model code. A screenshot can be provided.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-757](https://fecgov.atlassian.net/browse/FECFILE-757)

mjtravers:
Passes CR. Sending to QA.

mjtravers:
Unit test results for QA: https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/2994/workflows/76523d30-baf1-4b44-a13e-0642521f4339/jobs/11170

WiseQA:
QA review consisted of visual inspection of code passing as denoted in app.circleci per DEV.

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/83a2757e-e26c-4d29-ac88-daa6d62a4983)

QA Review Completed. Moved to Stage Ready,

PR #1130, opened 2023-06-06
Feature/1110-1111 Schedules C1 and C2

Elaine-Krauss-TCG:
#1110
#1111


front-end/src/app/shared/models/schc1-transaction-type.model.ts
```diff
@@ -0,0 +1,35 @@
...
+    memo_code: 'memo_code',
+    amount: 'loan_amount',
+    amountInputHeader: 'Loan Information',
+    aggregate: '',
+    purpose_description: '',
+    purposeDescripLabel: '',
+    text4000: 'text4000',
+    category_code: '',
+    election_code: 'election_code',
+    election_other_description: 'election_other_description',
```
mjtravers: I think election_code and election_other_description can be empty ''. I don't see those fields in the schedule C1 spec

maybe the same for memo_code and text400... we can add them back if they turn up on the input form. I'd rather remove them and have to add them later as if we leave them we might forget

front-end/src/app/shared/models/schc1-transaction.model.ts
```diff
@@ -0,0 +1,94 @@
...
+  authorized_title: string | undefined;
+  @Transform(BaseModel.dateTransform) authorized_date_signed: Date | undefined;
+
+  entity_type: string | undefined;
+  receipt_line_number: string | undefined;
+  election_code: string | undefined;
+  election_other_description: string | undefined;
+  memo_code: boolean | undefined;
+  memo_text_description: string | undefined;
+  aggregation_group: AggregationGroups | undefined;
```
mjtravers, line 54: Where are the above fields (entity_type - aggregation_group) coming from?

front-end/src/app/shared/models/schc2-transaction.model.ts
```diff
@@ -0,0 +1,62 @@
...
+  guarantor_occupation: string | undefined;
+  guaranteed_amount: number | undefined;
+
+  entity_type: string | undefined;
+  receipt_line_number: string | undefined;
+  election_code: string | undefined;
+  election_other_description: string | undefined;
+  memo_code: boolean | undefined;
+  memo_text_description: string | undefined;
+  aggregation_group: AggregationGroups | undefined;
```
mjtravers, line 22: Same question as C1 above about adding these fields.

front-end/src/app/shared/models/schc2-transaction-type.model.ts
```diff
@@ -0,0 +1,35 @@
...
+    memo_code: 'memo_code',
+    amount: 'guaranteed_amount',
+    amountInputHeader: 'Loan Information',
+    aggregate: '',
+    purpose_description: '',
+    purposeDescripLabel: '',
+    text4000: 'text4000',
+    category_code: '',
+    election_code: 'election_code',
+    election_other_description: 'election_other_description',
```
mjtravers: Same comment as C1 above about defining values for fields not in the spec

mjtravers (commented):
@Elaine-Krauss-TCG Have some questions about fields not in the schedule c1 or c2 spec but added to the models. Will check in with you to discuss

front-end/src/app/shared/models/schc1-transaction-type.model.ts
Elaine-Krauss-TCG: I election_code, election_other_description, and memo_code to empty.

I kept text4000, as I *expect* that that's necessary.

front-end/src/app/shared/models/schc2-transaction-type.model.ts
Elaine-Krauss-TCG: Same response as C1 type model.

mjtravers (approved):

PR #1139, opened 2023-06-09
Patch for out of stale contact data being submitted

mjtravers:


front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
```diff
@@ -145,4 +145,38 @@ export class TransactionContactUtils {
       }
     }
   }
+
+  static updateContactFromForm(form: FormGroup, transaction: Transaction) {
```
toddlees: it seems like this is fulfilling the same purpose as setTransactionContactFormChanges().  Are we doubling up our code?

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
mjtravers: yes, the old function did the trick. am removing the new one

toddlees (approved):

Issue #1129, opened 2023-06-06
UI: System to show status for Committee in site header


AureliaKhorsand:
Task: Include Committee status in the header

Under the Committee name and to the left of the committee type please the following:
- [x] Display a 16px diameter circle either Grey (if terminated) or Green (if active)
- [x] these are the three possible statuses
- Active
- Terminated
- Administratively Terminated
- [x] If active please display the reporting cycle to the right (Choices are: Quarterly and Monthly)
- [x] Lastly make sure this data is wired correctly to the database

# Examples 
#
![The Dashboard – 14.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/cabd2faf-58af-4ec2-bcd2-4779e8fbfe4f)

![The Dashboard – 15.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/ea6f890b-8429-4865-884c-1b9dd45fd356)

![The Dashboard – 16.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/d0b2ec78-2e25-490d-9347-0337445a6954)

![The Dashboard – 17.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/77af34e6-fcd2-4fb5-a4ea-ad5bc1e7c576)

![The Dashboard – 18.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/8a7cf378-042e-4cba-99ce-e3c68735ccee)

### QA Notes ### 
 null
### DEV Notes ### 
 null
### Design ### 
 null

MitchellTCG:
@mjtravers 
Please review ACs. let me know if you need more.
I know this is already built by this ticket will track it

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified status is displayed under the committee name for the following status:  Active. 
Terminated, and  Administratively Terminated

QA Review Completed. Move to Stage Ready after Sprint 26 review 22 June 2023.

PR #1159, opened 2023-06-15
Feature/1129 Update committee header styling

mjtravers:
#1129

### DIFF
--- a/front-end/src/app/layout/committee-banner/committee-banner.component.ts
+++ b/front-end/src/app/layout/committee-banner/committee-banner.component.ts
@@ -4,6 +4,36 @@ import { takeUntil } from 'rxjs';
 import { selectCommitteeAccount } from 'app/store/committee-account.selectors';
 import { DestroyerComponent } from 'app/shared/components/app-destroyer.component';
 
+const committeeStatusCodes: { [key: string]: string } = {
+  T: 'Terminated',
+  A: 'Administratively terminated',
+  D: 'Active - Debt',
+  W: 'Active - Waived',
+  M: 'Active - Monthly',
+  Q: 'Active - Quarterly',
+};
+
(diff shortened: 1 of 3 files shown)


front-end/src/app/layout/committee-banner/committee-banner.component.ts
```diff
@@ -46,28 +48,26 @@ export class CommitteeBannerComponent implements OnInit, OnDestroy {
     Z: 'national party non-federal account',
   };
 
-  public active_committees = ['M', 'Q', 'W', 'D'];
+  public activeCommittees = ['M', 'Q', 'W', 'D'];
```
toddlees: Maybe since you're already renaming it, we could call it `activeStatuses` since it's a list of status codes rather than committees

front-end/src/app/layout/committee-banner/committee-banner.component.ts
```diff
@@ -46,28 +48,26 @@ export class CommitteeBannerComponent implements OnInit, OnDestroy {
...
 
-      const committee_type_code = committeeAccount.committee_type;
-      if (committee_type_code) {
-        this.committeeType = this.committee_types[committee_type_code] || '';
-      }
-    });
-  }
+        const frequencyCode = committeeAccount.filing_frequency;
+        if (frequencyCode) {
+          this.committeeFrequency = this.committee_statuses[frequencyCode] ?? '';
```
toddlees: cool! i've never used `??` I like that it's more specific to undefined|null

toddlees (approved):

front-end/src/app/layout/committee-banner/committee-banner.component.ts
mjtravers: This variable name was updated to activeStatusCodes

Issue #990, opened 2023-04-19
System to prevent user from saving new contact with existing Committee or Candidate ID 


AureliaKhorsand:
Task: Prevent users from saving new contacts that share the same Committee or Candidate ID as an existing contact in Committee Account.

Acceptance Criteria

***Given*** that I am creating or editing a new contact in Contact Management or within a transaction,

  ***If*** the new committee or candidate contact I am trying to save matches the Committee ID or Candidate ID (respectively) of an existing contact in my Committee Account,

  ***Then*** I will be unable to save that new contact, and the system will indicate the reason.

Dev Notes

  Check on front _and_ back-end to guard against edge cases.

Wireframe

1.  FEC ID's must be unique.

       ![990 - Unique IDs.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/116ce995-0eda-4edc-b3bf-71fc91cf8527](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/116ce995-0eda-4edc-b3bf-71fc91cf8527))

### QA Notes ###
Note: This will be when creating and editing a contact in a transaction **and** in Contact Management.

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-819](https://fecgov.atlassian.net/browse/FECFILE-819)

AureliaKhorsand:
@MitchellTCG @mary-tcg can we get a warning message for the user for this ticket? Simple is fine - we just want to functionally prevent this from happening.

MitchellTCG:
@AureliaKhorsand 
I added wireframes to this ticket that correspond to Todd's suggested solution.

mjtravers:
There are a couple of small requested changes from CR. Sending back to In Progress.

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified the Committee ID field displays the "FEC IDs must be unique" error when entering an existing Committee ID.

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/f25cb4d6-feff-4ebd-b476-92a4ae7b0aee)

QA review verified the Candidate ID field displays the "FEC IDs must be unique" error when entering an existing Candidate ID.

![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/0f919bc9-f717-4290-a413-9028ad35e227)

QA Review Completed.   Moved to Stage Ready.

PR #1163, opened 2023-06-15
invalidate contact form if fec id in use

toddlees:
#990

### DIFF
--- a/front-end/src/app/shared/components/contact-form/contact-form.component.ts
+++ b/front-end/src/app/shared/components/contact-form/contact-form.component.ts
@@ -1,5 +1,5 @@
 import { Component, Input, OnInit } from '@angular/core';
-import { FormBuilder, FormGroup } from '@angular/forms';
+import { AbstractControl, AsyncValidatorFn, FormBuilder, FormGroup, ValidationErrors } from '@angular/forms';
 import { ContactService } from 'app/shared/services/contact.service';
 import { FecApiService } from 'app/shared/services/fec-api.service';
 import { CountryCodeLabels, LabelUtils, PrimeOptions, StatesCodeLabels } from 'app/shared/utils/label.utils';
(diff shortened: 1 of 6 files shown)


front-end/src/app/shared/components/contact-form/contact-form.component.ts
```diff
@@ -107,6 +107,8 @@ export class ContactFormComponent extends DestroyerComponent implements OnInit {
           this.candidateDistrictOptions = [];
         }
       });
+    this.form?.get('candidate_id')?.addAsyncValidators(this.contactService.fecIdValidator);
```
dheitzer, line 110: I wonder if we should add debounce to these as you mentioned so we're not hitting the api too much:

https://stackoverflow.com/questions/49090659/angular-reactive-forms-debounce-only-some-specific-form-control#answer-49093827

Alternatively, you could add it to the validator fn below like here https://stackoverflow.com/questions/36919011/how-to-add-debounce-time-to-an-async-validator-in-angular-2#answer-62662296

dheitzer (dismissed):

front-end/src/app/shared/components/error-messages/error-messages.component.html
```diff
@@ -1,5 +1,6 @@
 <ng-container *ngIf="formSubmitted || (control?.invalid && (control?.dirty || control?.touched))">
   <small class="p-error" *ngIf="control?.hasError('required')">{{ requiredErrorMessage }}</small>
+  <small class="p-error" *ngIf="control?.hasError('fec_id_unique')">{{ uniqueFecIdMessage }}</small>
```
mjtravers: small nit pick: can we camelCase fec_id_unique as fecIdUnique?

front-end/src/app/shared/components/contact-form/contact-form.component.ts
mjtravers: While testing, it looks like the synchronous pattern validation rule is throwing an error while the id string is less than 9 characters long and blocking the async validation rule from firing. I am only seeing network activity to check the uniqueness of the id when the string reaches 9 characters long.

image

So at the moment it doesn't look like debouncing is needed but I think it would be nice to put in anyway because at some point we may be asked to clean up that pattern error message and that may remove the sync validation block.

mjtravers (changes requested):
@dheitzer had a good suggestion about debouncing. It looks like a small addition to the observable pipe()

front-end/src/app/shared/components/contact-form/contact-form.component.ts
toddlees: I was in fact relying on the pattern validator to limit the usage of the async validator. the uniqueness validator is only called when the id is a complete and valid id, so it's infrequent.  But I went ahead and added the debounce to Matt's point that the pattern validator could change.

front-end/src/app/shared/components/error-messages/error-messages.component.html
toddlees: addressed

mjtravers (approved):

Issue #1120, opened 2023-05-24
Update transactions listed in Receipt/Disbursement accordion list


GreggMoreland:
***Task:*** Remove duplicate transaction and transactions which had names changed and are no longer applicable.

Acceptance Criteria

1\. ***Reattribution*** should be removed from "Contributions from Individuals/Persons" in Receipts

  1\. ***Void of Operating Expenditure*** should be removed from 'Contributions/Expenditures to Regular Filers' in Disbursements - it should still be visible and accessible under 'Operating Expenditures'

  2\. The following should all be removed under "Other Expenditures" in Disbursements

*   Headquarters Account - Individual Refund
*   Headquarters Account - Regular Refund
*   Convention Account - Individual Refund
*   Convention Account - Regular Refund
*   National Party Recount Account - Individual Refund
*   National Party Recount Account - Regular Filer Refund

Dev Notes

*   These are all dead links at the moment.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-751](https://fecgov.atlassian.net/browse/FECFILE-751)

mjtravers:
@toddlees See PR comments with a question about removing the references to the transaction types in the schedule A/B transaction models as well.

mjtravers:
There are a couple of small requested changes from CR. Sending back to In Progress.

WiseQA:
QA review verified the transaction names have been updated for both the Receipt and Disbursement accordion lists.

1.  Reattribution has been removed from "Contributions from Individuals/Persons" in Receipts
2.  Void of Operating Expenditure has been removed from 'Contributions/Expenditures to Regular Filers' in Disbursements.  
3.  Void of Operating Expenditure is listed under 'Operating Expenditures"

4.  The following have been removed under "Other Expenditures" in Disbursements
Headquarters Account - Individual Refund
Headquarters Account - Regular Refund
Convention Account - Individual Refund
Convention Account - Regular Refund
National Party Recount Account - Individual Refund
National Party Recount Account - Regular Filer Refund

QA Review Completed. Move to Stage Ready after Sprint 26 review 22 June 2023.

PR #1166, opened 2023-06-16
Feature/1120

toddlees:
#1120


mjtravers (commented):
Except for OPERATING_EXPENDITURE_VOID, are any of the others going to be transaction types in the future? If not, we can also remove the references to them in the scha-transaction.model.ts and schb-transaction.model.ts files.

mjtravers (approved):

PR #1180, opened 2023-06-22
feature/731+738 - SchB GroupO 100% Federal Election Activity Payment/Void

dheitzer:


front-end/src/app/shared/models/transaction-groups/transaction-group-a.model.spec.ts
```diff
@@ -56,4 +56,9 @@ describe('TransactionGroupA', () => {
     const retval = component.hasCandidateInformationInput();
     expect(retval).toBeFalse();
   });
+
+  it('#hasCandidateCommitteeInput happy path', () => {
+    const retval = component.hasCandidateCommitteeInput();
```
toddlees: "retval" isn't returned meaning it's not a "return value".  Any reason this can't be ```expect(component.hasCandidateCommitteeInput()).toBeFalse();```?

front-end/src/app/shared/models/transaction-types/FEDERAL_ELECTION_ACTIVITY_VOID.model.ts
```diff
@@ -0,0 +1,26 @@
...
+import { AggregationGroups } from '../transaction.model';
+
+export class FEDERAL_ELECTION_ACTIVITY_VOID extends SchBTransactionType {
+  transactionGroup = new TransactionGroupO();
+  title = LabelUtils.get(ScheduleBTransactionTypeLabels, ScheduleBTransactionTypes.FEDERAL_ELECTION_ACTIVITY_VOID);
+  schema = schema;
+  override negativeAmountValueOnly = true;
+  override showAggregate = false;
+  override purposeDescriptionLabelSuffix = PurposeDescriptionLabelSuffix.REQUIRED;
+  override navigationControls: TransactionNavigationControls = STANDARD_CONTROLS;
```
toddlees, line 18: should we include 
```
override isRefundAggregate = true;
```

toddlees (changes requested):

front-end/src/app/shared/models/transaction-groups/transaction-group-a.model.spec.ts
dheitzer: ah true - I overload the term to be 'return value' or 'returned value', but that could be confusing.

front-end/src/app/shared/models/transaction-types/FEDERAL_ELECTION_ACTIVITY_VOID.model.ts
dheitzer: @toddlees I think this may not be needed since showAggregate is false and isRefundAggregate is only used inside this if statement https://github.com/fecgov/fecfile-web-app/blob/99e3d7078cf3fb5c1f939655057fc37e682cb1be/front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts#L75

Should we set it in case we decide to display the aggregate?

toddlees: ah i see. i think you're right to not include it.  If the property was called isRefund, i think i'd say we should add it, but as it's named, it only pertains to the aggregate which is not shown

toddlees (approved):

Issue #738, opened 2022-11-16
Allow user to enter Void of 100% Federal Election Activity on Schedule B (Group O)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Void of 100% Federal Election Activity Payment”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a disbursement
**Then** Click(open) on the " Federal Election Activity Expenditures" Accordion link  
**Then** I can enter a transaction with the type “Void of 100% Federal Election Activity Payment"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE**
Dropdown options: Organization, Individual, Committee 
*Note - please default to Organization.

---
_if CONTACT TYPE = Committee_
-	**COMMITTEE NAME** (Required)
*Note - Committee ID not needed for this transaction

---
_if CONTACT TYPE = Organization_
-	**ORGANIZATION NAME** (Required)

---
_if CONTACT TYPE = Individual_
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional, will be required if DATE falls outside of report dates)
-	**AMOUNT** (Required)
**Note: System will automatically make the amount negative. AMOUNT MUST be a non-zero negative value to save, and max value is -$0.01. If amount exceeds that, then display error message 'Amount must be negative (example: -$20.00)'**
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL_DISBURSEMENT"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# CANDIDATE/COMMITTEE INFORMATION 
#

[...]

Issue #735, opened 2022-11-16
Allow user to enter Staff Reimbursement Memo for 100% Federal Election Activity on Schedule B (Group S)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Staff Reimbursement Memo for 100% Federal Election Activity”** sub-transaction under **”Staff Reimbursement for 100% Federal Election Activity”** on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator 
**And** I am editing a Staff Reimbursement for 100% Federal Election Activity transaction on an open report
**Then** I will have the _option_ to enter the following sub-transaction: 
- Staff Reimbursement Memo for 100% Federal Election Activity

**And** I will be required to enter the following information (unless otherwise noted):

 Contact 
#
-	**CONTACT TYPE**
Dropdown options: Organization, Individual, Committee 
*Note - please default to Organization.

---
_if CONTACT TYPE = Organization_
-	**ORGANIZATION NAME** (Required)

---
_if CONTACT TYPE = Committee_
-	**COMMITTEE NAME** (Required)

---
_if CONTACT TYPE = Individual_
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Required, must be checked)
-	**AMOUNT** (Required, can be negative)
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL_DISBURSEMENT"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# ELECTION INFORMATION 
#
-	**ELECTION CODE** (Required)
-	**ELECTION YEAR** (Required)
-	**ELECTION  DESCRIPTION** (Required if Election Code is Other)

# Additional information 
#
[...]

Issue #734, opened 2022-11-16
Allow user to enter Staff Reimbursement for 100% Federal Election Activity on Schedule B (Group A)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Staff Reimbursement for 100% Federal Election Activity”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a disbursement
**Then** Click(open) on the " Federal Election Activity Expenditures" Accordion link  
**Then** I can enter a transaction with the type “Staff Reimbursement for 100% Federal Election Activity"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE**
-Individual Only

-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional, will be required if DATE falls outside of report dates)
-	**AMOUNT** (Required, can be a negative)
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL_DISBURSEMENT"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# Additional information 
#
-	**PURPOSE OF DISBURSEMENT** (Required, System generated)
--Text must read: "Reimbursement: See Below"
-	**NOTE OR MEMO TEXT** (Optional)	
-	**CATEGORY CODE** (Optional)

---
**Then** I will be able to:
- Cancel
- Save & Add Memo (Memo Name: Staff Reimbursement Memo for 100% Federal Election Activity)
- Save
- Edit

**And if** the Aggregate amount is >= $200.01, 
[...]

Issue #731, opened 2022-11-16
Allow user to enter 100% Federal Election Activity Payment on Schedule B (Group O)


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”100% Federal Election Activity Payment”** transaction on Schedule B of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** Add a disbursement
**Then** Click(open) on the " Federal Election Activity Expenditures" Accordion link  
**Then** I can enter a transaction with the type “100% Federal Election Activity Payment"
**Then** I will be required to enter the following information (unless otherwise noted):

---

 Contact 
#
-	**CONTACT TYPE**
Dropdown options: Organization, Individual, Committee
*Note - please default to Organization.

---
_if CONTACT TYPE = Committee_
-	**COMMITTEE NAME** (Required)

---
_if CONTACT TYPE = Organization_
-	**ORGANIZATION NAME** (Required)

---
_if CONTACT TYPE = Individual_
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure Information 
#
-	**DATE** (Required)
-	**MEMO ITEM** (Optional, will be required if DATE falls outside of report dates)
-	**AMOUNT** (Required, can be a negative)
-	**AGGREGATE** (Required)
-- Amount Displayed will be the total of contributions that are:
--- From the person (ENTITY) 
-- within Aggregation Group of "GENERAL_DISBURSEMENT"
--- _from_ January 1 of report year _to_ the Contribution Date* of the transaction. 
--- *If there are two transactions with the same Contribution Date for the contributor, use date/time stamp transaction was created.

# CANDIDATE/COMMITTEE INFORMATION 
#

**And** the user will have the OPTION to select a Candidate using the **LOOKUP,** and upon selection the following will be populated:

-	**CANDIDATE ID** (Optional)
-	**LAST NAME** (Required if CANDIDATE ID is populated)
[...]

PR #1190, opened 2023-06-27
Fix missing and display of aggregate input box

mjtravers:
#738 
#731 
#735 
#734

### DIFF
--- a/front-end/src/app/shared/models/transaction-types/FEDERAL_ELECTION_ACTIVITY_VOID.model.ts
+++ b/front-end/src/app/shared/models/transaction-types/FEDERAL_ELECTION_ACTIVITY_VOID.model.ts
@@ -12,7 +12,6 @@ export class FEDERAL_ELECTION_ACTIVITY_VOID extends SchBTransactionType {
   title = LabelUtils.get(ScheduleBTransactionTypeLabels, ScheduleBTransactionTypes.FEDERAL_ELECTION_ACTIVITY_VOID);
   schema = schema;
   override negativeAmountValueOnly = true;
-  override showAggregate = false;
   override isRefundAggregate = true;
   override purposeDescriptionLabelSuffix = PurposeDescriptionLabelSuffix.REQUIRED;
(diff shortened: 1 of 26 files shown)


front-end/src/app/shared/models/transaction-types/FEDERAL_ELECTION_ACTIVITY_VOID.model.ts
```diff
@@ -12,7 +12,6 @@ export class FEDERAL_ELECTION_ACTIVITY_VOID extends SchBTransactionType {
   title = LabelUtils.get(ScheduleBTransactionTypeLabels, ScheduleBTransactionTypes.FEDERAL_ELECTION_ACTIVITY_VOID);
   schema = schema;
   override negativeAmountValueOnly = true;
-  override showAggregate = false;
```
dheitzer, line 15: @toddlees had suggested adding `isRefundAggregate = true;` in the original CR which I think we should add now since we are showing the aggregate (it's n/a otherwise).

mjtravers: @toddlees @dheitzer I am not convinced that this is the case. For refunds, this subtraction from the aggregate is explicitly stated as a requirement in the tickets. See here for an example: https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/723

I don't believe showing the aggregate on the form and subtracting the value of the transaction from the aggregate are tied so closely together. Maybe let's jump on a video about it.

mjtravers: After looking at the ticket with @toddlees it was discovered that nether of these tickets is a refund so the isRefundAggregate flag should be false.

Elaine-Krauss-TCG (approved):

PR #1295, opened 2023-07-25
Feature/1283

dheitzer:
removed debouncetime as it was not working and causing multiple requests to be sent.  delay() worked but caused sporadic unit test failures as discussed here https://github.com/angular/angular/issues/44186, so decided to remove it altogether as I didn't want to set `teardown: { destroyAfterEach: false }` globally


front-end/src/app/shared/services/contact.service.ts
```diff
@@ -85,19 +85,24 @@ export class ContactService implements TableListService<Contact> {
...
     return of(control.value).pipe(
-      debounceTime(500),
       switchMap((fecId) =>
-        this.checkFecIdForUniqness(fecId).pipe(
-          map((isUnique: boolean) => {
-            return isUnique ? null : { fecIdMustBeUnique: true };
-          })
-        )
+        this.checkFecIdForUniqness(fecId, (control.parent?.get(
+          'contact_1')?.value as Contact)?.id).pipe(
```
toddlees: is contact_1 available when editing a contact on the manage contacts page (not in a transaction form)?

mjtravers: I'm a bit concerned that we are hardcoding 'contact_1' in the contact service and the contact form component which ideally should be agnostic as to which transaction contact record they are working with. Can the contact form work with additional contacts beyond the first one declared for a transaction?

mjtravers (changes requested):
The contact components managing the lookup and contact form needs a review to make sure it is working with contacts beyond just contact_1.

toddlees (approved):

mjtravers (approved):

Issue #1300, opened 2023-07-28
4 - Allow user to View Loan guarantors assigned to a loan


GreggMoreland:
Business Reason

As a Committee Administrator, I will be able to see a table of guarantors associated with a loan and see the details of a guarantor so that I can accurately see the details of my loans.

 Acceptance Criteria

**Given** I am a Committee Administrator editing an open Form 3X report
**And** I am editing a **"Loan Received from Individual"**, **”Loan Received from Bank”**, or **”Loan by Committee”** transaction on an open report
**Then** I will see a table that lists all the guarantors associated with my loan with the following columns:

- Name (includes a link to view the guarantor record)
- Guaranteed financial information amount
- Actions (Edit)

**And** when I click on one of the guarantors in the list, I will review the following information for the guarantor selected:

 Guarantor 
#
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)
-	**EMPLOYER** (Required if **Amount** > $200)
-	**OCCUPATION** (Required if **Amount**  > $200)

# Guaranteed Financial Information 
#
-	**AMOUNT** (AMT-12, Required, can be negative)

---
**And** when I click on the 'Edit' action next to one of the guarantors in the list, I will be able to view and update the above information for the guarantor selected
**Then** I will be able to:
- Cancel
- Save & add loan guarantor
- Save (goes back to transactions table)

 Wireframes

**Schedule C with Guarantor table**
![1000 - Loan Received From Individual - with Gaurantor table.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/0f6e6617-75f2-4e2e-9412-85b2aed3908e)

**When Guarantor in table is clicked or Edit action next to a guarantor is clicked**
![1012 - Guarantor – 1.png](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/94894812-32d2-4e95-b9b0-2d3f18f54d15)

### QA Notes ### 
  DEV Notes

### DEV Notes ### 
 null
### Design ### 
 null

AureliaKhorsand:
[...]

PR #1348, opened 2023-08-30
Feature/1300 view loan guarantors

Elaine-Krauss-TCG:
#1300


front-end/src/app/reports/transactions/transaction-children-tables/transaction-children-tables.component.html
```diff
@@ -0,0 +1,6 @@
+<ng-container *ngIf="transaction?.transactionType?.showChildrenTable">
```
mjtravers: My preference is to wrap the hide/show conditional logic around the HTML tag that is being affected so devs looking at the template can see in the parent if the tag is visible and not have to know or dig around in the component itself to see the *ngIf clause. Please move these out to to wrap the <app-transaction-children-tables> tags in the detail components.

mjtravers (approved):

Issue #1332, opened 2023-08-23
When creating a new contact the "Required Field" text boxes are displaying in red upon New Contact page being displayed.


WiseQA:
During testing when creating a new contact the "Required Field" text boxes are displaying in red upon New Contact page being displayed. This bug was found when testing Contribution to Candidate (Schedule B) in Sprint 30.

This bug works the same way if you have an Individual transaction or tier transaction.

The steps below is what is being displayed if you are viewing the Contribution to Candidate transaction.

1\. Under Committee/Candidate information if the required field example "Candidate information is required"

[Private Zenhub Image]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb0F2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--5bcb8dd36bb56c9de5049fb393ba29fca5a7e7d5/image.png](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb0F2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--5bcb8dd36bb56c9de5049fb393ba29fca5a7e7d5/image.png))

2\. Select "Create a new contact"

[Private Zenhub Image]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb2t2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--ff7a87b59be24f5015ab80b5c3cd4f6de1bf54a3/image.png](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb2t2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--ff7a87b59be24f5015ab80b5c3cd4f6de1bf54a3/image.png))

3\. Create a new contact page is displayed and the required fields are highlighted in red, (This is the bug error)

[...]

PR #1362, opened 2023-09-11
Bug/1332 Contact dialog required field messages set on form open

mjtravers:
#1332


Elaine-Krauss-TCG (commented):
The core of the ticket is working beautifully, but I noticed a bug where the X at the top-right corner of the contact-dialog modal does not close the modal.

mjtravers:
> The core of the ticket is working beautifully, but I noticed a bug where the X at the top-right corner of the contact-dialog modal does not close the modal.

Good catch! The close X button in the header has been fixed.

Elaine-Krauss-TCG (approved):

PR #1371, opened 2023-09-13
FEC-1329: removed contactLookupLabels from all models except the base…

andrew-bowers-tcg:
Updated contactLookupLabels to use "LOOKUP" globally. Removed excess contactLookupLabels overrides on individual models


front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="grid" *ngIf="isEditable && !!!transactionType?.useParentContact">
   <div class="col-12">
     <div class="field">
-      <label for="entity_type">{{ transactionType.contactLookupLabel ?? contactLookupLabel }}</label>
+      <label for="entity_type">{{contactLookupLabel}} </label>
```
toddlees: Since this is always going to Be `LOOKUP` can we remove the whole variable provenance leading up to this point?

toddlees (changes requested):

front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
andrew-bowers-tcg: Sure. I will update it.

front-end/src/app/shared/models/transaction-type.model.ts
```diff
@@ -75,7 +76,7 @@ export abstract class TransactionType {
   formTitle?: string; // Title of form within accordion section
   footer?: string; // Text at the end of form
   contactTitle?: string; // Title for primary contact
-  contactLookupLabel?: string; //Label above contact lookup
+  contactLookupLabel = "LOOKUP"; // Set label for all lookups
```
toddlees: Recommend removing this property from transaction type models if its not variable any longer

toddlees (approved):

PR #1383, opened 2023-09-20
FEC-1296: updated radio button styles

andrew-bowers-tcg:
Updated radio button styling based on wireframe/mock.

### DIFF
--- a/front-end/src/assets/styles/theme.css
+++ b/front-end/src/assets/styles/theme.css
@@ -1539,7 +1539,6 @@ p-password.p-password-clearable.p-password-mask .p-password-clear-icon {
 .p-radiobutton {
   width: 20px;
   height: 20px;
-  padding-bottom: 5px !important;
 }
 .p-radiobutton .p-radiobutton-box {
   border: 2px solid #5b616b;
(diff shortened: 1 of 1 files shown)


front-end/src/assets/styles/theme.css
```diff
@@ -1539,13 +1539,14 @@ p-password.p-password-clearable.p-password-mask .p-password-clear-icon {
 .p-radiobutton {
   width: 20px;
   height: 20px;
+  padding-bottom: 5px !important;
```
mjtravers: I find using !important to be dangerous. Is there a way to rework the precedence of that padding-bottom directive without using !important? Order of declaration in the style sheet file is one way to affect precedence.

mjtravers (changes requested):
See inline comment about the use of !important

front-end/src/assets/styles/theme.css
andrew-bowers-tcg: The !important line there was an accidental commit. Should be removed now. Also I will refrain from using it in the future!

mjtravers (approved):

Issue #1069, opened 2023-05-11
6 - System to pull outstanding loan transaction(s) forward when a new report is created


GreggMoreland:
Business Reason

As a user, I will be able to see any outstanding Schedule C loans pull forward onto any new report I create until that loan balance is $0 so that I can continue to report outstanding loans and associate repayments as needed.

Acceptance Criteria

***Given*** a user has Schedule C loan(s) on a previous report,

  ***When*** the user creates a new report in the future,

  ***Then*** the Schedule C Loan(s) will be pre-populated in the report along with any associated Guarantor information (C2s) if they meet the following conditions:

*   loan balance ***other than $0***
*   loan is _not_ marked as a memo item.

***And*** the transaction ID will be the same as the original.

—

***Given*** the following transactions have pulled forward,

  ***then*** the fields will not change rules regarding if they can be edited or not (i.e., if they were editable before, they are editable now):

*   Loan Received from Bank
*   Loan Received from Individual
*   Loan by Committee

Wireframe

  ![1069-System to pull outstanding loan forward.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7f41d071-a68e-418d-aa13-919fb16f691c](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7f41d071-a68e-418d-aa13-919fb16f691c))

### QA Notes ###
- Note that in order for the loan to not pull forward, it either must be marked as a memo item or have a balance of _exactly_ $0, not more or less.
- This will NOT include the auto-generated Schedule A or Schedule B transactions or the C1 (Loan Agreement) created when the user first saved these loans on the original report.
- The transaction ID must be the same as the original.

### DEV Notes ###
- "Date Incurred " will NO LONGER have the validation that it must be within report coverage dates. 
- The transaction ID must be the same as the original. 
- A new field will be added to the transaction table that will hold the UUID of the originating transaction for copied forward transactions. We can use this as a flag so the transaction type code knows the transaction is a pulled forward transaction

### Design ###
null

 See full ticket and images here: [FECFILE-774](https://fecgov.atlassian.net/browse/FECFILE-774)

MitchellTCG:
@AureliaKhorsand 
Please review

MitchellTCG:
@GreggMoreland 
This appears to only be a reference wire but all looks good. 
UX review complete

GreggMoreland:
BA Review Complete

mjtravers:
[...]

PR #1388, opened 2023-09-22
Feature/1069 Add loan pull forward feature

mjtravers:
#1069

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
@@ -26,24 +26,10 @@ export class TransactionContainerComponent extends DestroyerComponent {
     });
   }
 
-  isSingleTransaction(): boolean {
-    return (
-      isPulledForwardLoan(this.transaction) ||
-      !this.transaction?.transactionType?.dependentChildTransactionTypes?.length
-    );
-  }
-
-  isDoubleTransaction(): boolean {
-    return (
-      !isPulledForwardLoan(this.transaction) &&
-      this.transaction?.transactionType?.dependentChildTransactionTypes?.length === 1
-    );
-  }
-
-  isTripleTransaction(): boolean {
-    return (
-      !isPulledForwardLoan(this.transaction) &&
-      this.transaction?.transactionType?.dependentChildTransactionTypes?.length === 2
-    );
+  transctionCardinality(): number {
+    if (isPulledForwardLoan(this.transaction)) {
+      return 1;
+    }
+    return (this.transaction?.transactionType?.dependentChildTransactionTypes?.length ?? 0) + 1;
   }
 }
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
@@ -3,7 +3,7 @@ import { ActivatedRoute } from '@angular/router';
 import { takeUntil } from 'rxjs';
 import { Store } from '@ngrx/store';
 import { Title } from '@angular/platform-browser';
-import { Transaction } from 'app/shared/models/transaction.model';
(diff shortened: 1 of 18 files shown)


front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
```diff
@@ -27,14 +27,23 @@ export class TransactionContainerComponent extends DestroyerComponent {
   }
 
   isSingleTransaction(): boolean {
-    return !this.transaction?.transactionType?.dependentChildTransactionTypes?.length;
+    return (
```
toddlees: Nice to have:
how about one function 
```
transctionCardinality(): number {
  if(isPulledForwardLoan(this.transaction) return 1;
  return (this.transaction?.transactionType?.dependentChildTransactionTypes?.length ?? 1)+ 1
}
```

then the html would be `*ngIf="transactionCardinality() === 1 "` and so on in

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.ts
```diff
@@ -55,14 +57,16 @@ export class LoanTermsDatesInputComponent extends BaseInputComponent implements
...
-    this.store
-      .select(selectActiveReport)
-      .pipe(take(1))
-      .subscribe((report) => {
-        this.form
-          .get(this.templateMap.date)
-          ?.addValidators(dateWithinReportRange(report.coverage_from_date, report.coverage_through_date));
-      });
+    if (!isPulledForwardLoan(this.transaction)) {
+      this.store
```
toddlees, line 61: i'm starting to think we should have the active transaction in the store so we don't have to pass it through all the component inputs

toddlees: then we could also access it as an observable and act on it as it changes.  Definitely not a thing to do right now but a thought

front-end/src/app/shared/models/schc-transaction-type.model.ts
```diff
@@ -7,6 +9,20 @@ export abstract class SchCTransactionType extends TransactionType {
   // Labels
   override amountInputHeader = 'Loan information';
 
+  override getNavigationControls(transaction: Transaction): TransactionNavigationControls | undefined {
+    if (isPulledForwardLoan(transaction) && this.navigationControls) {
+      this.navigationControls.continueControls = [SAVE_LIST_CONTROL];
```
toddlees: I changed the order of the test in the LOAN_RECEIVED_FROM_INDIVIDUAL so that it tested a pull-forward before a new loan, and this mutation persisted.  Anyway you can return a new navigationControls object instead of changing the state of the one tied to the transaction type instance?

```
if(isPulledForwardLoan(transaction) && this.navigationControls) {
return new TransactionNavigationControls(this.navigationControls.inlineControls, this.navigationControls.cancelControls, [SAVE_LIST_CONTROL]);
}
```

toddlees (changes requested):

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.ts
mjtravers: Agreed. We have our fancy ngrx state manager and haven't been leveraging it much

front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
mjtravers: Oooh. I've added it in.

front-end/src/app/shared/models/schc-transaction-type.model.ts
mjtravers: Much better. Done.

toddlees (approved):

Issue #1366, opened 2023-09-12
Remove 'force itemize/unitemize' function from Schedule C and Schedule D transactions


GreggMoreland:
Business Reason

As a user, I should NOT be able to force itemize/unitemize Schedule C and Schedule D transaction so that I can accurately report on my committee transactions.

Acceptance Criteria

***Given*** an existing Schedule C or Schedule D transaction in the Loans and Debts Transactions table of a report

  ***When*** the user selects the action menu for that transaction,

  ***Then*** they will NOT see an option to manually "itemize" or "unitemize" that transaction.

### QA Notes ###
- [x] BA/Design Quality Checklist complete

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-627](https://fecgov.atlassian.net/browse/FECFILE-627)

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified under the "Actions" dropdown menu "Itemize" and/or "Unitemize" option for Loans (Schedule C) and Debts (Schedule D) transactions have been removed.  

Dropdown for "Loan Received from Bank" (showing as example of dropdown)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdlA4IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--23405664bdae34e2b866a781ccd6abe2ff5c326c/image.png)

Dropdown for "Loan By Committee"  (showing as example of dropdown)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdkg4IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--0563cd4ecd62daa35bdaa338ae60f10d5b99f91f/image.png)

Transactions tested:

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdXI4IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--335bab3a09892974b7fc5b89f182bf704ddadeee/image.png)

[...]

PR #1396, opened 2023-09-26
Feature/1366 Remove itemize/unitemize options from Sch C & D

mjtravers:
#1366 
Remove the "itemize" and "unitemize" options from the action select on the Loans & Debts transaction table.

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
@@ -45,7 +45,8 @@ export abstract class TransactionListTableBaseComponent extends TableListBaseCom
         transaction.itemized === false &&
         this.reportIsEditable &&
         !transaction.parent_transaction &&
-        !transaction.parent_transaction_id,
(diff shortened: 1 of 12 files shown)


front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -45,7 +45,8 @@ export abstract class TransactionListTableBaseComponent extends TableListBaseCom
         transaction.itemized === false &&
         this.reportIsEditable &&
         !transaction.parent_transaction &&
-        !transaction.parent_transaction_id,
+        !transaction.parent_transaction_id &&
+        !['C', 'D'].includes(transaction.transactionType.scheduleId),
```
toddlees: a shame we don't have constants for these

toddlees (approved):

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
mjtravers: an enum, ScheduleIds, has been added

Issue #798, opened 2022-12-19
Allow user to unaggregate and aggregate a transaction on Schedule A


GreggMoreland:
Business Reason

As a user, I will be able to "unaggregate" (then with the option to aggregate) a transaction so that I can provide additional information for a previous transaction without double-counting the amount for the associated entity.

Acceptance Criteria

***Given*** the user has a Schedule A transaction

  ***When*** the user takes the action to "unaggregate" that transaction

  ***Then*** the transaction amount will _not be included_ in the entity's total aggregation amount.

***And*** the user can re-aggregate the transaction if needed.

DEV notes

*   In *.fec file, there is nothing to indicate that a transaction is aggregated or not. It's just not added in the aggregation for that entity.

Wireframes

1.  Unaggregate/Aggregate in Action Kebab

       ![798 Unaggregate - Select.png]([https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7864492e-50e9-4a2a-b9b8-2a92ffe3f16e](https://images.zenhubusercontent.com/619e83beec2d9fadb74ad6d1/7864492e-50e9-4a2a-b9b8-2a92ffe3f16e))

1.  Unaggregate is denoted under the transaction name like Unitemized (previously had a different hex code, is now the default 212121)
    [Private Zenhub Image]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaFJyIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--9883e63e916b68a27e6c598d71733041fa5c24a6/798%20Unaggregate%20-%20Displayed.png](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaFJyIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--9883e63e916b68a27e6c598d71733041fa5c24a6/798%20Unaggregate%20-%20Displayed.png))

1.  Version showing a transaction both Unaggregated and Unitemized
[...]

PR #1398, opened 2023-09-26
Feature/798 Add force unaggregate to schedule A

mjtravers:
#798


toddlees (changes requested):
we calculate a preview of the aggregate on the transaction form. is it supposed to not factor in the contribution amount if the transaction is unaggregated.  e.g. if the previous transaction aggregated to $200, and the user is editing a $100, unaggregated transaction, should the aggregate show $200?

mjtravers:
> we calculate a preview of the aggregate on the transaction form. is it supposed to not factor in the contribution amount if the transaction is unaggregated. e.g. if the previous transaction aggregated to $200, and the user is editing a $100, unaggregated transaction, should the aggregate show $200?

Ah, I will correct that. No, I don't think the JIT update in the UI should update the aggregate if they've selected force unaggregate. I will correct that.

Confirmed by Gregg: If user has manually aggregated or unaggregated a transaction then the system should not dynamically show something different.

front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
```diff
@@ -89,7 +89,7 @@ export class TransactionFormUtils {
         });
     }
 
-    if (transactionType.showAggregate) {
+    if (transactionType.showAggregate && !transaction.force_unaggregated) {
```
toddlees, line 92: looks like the aggregate still needs to show for these transactions.  the difference being that we don't add the amount of the unaggegated transaction to the aggregate.  seems like we could update `updateAggregate` to add $0 if the transaction is unaggregated

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
mjtravers: I believe that is how it is working currently. The !transaction.force_unaggregated above simply keeps values and changes made to the Amount in the form from adding to the displayed Aggregate, which is the value from the previous transaction.

toddlees (approved):

PR #1410, opened 2023-10-03
Fix race condition in Sch D e2e test

mjtravers:
We were getting periodic fails from the nightly e2e tests for the SchD test. It was being caused by an element selector that was finding the wrong element on the page being viewed prior to the one being tested. The selector was improved to be more specific.


andrew-bowers-tcg (approved):
Looks good. Better to check for the proper page elements instead of waiting and assuming they appear

Issue #1386, opened 2023-09-20
System to  Calculate 'CALENDAR YEAR-TO-DATE' aggregate for Schedule E Tier 1 transactions


GreggMoreland:
Business Reason
The aggregate 'CALENDAR YEAR-TO-DATE' field must display the total contributions from the calendar year of the report _per election._ This is necessary to track contribution limits. This amount will be a _rolling total_. This will only apply to Tier 1 transacitons on Schedule E.

_Please note that this is not the same thing as the entity aggregate, which will be calculated to determine itemization on relevant transactions in Schedule E._

 Acceptance Criteria

**Given** I am a Committee Administrator editing an open Form 3X report
**Then** go to Add a disbursement
**Then** I can enter a transaction with the type “Independent Expenditure" under the "Independent Expenditures" Accordion link  
**And** the following will be automatically calculated by the system:

**CALENDAR YEAR-TO-DATE** field must meet the following criteria:

-- Amount Displayed will be the total of contributions that are:
--- For the selected ELECTION TYPE, ELECTION YEAR, CANDIDATE OFFICE, CANDIDATE STATE, and CANDIDATE DISTRICT
--- within Aggregation Group of "INDEPENDENT_EXPENDITURE"
--- _from_ January 1 of report year _to_ the Disbursement Date of the transaction. 
--- (**IF** Disbursement Date is not entered, then use Dissemination Date of the transaction.)

AND when saving a transaction, any subsequent transactions for the same ELECTION CODE, ELECTION YEAR, CANDIDATE OFFICE, CANDIDATE STATE, and CANDIDATE DISTRICT will update. (E.g., if a previous transaction receipt amount is _changed_, it will affect the future aggregations). 

### QA Notes ### 
 - [x] BA/Design Quality Checklist complete
- This field applies to Schedule E transactions on Line 24. 

Please verify this functionality is working in the following transactions:
-  [ ] Independent Expenditure
-  [ ] Credit Card Payment for Independent Expenditure
-  [ ] Staff Reimbursement for Independent Expenditure
-  [ ] Payment to Payroll for Independent Expenditure
-  [ ] Void of Independent Expenditure
### DEV Notes ### 
  Wireframe
[...]

PR #1440, opened 2023-10-16
Feature/1386 calendar ytd calculation

Elaine-Krauss-TCG:
#1386 

[API PR](https://github.com/fecgov/fecfile-web-api/pull/557)


front-end/src/app/shared/services/transaction.service.ts
```diff
@@ -77,6 +77,53 @@ export class TransactionService implements TableListService<Transaction> {
...
+
+    if (transaction && actionDateString && candidate_office && aggregation_group && election_code) {
+      const params: { [key: string]: string } = {
+        transaction_id,
+        date: actionDateString,
+        aggregation_group,
+        election_code,
+        candidate_office,
+      };
+      if (candidate_state) {
```
toddlees, line 106: you can go ahead and just assign state and district in the params.  undefined will have the same behavior as not assigning them:

```
const params:  { [key: string]: string } = {
...
candidate_office,
candidate_state,
candidate_district
};
```

toddlees (changes requested):

front-end/src/app/shared/services/transaction.service.ts
Elaine-Krauss-TCG: This is actually in here to deal with a type error.  The API service's implementation of .get does not accept undefined values for parameters
image

toddlees (approved):

PR #1512, opened 2023-11-09
feature/1477 - Update so Debt Repayment will never show for Schedule D Transactions.

sasha-dresden:


Elaine-Krauss-TCG (approved):

dheitzer:
Might be worth noting that I had to adjust this same logic for 1391 and may have ended up fixing it.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
```diff
@@ -83,8 +84,7 @@ export abstract class TransactionTypeBaseComponent implements OnInit, OnDestroy
         });
     }
 
-    this.isDebtRepayment = !!this.transaction.debt_id;
-
+    this.isDebtRepayment = !!this.transaction.debt_id && !Object.keys(ScheduleDTransactionTypes).includes(this.transaction.transaction_type_identifier ?? "");
```
mjtravers: @sasha-dresden We have a new function to determine whether a transaction is a debt repayment that we can leverage here. See: https://github.com/fecgov/fecfile-web-app/blob/develop/front-end/src/app/shared/models/transaction.model.ts#L190-L192

mjtravers (changes requested):
See inline comment for requested change.

front-end/src/app/shared/components/transaction-type-base/transaction-type-base.component.ts
sasha-dresden: Oh that's good to know. At first I thought I could use that without any changes to the code but since the function and the variable had the exact same name, the variable was overriding the function. Updated code to remove the variable and simply call the function directly.

mjtravers (approved):

PR #1540, opened 2023-11-21
Feature/1523 - Turn action-kebab into a bootstrap dropdown to facilitate better 508 keyboard compliance.

sasha-dresden:
I also updated some of the logic to better reflect the usage. Converting a function input into an output. And cleaned up some code to reduce some issues that came up from circle/sonarcloud.

### DIFF
--- a/front-end/src/app/layout/header/header.component.spec.ts
+++ b/front-end/src/app/layout/header/header.component.spec.ts
@@ -5,11 +5,13 @@ import { testMockStore } from 'app/shared/utils/unit-test.utils';
 import { MenubarModule } from 'primeng/menubar';
 import { RouterTestingModule } from '@angular/router/testing';
 import { HeaderComponent } from './header.component';
+import { toggleSidebarVisibleAction } from "../../store/sidebar-state.actions";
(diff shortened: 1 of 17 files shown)


mjtravers (dismissed):

sasha-dresden:
The vertical alignment issue was due to a long string of changes. To make the kebab buttons tabable, I switched to a dropdown as opposed to an overlay panel which is unfortunately not the most 508 friendly and a dropdown fit with the concept of how these buttons were behaving. However, the benefit to the overlay was that it appeared outside the table. With the dropdown menu it caused the table to expand in size. To fix that I had to make it's position absolute and slightly adjust the vertical positioning to get it centered in the row, which unfortunately caused an issue with the "Add new transaction" button since it used the same logic. I added an Input option to determine if it should use absolute positioning or not to fix that issue.

sasha-dresden:
I forgot to update the unit test to reflect the change that reports side panel would not be visible at the base of the reports module. Added a new unit test to confirm that and updated the old unit test to go to a view that would have it.

mjtravers (approved):

Issue #1543, opened 2023-11-22
Fix Line column in "Transactions in this report" table to initially display ascending order when table is populated


WiseQA:
During QA testing and DEV review the Line column in "Transactions in this report" table under the Loans and Debts is displaying the order descendng when the "Transactions in this report" table is populated within a transaction.

Fix = The table should be displaying the Line column in Ascending order (e.g., 9, 10). Currently when "Transactions in this report" table is populated the Line column is displaying in decending order (e.g., 10, 9)

[Private Zenhub Image]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOS85QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b18e72455abb451a762b0513fb7a242d447d785d/image.png](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOS85QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b18e72455abb451a762b0513fb7a242d447d785d/image.png))

(NOTE: Sort is working correctly)

@MitchellTCG @mjtravers fix the bug for Ascending order display in line column under Loans and Debts in the transactions table. Bug fix pointed as a 2 point bug fix and assigned to Sprint 35.

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-546](https://fecgov.atlassian.net/browse/FECFILE-546)

dheitzer:
Line is a character field ("11(a)(i)" for example) and is currently being sorted in lexicographical order.

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified the Fix for "Transactions in this report" table is displaying the Line column in Ascending order (e.g., 9, 10) when "Transactions in this report" table is populated.

<img width="246" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDgwQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--19f25fa07d51b5c743557079cca767cbf57a8cf0/image.png" alt="image.png" />

[...]

PR #1552, opened 2023-11-28
feature/1543

dheitzer:
#1543 

we're just going to use a numerical sort order to address this

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/transaction-receipts/transaction-receipts.component.html
+++ b/front-end/src/app/reports/transactions/transaction-list/transaction-receipts/transaction-receipts.component.html
@@ -80,12 +80,7 @@
           <div *ngIf="item?.itemized === false"><strong>Unitemized</strong></div>
         </td>
         <td>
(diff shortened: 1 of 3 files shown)


front-end/src/app/reports/transactions/transaction-list/transaction-receipts/transaction-receipts.component.html
```diff
@@ -80,7 +80,12 @@ <h5 class="gandhi-bold size-24">Receipts</h5>
           <div *ngIf="item?.itemized === false"><strong>Unitemized</strong></div>
         </td>
         <td>
-          {{ item.name }}
+          {{
```
toddlees: `name` was not being serialized correctly in the api.  it's fixed in the e2e ticket and can be used again here (since we're sorting on the name value)

toddlees (dismissed):

Elaine-Krauss-TCG (approved):

mjtravers (approved):

PR #1558, opened 2023-11-29
Feature/1524 - Add Caption to table for screen readers

sasha-dresden:
So, this is where PrimeNG caused a bit of an issue. I love the PrimeNG table, it is incredibly powerful. However, it has a major problem for 508 compliance in that it does not allow you to put a caption for screen readers. Confusingly, they have a caption template but it doesn't actually use the caption html tag, which is what is required for 508 compliance.

I rebuilt this table using pure html with as few changes as possible. Visually it should appear exactly the same. I did have to add some logic for handling changing the page and sorting, but thankfully those were just hitting a back end call, so I didn't have to completely design a full front end sorting algorithm, though I did notice the sorting isn't actually working for most of the fields.

### DIFF
--- a/front-end/src/app/reports/report-list/report-list.component.ts
+++ b/front-end/src/app/reports/report-list/report-list.component.ts
@@ -50,6 +50,7 @@ export class ReportListComponent extends TableListBaseComponent<Report> implemen
     public router: Router
   ) {
     super(messageService, confirmationService, elementRef);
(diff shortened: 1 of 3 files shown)


sasha-dresden:
Oh my goodness that was so much easier than refactoring the whole table. This was a little different than the other ones, there it was adding an attribute. In this case I needed to add a new child element, but there was actually native dom method for tables called createCaption() that worked perfectly

mjtravers (approved):

PR #1566, opened 2023-12-04
1525 - Style modal to match ticket request

sasha-dresden:


front-end/src/app/reports/report-list/report-list.component.html
```diff
@@ -44,7 +45,7 @@ <h5 class="m-0">Recent reports</h5>
           <p-sortIcon field="TBD"></p-sortIcon>
         </th>
         <th pSortableColumn="report_code_label" id="report-code-column" role="columnheader">
-          Type of report
+          Type ofweb report
```
Elaine-Krauss-TCG: I think this is a typo?

sasha-dresden: Oh good catch. That wasn't even supposed to be there. Should have stayed as "Type of report". Did a commit to remove that change.

mjtravers (approved):

PR #1573, opened 2023-12-06
Feature/1563 Make save buttons disable on click and re-enable after process is complete

sasha-dresden:
Also added a bunch of tests to pass the 80% mark

### DIFF
--- a/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
+++ b/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
@@ -20,7 +20,7 @@
     styleClass="navigation-dropdown"
     panelStyleClass="navigation-dropdown-panel"
     (onChange)="onDropdownChange($event)"
-    [disabled]="isDisabled()"
+    [disabled]="isDisabled() || processing"
     [group]="isGroupedDropdown"
   >
     <ng-template let-group pTemplate="group">
--- a/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
+++ b/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
@@ -7,7 +7,8 @@
     [icon]="navigationControl?.icon || ''"
     [class]="navigationControl?.ngClass || ''"
     (click)="click()"
-    [disabled]="isDisabled() || processing"
+    [disabled]="isDisabled()"
+    appSingleClick
   ></button>
   <p-dropdown
     *ngIf="controlType === 'dropdown'"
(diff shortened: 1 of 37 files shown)


front-end/src/app/shared/components/navigation-control/navigation-control.component.html
```diff
@@ -3,12 +3,11 @@
...
-    appSingleClick
     [label]="navigationControl?.label || ''"
     [icon]="navigationControl?.icon || ''"
     [class]="navigationControl?.ngClass || ''"
     (click)="click()"
-    [disabled]="isDisabled()"
+    [disabled]="isDisabled() || processing"
   ></button>
   <p-dropdown
     *ngIf="controlType === 'dropdown'"
```
toddlees, line 14: we'll want to disable this dropdown as well.  It allows the user to save and add something from the list

front-end/src/app/shared/components/transaction-type-base/triple-transaction-type-base.component.ts
```diff
@@ -119,47 +120,21 @@ export abstract class TripleTransactionTypeBaseComponent
...
-    } else {
-      this.navigateTo(navigationEvent);
-    }
+  override isInvalid(): boolean {
+    return super.isInvalid() ||
+      this.childForm_2.invalid ||
+      !this.childTransaction_2;
+  }
+
+  override get confirmation$(): Observable<boolean> {
```
toddlees, line 131: WOW! This is a really impressive way of declaring the confirmations for each part once AND in their own component.

toddlees (changes requested):

front-end/src/app/shared/components/navigation-control/navigation-control.component.html
sasha-dresden: Oh good catch. I added a commit to add the "or processing" to the dropdown as well.

toddlees (approved):

sasha-dresden:
Updated code to make use of spinner store.
This reduced the need for passing the processing variable up and down the component hierarchy and also simplified the code.
I am now reusing the single click directive which has been updated to subscribe to the spinner store for updating the disable state.
I also update the ApiService::put and ::post methods so that there is a new Boolean variable it can accept which defaults to false, but if passed as true will use the Spinner variation.

The one potential issue with all of this is that the button gets disabled immediately on click, which is important to prevent double click submissions, however, if the button click doesn't result in an api call (for example if the form fails to validate) there are no calls to re-enable the button. In order to handle this situation, I allowed the button to re-enable itself through the directive after a delay, if the spinner status is false.

mjtravers (approved):

PR #1586, opened 2023-12-13
Feature/1492 delete guarantors 

toddlees:


mjtravers (changes requested):
Unit test "should filter to only the loan guarantors" is failing.

mjtravers (approved):

PR #1589, opened 2023-12-14
Remove report store

mjtravers:


mjtravers:
The reports are needed in the store for the menu. Closing PR without merging.

PR #1591, opened 2023-12-14
1590 - Update hide/show logic of left nav

mjtravers:
The left nav was appearing on the report create pages. The patch updates the logic to hide/show the left nav based on if a SideBar state exists for the route.


front-end/src/app/layout/layout.component.html
```diff
@@ -1,39 +1,39 @@
 <app-banner></app-banner>
-<div [ngClass]="isVisible ? '': 'container'">
-    <div class="row">
-        <div class="col-xxl-2 col-xl-3 col-lg-4 col-md-5 col-sm-5  fec-background-gradient" *ngIf="isVisible">
-            <app-sidebar></app-sidebar>
-        </div>
-        <div [ngClass]="isVisible ? 'col-xxl-10 col-xl-9 col-lg-8 col-md-7 col-sm-7  ' : 'col-12'">
-            <div class="col-12 header-container">
-                <app-header></app-header>
+<div [ngClass]="showSidebar ? '' : 'container'">
```
toddlees, line 2: is this class condition necessary? seems to work with just <div>

toddlees (dismissed):

front-end/src/app/layout/layout.component.html
mjtravers: @sasha-dresden Is there a accessibility or other reason for the dynamic 'container' class?

mjtravers: @toddlees I spoke to Sasha and that ngClass is necessary to manage the responsive design properly as the screen is expanded or shrunk.

sasha-dresden: So, two things on this. 
For the dynamic container class. That's there to prevent content stretching on large screen sizes when there's no side banner. 
Two, Github really made this change confusing. Looking at what actually changed, it looks like the only actual change was all of the calls to "isVisible" were changed to "showSidebar".

Elaine-Krauss-TCG:
The sidebar disappears if you click a sidebar link while *already on* that page.

I.E. if I'm on Loans and Debts, and I click the Loans and Debts button, the sidebar goes away
image

Elaine-Krauss-TCG (approved):

PR #1593, opened 2023-12-15
1559 - Change report submit success image

sasha-dresden:


front-end/src/app/reports/reports.module.ts
```diff
@@ -1,4 +1,4 @@
-import { CommonModule } from '@angular/common';
+import { CommonModule, NgOptimizedImage } from '@angular/common';
```
mjtravers, line 1: cool

mjtravers (approved):

PR #1602, opened 2023-12-19
1563 - Update spinner logic to be less dependent on api call

sasha-dresden:
Fixed the issue with the button disable status not working correctly. I had tried to tie it directly to the call to the rest api, but also disable immediately, so it was leading to some odd behavior. I updated the logic so the loading spinner and disable happen immediately on click and then handle any bail outs by dispatching a spinner off action.

I also took the spinner state out of the local storage sync as I ran into an issue where I refreshed my browser while it was saving a transaction and the spinner kept going after it reloaded without any end. Now the on browser refresh it always defaults to false for spinner state.


toddlees (approved):
Works for me.  Turning off the spinner on destroy is really strong. Also I'm a huge fan of taking the button concern out of the services.

The only issue I could imagine is if a request hangs, there will be unexpected behavior across tabs.  That's an extremely farfetched scenario, so i'm not concerned about it as a risk

PR #1606, opened 2023-12-20
1504 - Add print preview to F99

mjtravers:
Add print preview screen to F99 report.


toddlees (changes requested):
Do you think BA and UX take it as a given that clicking "Create a report" on the sidenav will take us to the main form?  I'm not sure if we need to just make that assumption and implement it or request AC for it

mjtravers:
> Do you think BA and UX take it as a given that clicking "Create a report" on the sidenav will take us to the main form? I'm not sure if we need to just make that assumption and implement it or request AC for it

@toddlees I brought the menu issue up with Gregg and he said the left nav is still a work in progress and to leave it as is for now.

toddlees (approved):

Issue #1339, opened 2023-08-24
Allow user to submit Form 99 report 


GreggMoreland:
Business Reason

As a user, I will be able to submit a completed Form 99 report so that I can generate a *.fec file and send it to the FEC.

Acceptance Criteria

***Given*** the Form 99 report

  ***Then*** there will be the pages "Confirm information" and "Submit report" under the SIGN & SUBMIT section of left-hand navigation

***and when*** a user is on the "Confirm information" page, they will enter the following information:

*   ***EMAIL #1*** (Required)
*   ***Email #2*** (OPTIONAL)

***When*** a user is on the "Submit report" page, they will enter the following information:

1.  Treasurer Information

       DEV NOTES

Let's refactor the submit-f3x-step1 (call it submit-confirm-info), submit-f3x-step2 (call it submit-report), and submit-f3x-status (call it submit-status) components into a new reports/shared directory. Each component will have it's own directory in the reports/shared directory.

*   DATE SIGNED in the specification should work like it does on Form 3X, and populate when the user confirms submission.
*   Confirm that DEV and STAGE are configured to test with test EFO filing

1.  1.  UX Comment

             We will not change the general look and feel. No new grey box. we will add in illustration instead of old fan of 5 pages

Wireframes

1.  Confirm Information
    [Private Zenhub Image]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWNPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--04a441a415669b5bccd348d697ffb955a09bcc66/1339%20Confirm%20info.png](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWNPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--04a441a415669b5bccd348d697ffb955a09bcc66/1339%20Confirm%20info.png))

1.  Submit report
[...]

PR #1611, opened 2023-12-21
#1339 Feature/form99 submit

toddlees:


front-end/src/app/reports/f99/f99-routing.module.ts
```diff
@@ -34,7 +37,45 @@ const routes: Routes = [
     title: 'Print preview',
     component: PrintPreviewComponent,
     resolve: { report: ReportResolver, sidebar: SidebarStateResolver },
-    data: { sidebarSection: ReportSidebarSection.REVIEW },
+    data: {
+      sidebarSection: ReportSidebarSection.REVIEW,
+      getBackUrl: (report?: Report) => '/reports/f99/edit/' + report?.id,
+      getContinueUrl: (report?: Report) => '/reports/f99/submit/step1/' + report?.id,
```
mjtravers, line 43: Good idea to put this in the router.

front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
```diff
@@ -83,20 +85,20 @@ <h3>Enter password</h3>
...
           </div>
         </div>
         <div class="col-6">
           <div class="field">
-            <label for="filing_password">
+            <label for="filingPassword">
               <strong>ELECTRONIC FILING PASSWORD </strong>
             </label>
-            <input type="text" pInputText id="filing_password" formControlName="filing_password"/>
+            <input type="text" pInputText id="filingPassword" formControlName="filingPassword" />
```
mjtravers, line 98: Let's make this type="password" to hide it when being typed out.

mjtravers (approved):
Not part of scope for this ticket, but I think we should change the password input to type="password" to hide it while the user is typing it in.

front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
toddlees: side effects of setting it to type="password" moved this out of scope

Issue #1064, opened 2023-05-10
Allow user to redesignate a Tier 1 Schedule B transaction from a previous report


GreggMoreland:
Business Reason

As a user, I will be able to "Redesignate" a transaction within a previous report so that I can attribute the disbursement amount to a different election.

Acceptance Criteria

***Given*** I am viewing a ***Tier 1*** Schedule B Line 23 and 30b transaction on a previously submitted report (Submission success),

*   CONDUIT_EARMARK_OUT
*   CONTRIBUTION_TO_CANDIDATE
*   CONTRIBUTION_TO_CANDIDATE_VOID
*   FEDERAL_ELECTION_ACTIVITY_100PCT_PAYMENT
*   FEDERAL_ELECTION_ACTIVITY_CREDIT_CARD_PAYMENT
*   FEDERAL_ELECTION_ACTIVITY_PAYMENT_TO_PAYROLL
*   FEDERAL_ELECTION_ACTIVITY_STAFF_REIMBURSEMENT
*   FEDERAL_ELECTION_ACTIVITY_VOID
*   IN_KIND_CONTRIBUTION_TO_CANDIDATE
*   PAC_CONDUIT_EARMARK_OUT

       ***When*** I initiate a ‘Redesignate’ function for that transaction

       ***Then*** the system will ask me to confirm which "In progress" report of the ***same report type*** I would like to add the redesignation to, 

       ***Then*** I will be able to indicate the how much should be redesignated and to which election, as follows:

—

1.  Contact

       #

*   _(Contact will auto-pull from original contribution and cannot be edited)_

1.  Address

       #

*   _(Address will auto-pull from original contribution and cannot be edited)_

1.  Expenditure information

       #

*   ***REDESIGNATION DATE*** (Required)
*   ***MEMO ITEM*** (Required, automatically checked by the system and user cannot change it)
*   ***REDESIGNATION AMOUNT*** (Required, must be positive and cannot exceed amount of original disbursement)

1.  Committee/Candidate information

       #

*   Committee/Candidate values should match the original contribution and cannot be edited

1.  ELECTION INFORMATION

       #

*   ***ELECTION CODE*** (Required)
*   ***ELECTION YEAR*** (Required)
*   ***ELECTION DESCRIPTION*** (Required if Election Code is Other)

1.  Additional information

       #

*   ***PURPOSE OF DISBURSEMENT*** (Required, system generated) 
[...]

Issue #624, opened 2022-10-19
Allow user to reattribute a Tier 1 Schedule A transaction from a previous report


AureliaKhorsand:
Business Reason

As a user, I will be able to "reattribute" a Tier 1 transaction within a previous report so that I can reattribute part or all of the receipt amount to a different entity, changing the aggregation amount for the entity on each new transaction.

Acceptance Criteria

***Given*** a Tier 1 receipt on Schedule A in a previous report with status "Submission success"

 ***When*** I initiate the 'Reattribute' action for that transaction,

 ***Then*** the system will ask me to confirm which "In progress" report of the ***same report type*** I would like to add the reattribution to,

 ***And*** when I confirm the report selection,

 ***Then*** I will be able to indicate the how much should be reattributed, to whom, and when as follows:

1.  Contact

      #

*   ***CONTACT TYPE***- _(Required, contact type values should match the original receipt specification)_

—

 _if CONTACT TYPE = Organization_

*   ***ORGANIZATION NAME*** (Required, LOOKUP)

—

 _if CONTACT TYPE = Committee_

*   ***COMMITTEE NAME*** (Required, LOOKUP)
*   ***COMMITTEE ID*** (Required)

—

 _if CONTACT TYPE = Individual_

*   ***LAST NAME*** (Required)
*   ***FIRST NAME*** (Required)
*   ***MIDDLE NAME*** (Optional)
*   ***PREFIX*** (Optional)
*   ***SUFFIX*** (Optional)

1.  Address

      #

*   ***STREET ADDRESS*** (Required)
*   ***APARTMENT, SUITE, ETC.*** (Optional)
*   ***CITY*** (Required)
*   ***STATE/PROVINCE*** (Required)
*   ***ZIP/POSTAL CODE*** (Required)

1.  Receipt Information

      #

*   ***REATTRIBUTION DATE*** (required)
*   ***MEMO ITEM*** (required, automatically checked by the system and user cannot change it)
*   ***REATTRIBUTED AMOUNT*** (required, must be positive and cannot exceed amount of original contribution)
*   ***AGGREGATE*** (read-only, will match logic of original transaction)

1.  Additional information

      #

*   ***PURPOSE OF RECEIPT*** (Required, system generated)
      *   "Reattribution from XX", where XX is the Entity Name from the original contribution
*   ***NOTE OR MEMO TEXT*** (Optional)

—

***And*** the system will auto-generate a transaction showing amount _reattributed from_ with the following information:

[...]

PR #1635, opened 2024-01-04
Feature/1063 - Redesignation configuration 

sasha-dresden:


Elaine-Krauss-TCG (approved):

front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
```diff
@@ -1,7 +1,7 @@
 <div
   class="grid"
   *ngIf="
-    isEditable && !!!transactionType?.getUseParentContact(transaction) && !!!transactionType?.hidePrimaryContactLookup
+    isEditable && !transactionType?.getUseParentContact(transaction) && !transactionType?.hidePrimaryContactLookup
```
toddlees: 🙏 thank you! https://stackoverflow.com/a/25318045

front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
```diff
@@ -188,10 +188,12 @@ <h3>{{ transactionType.committeeCandidateHeader }}</h3>
     <ng-container *ngIf="transaction?.transactionType?.hasSupportOpposeCode()">
       <app-support-oppose-input [form]="form" [formSubmitted]="formSubmitted"></app-support-oppose-input>
     </ng-container>
-    <ng-container *ngIf="transaction?.transactionType?.hasCandidateInformation(form)">
+    <ng-container
+      *ngIf="transaction?.transactionType?.hasCandidateInformation(form) && !transactionType?.hidePrimaryContactLookup ">
```
toddlees: recommend changing `hidePrimaryContactLookup` to `hideContactLookups` because it's not just used for primary anymore.

front-end/src/app/shared/resolvers/transaction.resolver.ts
```diff
@@ -102,26 +111,26 @@ export class TransactionResolver {
...
-              return this.getNewChildTransaction(parentTransaction, childTransactionTypeName);
-            })
-          );
-        }
-        // Otherwise we just need to return an observable of the parent transaction
-        return of(this.getNewChildTransaction(parentTransaction, childTransactionTypeName));
-      })
-    );
-  }
+  // resolveNewChildTransaction(
```
lbeaufort: Can this block be removed? Stolen from https://sonarcloud.io/project/issues?id=fecgov_fecfile-web-app&pullRequest=1635&resolved=false&inNewCodePeriod=true

lbeaufort: There are a few blocks of commented out code - if they can be removed, I think it would improve readability. Thanks @sasha-dresden!

toddlees: nice job finding this stub.  I'd remove it entirely

front-end/src/app/shared/utils/reatt-redes/abstract-from.utils.ts
toddlees, line 1: is this file supposed to be here?

front-end/src/app/shared/utils/reatt-redes/reattribution-from.utils.ts
```diff
@@ -46,21 +46,42 @@ export class ReattributionFromUtils {
...
+    'state',
+    'zip',
+    'amount',
+    'purpose_description',
+    'committee_fec_id',
+    'committee_name',
+  ];
+
   public static overlayForm(fromForm: FormGroup, transaction: SchATransaction, toForm: FormGroup): FormGroup {
+    const purposeDescControl = fromForm.get(transaction.transactionType.templateMap.purpose_description);
```
toddlees: since you've abstracted the raw field name, we're free to follow good variable naming practices.  Recommend not abbreviating 'description':  `purpsoseDescriptionControl`

front-end/src/app/shared/components/transaction-type-base/double-transaction-type-base.component.ts
```diff
@@ -100,6 +99,12 @@ export abstract class DoubleTransactionTypeBaseComponent
       this.childUpdateFormWithPrimaryContact({
         value: this.transaction?.reatt_redes?.contact_1,
       } as SelectItem);
+      if ((this.transaction as SchATransaction | SchBTransaction).reattribution_redesignation_tag === ReattRedesTypes.REDESIGNATION_TO) {
```
toddlees, line 102: Is this not handled in the ReattRedesUtils.overlayForms() call in the line 92 condition?

toddlees (dismissed):
I'm not seeing the candidate information in either form
original: 
image
redesignation:
image

front-end/src/app/shared/resolvers/transaction.resolver.ts
sasha-dresden: Yea I can remove those. Those blocks were to be used for dealing with reattribution or redesignation of transactions on past reports, which will be handled on future tickets, #624 and #1064

front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
sasha-dresden: Good point. Renamed.

front-end/src/app/shared/resolvers/transaction.resolver.ts
sasha-dresden: Yea I'll just remove that.

front-end/src/app/shared/utils/reatt-redes/abstract-from.utils.ts
sasha-dresden: Ah good catch. I had originally tried making an abstract file that both the reattribution and the redesignation drew from, but then, after talking with Matt and Gregg, found out they were quite different and the common code was no longer the same. But I forgot to delete the file.
Deleted now.

front-end/src/app/shared/components/transaction-type-base/double-transaction-type-base.component.ts
sasha-dresden: Hmm, I kind of feel like it should be handled there, but it wasn't being done there for the reattributions to update the child primary contact. Line 99 above this, so I was just following that. 

I was just digging into this some more and, the overlay was only updating the transaction, but didn't update the form data for the contacts. I think the reason this is being handled here instead of in the overlay is because there are some methods that are being used here like updateInheritedFields(), which would need to be duplicated if these updates were moved to the overlay.

front-end/src/app/shared/utils/reatt-redes/reattribution-from.utils.ts
sasha-dresden: Updated to purposeDescriptionControl

front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
sasha-dresden: Ah the reason that data wasn't showing was because I had the hidePrimaryContactLookup call in the wrong place. I wanted to only hide the Lookup part of the candidate information but accidentally hid the whole thing. I moved my check to the right location and it's showing up now. I also noticed it wasn't making those fields read only so I had to make some fixes for that.

PR #1644, opened 2024-01-05
Feature/1255 - Create 2 different layouts depending on whether or not the sidebar is present

sasha-dresden:


sasha-dresden:
Updated layout so it won't apply to the login page. I also addressed an issue with the sidebar not stretching to the bottom, on larger screens.

mjtravers (approved):

PR #1658, opened 2024-01-10
submit f24

toddlees:
had to update f24 creation to define address information.


dheitzer:
I think the dev note still needs addressed:

DEV NOTES
Let's refactor the submit-f3x-step1 (call it submit-confirm-info), submit-f3x-step2 (call it submit-report),

mjtravers (approved):

PR #1676, opened 2024-01-18
Feature/624 - Reattribute a transaction from a previous report.

sasha-dresden:


toddlees (changes requested):
Reattributing and Redesignating from a previous report is somewhat different from for the current report in that it requires a new copy of the original transaction in the target report, that is unaggregated.  This means that you'll need to use a triple transaction component for this scenario
image

sasha-dresden:
Had to undergo quite a bit of a rework to incorporate pulling forward the original transaction into the new report. Unfortunately the triple transaction component we already had didn't quite fit because the pulled forward needed to be the final accordion tab and it was not a child of the Reattribution To, which was the first. I ended up creating a new component specific for Reattributions and Redesignations the extends the Double Transaction. This actually worked well in pulling out some code that was specific to those from the Double Transaction Component.

Additionally, I needed to make a new api endpoint to handle pulling the original transaction forward into the new report. So there is also a [PR695](https://github.com/fecgov/fecfile-web-api/pull/695) in the api to go along with this.

mjtravers:
This PR is OBE as the branch was merged into this #1700 
Closing PR without merging.

PR #1682, opened 2024-01-23
feature/1230 report SE repayment on f3x

dheitzer:


dheitzer:
Seeing some weirdness on the DSP so pulling back to in-progress

front-end/src/app/reports/transactions/transaction-list/transaction-loans-and-debts/transaction-loans-and-debts.component.html
```diff
@@ -61,7 +61,7 @@ <h5 class="gandhi-bold size-24">Loans and Debts</h5>
           {{ item[item.transactionType.templateMap.date] | fecDate }}
         </td>
         <td>
-          {{ item['amount'] | currency }}
+          {{ item['loans_debts_transaction_table_amount'] | currency }}
```
mjtravers: I'm wondering if this identifier should be so specific. If it comes up down the road that this value will be used elsewhere outside the transaction table as well. What about "debt_incurred_amount" since that is the origin of the amount?

mjtravers (changes requested):

front-end/src/app/reports/transactions/transaction-list/transaction-loans-and-debts/transaction-loans-and-debts.component.html
dheitzer: Yes, I think that is a great idea - pushed changes with this.  My only concern is that sometimes it could be the debt incurred amount and sometimes it could be another amount (in which case we won't care about it elsewhere), but I still think your suggestion makes the most sense since we could use it elsewhere.

dheitzer: for clarity, the corresponding api changes (notice it defaults to amount):

```
debt_incurred_amount=Case(
                    When(
                        schedule_d__isnull=False,
                        debt__isnull=False,
                        then=F("debt__schedule_d__incurred_amount"),
                    ),
                    default=F("amount"),
                ),
```

mjtravers (approved):

PR #1701, opened 2024-02-02
feature/92 - create user profile data

dheitzer:
https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/92

As part of this ticket we had to the move the place where we set the cookies from the dashboard to the main angular app component.  This is the correct place for it and is needed since the dashboard may not be the first authenticated landing page anymore.


front-end/src/app/login/security-notice/security-notice.component.ts
```diff
@@ -0,0 +1,72 @@
...
+import { singleClickEnableAction } from 'app/store/single-click.actions';
+import { map, takeUntil } from 'rxjs';
+
+@Component({
+  selector: 'app-security-notice',
+  templateUrl: './security-notice.component.html',
+  styleUrls: ['./security-notice.component.scss'],
+})
+export class SecurityNoticeComponent extends DestroyerComponent implements OnInit {
+  public loginDotGovAuthUrl: string | undefined;
```
toddlees: `loginDotGovAuthUrl` is this used?

toddlees (approved):

PR #1703, opened 2024-02-05
Feature/92-120-merge

Elaine-Krauss-TCG:
Merges work from Feature/120 into Feature/92


front-end/src/app/shared/services/login.service.ts
```diff
@@ -83,13 +81,19 @@ export class LoginService extends DestroyerComponent {
...
-      !!this.userLoginData.last_name;
+    return !!this.userLoginData?.first_name && !!this.userLoginData.last_name;
+  }
+
+  public userHasRecentSecurityConsentDate() {
+    const security_date = this.userLoginData?.security_consent_date;
+    const one_year_ago = new Date();
+    one_year_ago.setFullYear(one_year_ago.getFullYear() - 1);
+
+    return security_date !== undefined && security_date !== '' && new Date(security_date) > one_year_ago;
```
dheitzer: Can we just check truthy here? `!!security_date &&  new Date(security_date) > one_year_ago;`

dheitzer (changes requested):
Just one minor, otherwise looks good

front-end/src/app/shared/services/login.service.ts
Elaine-Krauss-TCG: Great call, implemented as suggested

dheitzer (approved):

PR #1707, opened 2024-02-06
Patch/restore route data to fix layout on login

Elaine-Krauss-TCG:


front-end/src/app/layout/header/header.component.ts
```diff
@@ -11,9 +17,12 @@ import { toggleSidebarVisibleAction } from '../../store/sidebar-state.actions';
 export class HeaderComponent {
   loginService: LoginService;
   private window = window;
-  @Input() loginHeader = false;
+  @Input() headerStyle = HeaderStyles.DEFAULT;
```
toddlees, line 20: extensible is a good improvement over the boolean 👍

toddlees (approved):

PR #1712, opened 2024-02-07
feature/1569 - System will determine if Committee is qualified on Form 3X

dheitzer:
https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/1569


front-end/src/app/reports/submission-workflow/submit-report-step2.component.ts
```diff
@@ -132,6 +135,10 @@ export class SubmitReportStep2Component extends DestroyerComponent implements On
       ...this.report,
       ...ValidateUtils.getFormValues(this.form, this.report.schema, this.formProperties),
     });
+    if (payload instanceof Form3X) {
+      payload.qualified_committee = !!this.committeeAccount?.committee_type &&
+        F3xQualifiedCommitteeTypeCodes.includes(this.committeeAccount.committee_type);
+    }
```
Elaine-Krauss-TCG, line 142: This does get the job done nicely for submission, but if you generate a .fec for print previews or through the generate .fec debug page, qualified committee won't be populated.

The business reason in the ticket implies that it should be populated whenever a .fec is generated, but the AC only specifically mentions submission.  I think this should be double checked.

If that is intended, though, then I fully approve!

dheitzer: @Elaine-Krauss-TCG great catch.  I've reach out to Gregg and Aurelia to confirm the A/C.

> This does get the job done nicely for submission, but if you generate a .fec for print previews or through the generate .fec debug page, qualified committee won't be populated.
> 
> The business reason in the ticket implies that it should be populated whenever a .fec is generated, but the AC only specifically mentions submission. I think this should be double checked.
> 
> If that is intended, though, then I fully approve!

dheitzer: Aurelia updated the ACs to add .fec file generation and print preview.  Pushed updated code covering these scenarios.

mjtravers (approved):

Issue #120, opened 2022-02-14
System to require users to consent to security notice annually


AureliaKhorsand:
Business Reason

Per FEC Security team guidance, users will be required to consent to a security notice the first time they log in to FECFile Online, then annually after that. We should require all users to acknowledge the system use notification the first time they access the application. We should track this acknowledgement. Subsequent displays of the notification should happen at least yearly. We only need to keep a record of the most recent acknowledgement.

Acceptance Criteria

***When*** a newly provisioned user logs in to FECFile Online (first time login)

  ***And*** they have entered their user profile information (#92) and clicked "Continue"

  ***Then*** the system will display the following message:

  "insert PLACEHOLDER text here. We will updated with approved language closer to go-live"

***And*** they can either:

*   "Decline and exit" and be sent back to the Landing Page (PER ticket #1719), and the security notice will appear on the next login attempt or;
*   "Consent" _without_ the checkbox selected and continue to the dashboard (or Committee Account selection with #1654), whereupon they will see the security message on next login as well
*   "Consent" _with_ the checkbox selected and continue to the dashboard (or Committee Account selection with #1654), then they will not see the message again for 365 days.

—

***When*** an authorized user logs in to FECFile Online (they have logged in before)

  ***Then*** the system will display the following message:

  "insert PLACEHOLDER text here. We will updated with approved language closer to go-live"

***And*** they can either:

*   "Decline and exit" and be sent back to the Landing Page, and the security notice will appear on the next login attempt or;
*   "Consent" _without_ the checkbox selected and continue to the dashboard (or Committee Account selection with #1654), whereupon they will see the security message on next login as well
*   "Consent" _with_ the checkbox selected and continue to the dashboard (or Committee Account selection with #1654), then they will not see the message again for 365 days.

Dev Notes

1.  Here is the text you should use:

       #

       ***You are about to access the Federal Election Commission computer system.***

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.

***By using this system, you consent to the terms set forth in this notice.***

[...]

PR #1726, opened 2024-02-12
Patch/120 - Security Notice Fixes & Routing Refactor

Elaine-Krauss-TCG:
#120 

The Security Notice page:
 - Clicking the "consent" button on the Security Notice page without checking the box allows the user to continue, but the user will see the security notice again on their next session.

Routing:
 - The login module has its own routing
 - Sidebar state is managed with route data instead of resolvers
 
 Store/Login:
 - The userLoggedOut$ effect handles clearing user, commitee, and session data instead of the Login page


front-end/src/app/layout/layout.component.ts
```diff
@@ -22,25 +20,11 @@ export enum BackgroundStyles {
...
-    this.store
-      .select(selectSidebarState)
-      .pipe(takeUntil(this.destroy$))
-      .subscribe((state) => {
-        // Show sidebar if there is an associated "sidebar state" declared in a routing module.
-        this.showSidebar = !!state;
-      });
-
     this.onRouteChange();
     this.router.events.pipe(filter((event) => event instanceof NavigationEnd)).subscribe(() => {
```
toddlees, line 29: we probably want a takeUntil destroy on this events subscription

front-end/src/app/login/login/login.component.ts
```diff
@@ -15,16 +11,9 @@ export class LoginComponent implements OnInit {
...
-  constructor(
-    private loginService: LoginService,
-    private store: Store,
-  ) {}
+  constructor(private loginService: LoginService) {}
 
   ngOnInit() {
-    localStorage.clear();
-    this.loginService.clearUserLoggedInCookies();
-    this.store.dispatch(setCommitteeAccountDetailsAction({ payload: new CommitteeAccount() }));
```
toddlees, line 26: It's my understanding that the reason these are here is because currently we can't guarantee that the logout code is called before we get here.  So we want to clear out potentially stale data

front-end/src/app/login/login-routing.module.ts
```diff
@@ -0,0 +1,65 @@
...
+import { SecurityNoticeComponent } from './security-notice/security-notice.component';
+import { BackgroundStyles } from 'app/layout/layout.component';
+import { LoginGuard } from 'app/shared/guards/login-page.guard';
+import { nameGuard } from 'app/shared/guards/name.guard';
+import { securityNoticeGuard } from 'app/shared/guards/security-notice.guard';
+import { SelectCommitteeComponent } from 'app/committee/select-committee/select-committee.component';
+import { HeaderStyles } from 'app/layout/header/header.component';
+import { UpdateCurrentUserComponent } from 'app/users/update-current-user/update-current-user.component';
+
+const routes: Routes = [
```
toddlees, line 13: this is a nice way to bundle up the entry pages

front-end/src/app/layout/sidebar/menus/abstract-menu.component.ts
```diff
@@ -1,41 +1,50 @@
...
-        const isEditable = this.reportService.isEditable(activeReport);
-
-        return of(this.getMenuItems(sidebarState, activeReport, isEditable));
+      map((activeReport: Report | undefined) => {
+        this.activeReport = activeReport;
+        return this.updateMenuItems(activeReport);
       }),
     );
+
+    this.router.events.pipe(filter((event) => event instanceof NavigationEnd)).subscribe(() => {
```
toddlees, line 44: you're going to want to combineLatest the router even and the active report observables.  when we take data out of an observable and put it into a synchronous variable like `this.activeReport` or `this.sidebarState` we forfeit control of flow.  so later in `this.getMenuItems(this.sidebarState as SidebarState, activeReport, isEditable);` we don't know whether sidebarState is ready or not.

I'd recommend using a combineLatest like this:
```
const routerEvents$ = this.router.events.pipe(filter((event) => event instaceof NavigationEnd));
this.items$ = combineLatest([this.activeReport$, routerEvents$]).pipe.map(([activeReport, sidebarState]) => {
    const isEditable = this.reportService.isEditable(activeReport);
    return this.getMenuItems(this.sidebarState as SidebarState, activeReport, isEditable);
});
```
then `activeReport` and `sidebarState` are no longer needed and don't pose any risk of being used when their data is not up to date

mjtravers (approved):

front-end/src/app/store/sidebar-state.actions.ts
toddlees, line 1: I love that we can take the sidebar state out of the store since it can be accurately derived from the route

toddlees (changes requested):

PR #1734, opened 2024-02-14
1636 - Update e2e tests to better handle dropdowns

sasha-dresden:
I was able to handle most of these issues via minor configuration changes, but the "Add a Memo" dropdown was a major blocker and was proving difficult to handle in a way that worked without breaking the appearance of dropdown. I ended up replacing the primeng dropdown with a bootstrap dropdown. This also gave some better keyboard control. But this section definitely needs a focus of the code review and testing.

As a note, this also includes a minor change update to primeng which led to some odd graphical changes that I was able to address with some css fixes. For example the Import button on the Contacts page was broken out to cssless file input which looks quite bad.

### DIFF
--- a/front-end/cypress/e2e/pages/loginPage.ts
+++ b/front-end/cypress/e2e/pages/loginPage.ts
@@ -1,3 +1,6 @@
+import { ContactListPage } from './contactListPage';
+import { ReportListPage } from './reportListPage';
+
 export class LoginPage {
   static login() {
     const sessionDuration = 10; //Login session duration in minutes
@@ -83,4 +86,10 @@ function retrieveAuthToken() {
   const storedData = localStorage.getItem('fecfile_online_userLoginData');
   const loginData = JSON.parse(storedData ?? '');
   return 'JWT ' + loginData.token;
+}
+
+export function Initialize() {
+  LoginPage.login();
+  ReportListPage.deleteAllReports();
+  ContactListPage.deleteAllContacts();
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,7 +305,7 @@ workflows:
           requires:
             - test
             - dependency-check
-            # - e2e-test -- temporarily turn off until dropdown issue fixed
+            - e2e-test
           filters:
             branches:
               only: /develop|release\/sprint-\d+|main/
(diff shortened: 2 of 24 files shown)


dheitzer (dismissed):
It looks like there are a lot of html indentation changes (from 2 to 4 spaces, [like this one for example](https://github.com/fecgov/fecfile-web-app/pull/1734/files#diff-bbf3bee0894592b4c3c8a8c4c7d2eda9c4909e19333cf2080db7eea42503cc5b)).  Would it be possible to change these back to 2? (that might also help us to see the non-formatting changes easier).

sasha-dresden:
Good point @dheitzer. I'm not sure how it was getting set to 4. I checked the prettier file and my ide settings, both were set to 2 spaces. I went ahead and ran a code cleanup to catch anything that needed cleaning.

dheitzer:
> Good point @dheitzer. I'm not sure how it was getting set to 4. I checked the prettier file and my ide settings, both were set to 2 spaces. I went ahead and ran a code cleanup to catch anything that needed cleaning.

Thanks @sasha-dresden helps a lot!

front-end/cypress/e2e/pages/loginPage.ts
```diff
@@ -72,19 +76,28 @@ function legacyLogin() {
...
+  cy.get(alias).find('.committee-list').click();
 }
 
 function retrieveAuthToken() {
   const storedData = localStorage.getItem('fecfile_online_userLoginData');
   const loginData = JSON.parse(storedData ?? '');
   return 'JWT ' + loginData.token;
 }
+
+export function Initialize() {
```
dheitzer, line 91: I really like that we're getting rid these code duplications

sasha-dresden:
Ok I rebased this project and started semi-fresh. This led to far fewer file changes and since 1632 had already been done for the package updates, let to this being focused on _just_ the e2e test fixes. Previously I'd had it working for the transactions e2e tests but for this round I went through every existing e2e test and made sure they passed. Every single test should be passing now.

However, every now and then, and I don't know why exactly, the committee search box times out. I added a wait command to wait for the committee lookup GET request to finish before proceeding, which resolved most of these but randomly, and not repeatedly, it will still timeout. Not 100% certain why that's happening sometimes. Seems to be due to a delay with receiving data from the api. It's fairly rare and only happened once when I ran the entire suite and didn't happen when I re-ran that test.

mjtravers (approved):

PR #1742, opened 2024-02-20
Feature/91 - Add delete contact service call

sasha-dresden:
There is an api feature that goes along with this that creates the new endpoint.


sasha-dresden:
Updated to match with reorganizational changes from adding a user to a committee.

mjtravers (approved):

PR #1753, opened 2024-02-27
Feature/1722 - Add a preparePayload function to these three services to strip out unnecessary data from being sent to the api.

sasha-dresden:


front-end/src/app/shared/interfaces/table-list-service.interface.ts
```diff
@@ -5,10 +5,17 @@ export interface TableListService<T> {
...
 
   delete(item: T): Observable<null>;
 
   update?(item: T, fieldsToValidate?: string[]): Observable<T>;
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function clearEmptyKeys(obj: Record<string, any>) {
+  for (const key of Object.keys(obj)) {
+    if (!obj[key] || (typeof obj[key] === 'object' && Object.keys(obj[key]).length === 0)) delete obj[key];
```
toddlees: Falsey values and nulls are data too.  If the intent is to set a boolean to False or to remove data from a previously populated field this method will undo that change.

toddlees (changes requested):

front-end/src/app/shared/interfaces/table-list-service.interface.ts
sasha-dresden: Fixed that. Sends over empty data now. Also fixed a few misspellings I noticed.

toddlees (approved):

PR #1756, opened 2024-02-28
1754 - Fix spelling of FECFILE

sasha-dresden:


mjtravers (changes requested):
I noticed  that there are lines altered by changing around the order of HTML tags. This get problematic when it is widespread and affects a large number of files and lines. Is there a way to turn this off in your IDE?

mjtravers (approved):

Issue #1583, opened 2023-12-12
User to confirm Form 3X that will include Independent Expenditure (Schedule E) transactions created from a Form 24


GreggMoreland:
Business Reason

As a user, I will be able to designate and select which Form 3X Report to link each Independent Expenditure transaction that is created on a Form 24 Report so that I can report on all my transactions

Acceptance Criteria

***Given*** a Form 24 report (in-progress)

 ***When*** the user CREATES one of the following transactions:

*   Independent Expenditure
*   Void of Independent Expenditure
*   Multistate Independent Expenditure
*   Credit Card Payment for Independent Expenditure
*   Staff Reimbursement for Independent Expenditure
*   Payment to Payroll for Independent Expenditure

***Then*** the user will see a message at the top of the transaction indicating that there must be an in-progress Form 3X with coverage dates that include the IE they will enter (see text below)

 ***And*** if they need to, they can click "Create Form 3X" to initiate a new Form 3X report (coverage date selection screen), or "Return to manage reports" to return to the Manage reports page.

***Given*** the user enters the dissemination and/or disbursement dates

 ***then*** the system will use the following logic to determine the linked in-progress Form 3X report:

*   if only one date is present, that date must be within the selected Form 3X coverage period. If both dates are present, the Date of Disbursement must be within the report coverage period.

***And*** there will be a tooltip with additional information (see text below) and an in-line validation error if there is no applicable Form 3X report.

Wireframe

 

1.  Example

      

1.  Errors

      

Video

 [Private Zenhub Video]([https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1NTQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ed5c413aea08a983be51db9cad6187a41939b516/Form%2024%20-%20video.mp4](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1NTQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ed5c413aea08a983be51db9cad6187a41939b516/Form%2024%20-%20video.mp4))

### QA Notes ###
- [x] BA/Design Quality Checklist complete
[...]

PR #1761, opened 2024-02-29
Feature/1583 - Linked Report Field

Elaine-Krauss-TCG:
#1583 

[App PR](https://github.com/fecgov/fecfile-web-app/pull/1761)
[API PR](https://github.com/fecgov/fecfile-web-api/pull/761)


front-end/src/app/shared/components/inputs/amount-input/amount-input.component.html
```diff
@@ -30,6 +30,9 @@
           ></app-error-messages>
         </div>
       </div>
+      <div class="col-4">
```
toddlees, line 33: What do you think about moving the `*ngIf=` that's in the linked-report-input out to this div so that if it's not there it doesn't take up space on the row?

toddlees (approved):

PR #1763, opened 2024-03-01
66 - Add transaction table to contact dialog only when opened from Contact Manager

sasha-dresden:
There is an API PR that goes along with this.


front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
```diff
@@ -260,4 +291,29 @@ export class ContactDialogComponent extends DestroyerComponent implements OnInit
     }
     this.resetForm();
   }
+
+  async openTransaction(transaction: Transaction) {
+    await this.router.navigate([`reports/transactions/report/${transaction.report_id}/list/${transaction.id}`]);
+  }
+
+  getTransactionDate(transaction: Transaction): string {
```
toddlees: this can be avoided if we use the transaction list endpoint.  it provides `amount` and `date` fields for generic tables letting us use and sort on different fields in one column.

toddlees (changes requested):

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
sasha-dresden: Ah for some reason I thought I had to go through the templateMap, I was using a new endpoint but it went through the same TransactionSerializer as the transaction list endpoint. So it did have amount and date available. Got that updated and much simplified.

toddlees (approved):

PR #1764, opened 2024-03-06
Feature/1662

dheitzer:
https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/1662


Elaine-Krauss-TCG (approved):

sasha-dresden (commented):
I don't think this is related to these changes, but when I tried to search for a committee, and then select it from the dropdown in filled in the search box with [object Object]

Screenshot from 2024-03-07 14-28-59
Screenshot from 2024-03-07 14-29-09

toddlees:
> I don't think this is related to these changes, but when I tried to search for a committee, and then select it from the dropdown in filled in the search box with [object Object]
> 
> 

Fixed that issue. I think it might have popped up in the latest primeng upgrade.  also addressed the split boxes when the committee can't be registered:
image
image
Spoke with mitchell and the intent was for "Committee account" to be "Results" like it is when you search

toddlees (approved):

toddlees:
adds margin between boxes
image

sasha-dresden (approved):
That makes sense. Looking at the code, it said that `field` attribute was deprecated, so maybe that was what changed. Clearing the field definitely works in this situation since we're loading the Committee in the results below.

PR #1809, opened 2024-03-22
feature/1786 - Add emptymessage template for tables

sasha-dresden:


dheitzer (dismissed):
This looks good @sasha-dresden but I reached out to BAs to see if we needed to add this empty message to the deleted contacts table as well since I saw it wasn't mentioned in this ticket.  They confirmed that we should, so this table should also receive this empty message as well.

dheitzer:
> This looks good @sasha-dresden but I reached out to BAs to see if we needed to add this empty message to the deleted contacts table as well since I saw it wasn't mentioned in this ticket. They confirmed that we should, so this table should also receive this empty message as well.

@sasha-dresden please disregard this.  BAs determined that this table does not need the message since "Restore deleted contact(s)" button should not be active if there are no contacts to restore.

sasha-dresden:
> > This looks good @sasha-dresden but I reached out to BAs to see if we needed to add this empty message to the deleted contacts table as well since I saw it wasn't mentioned in this ticket. They confirmed that we should, so this table should also receive this empty message as well.
> 
> @sasha-dresden please disregard this. BAs determined that this table does not need the message since "Restore deleted contact(s)" button should not be active if there are no contacts to restore.

Oh whoops, I just pushed a fix right before you added the update. Restoring to the original now.

dheitzer:
> > > This looks good @sasha-dresden but I reached out to BAs to see if we needed to add this empty message to the deleted contacts table as well since I saw it wasn't mentioned in this ticket. They confirmed that we should, so this table should also receive this empty message as well.
> > 
> > 
> > @sasha-dresden please disregard this. BAs determined that this table does not need the message since "Restore deleted contact(s)" button should not be active if there are no contacts to restore.
> 
> Oh whoops, I just pushed a fix right before you added the update. Restoring to the original now.

thanks @sasha-dresden! sorry for the confusion

mjtravers (approved):

PR #1813, opened 2024-03-22
feature/1766 - Add Show/Hide password to Submit Report Step

sasha-dresden:
There is a sonarcloud complaint about styling the eyeicon and eyeslashicon. This should be ok. All I did was move this from the debug-login.component.scss into the theme.css so that it would be available to both the debug-login page and the Submit Report page, as well as any future pages that make use of the show/hide password feature.

### DIFF
--- a/front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
@@ -101,6 +101,7 @@
               formControlName="filingPassword"
               [feedback]="false"
               [toggleMask]="true"
+              autocomplete="new-password"
             />
             <app-error-messages
               [form]="form"
--- a/front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
@@ -19,7 +19,7 @@
         <div class="col-4">
           <div class="field">
             <label for="treasurer_last_name"><strong>LAST NAME</strong></label>
-            <input type="text" pInputText id="treasurer_last_name" formControlName="treasurer_last_name" />
+            <input type="text" pInputText id="treasurer_last_name" formControlName="treasurer_last_name"/>
             <app-error-messages
               [form]="form"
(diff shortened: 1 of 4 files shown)


front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
```diff
@@ -95,7 +95,13 @@ <h3>Enter password</h3>
             <label for="filingPassword">
               <strong>ELECTRONIC FILING PASSWORD </strong>
             </label>
-            <input type="text" pInputText id="filingPassword" formControlName="filingPassword" />
+            <p-password
+              id="filingPassword"
+              type="password"
```
toddlees, line 100: So i tried this a while back and ran into a problem.  Because it's a password input, the browser throws in your password for the website which probably isn't the same as the filing password.  Do you think there's a way we make it a password input but disable the autofill?

toddlees (changes requested):

front-end/src/app/reports/submission-workflow/submit-report-step2.component.html
sasha-dresden: So I've been looking into this and there is an autocomplete field which you should technically be able to set to "off" to prevent this. However, modern browsers seem to have chosen to ignore it. 
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields

sasha-dresden: So after a lot of digging around I eventually found:
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields

Which was a little disheartening since it seemed modern browser just wouldn't allow you to disable until you get to the very bottom and it tells you you can use:
autocomplete="new-password"

That finally did the trick for me.

toddlees (approved):

mjtravers (approved):

PR #1823, opened 2024-03-29
Feature/386 - Add ability to delete reports with restrictions on when a report can be deleted

sasha-dresden:


front-end/src/app/reports/report-list/report-list.component.ts
```diff
@@ -45,6 +49,11 @@ export class ReportListComponent extends TableListBaseComponent<Report> implemen
     this.loadItemService(this.itemService);
   }
 
+  override async loadTableItems(event: TableLazyLoadEvent) {
+    await super.loadTableItems(event);
+    this.items.forEach((item) => this.canDelete(item));
```
toddlees: It seems like the same reports that we render with the serializer on the backend are the ones we fire off this request for.  What do you think of calculating canDelete in the serializer (maybe even only on the list endpoint) so we don't have to go back and forth with the api?

toddlees (changes requested):

mjtravers:
PR is OBE. Replaced by https://github.com/fecgov/fecfile-web-app/pull/1831
Closing PR without merging.

PR #1848, opened 2024-04-12
Feature/404 - Filter Transaction Types on the Transaction Select screen when committee is PAC

sasha-dresden:

PR #1860, opened 2024-04-18
feature/1849 - Add committee info to Form 3X creation

sasha-dresden:
Issue #1848

Related to API [PR835](https://github.com/fecgov/fecfile-web-api/pull/835)


front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -173,14 +176,22 @@ export class CreateF3XStep1Component extends DestroyerComponent implements OnIni
...
+  public async save(jump: 'continue' | undefined = undefined) {
     this.formSubmitted = true;
     if (this.form.invalid) {
       this.store.dispatch(singleClickEnableAction());
       return;
     }
 
     const summary: Form3X = Form3X.fromJSON(SchemaUtils.getFormValues(this.form, f3xSchema, this.formProperties));
+    if (this.committeeAccount) {
+      summary.committee_name = this.committeeAccount.name;
```
toddlees: I like the idea of assigning the committee_name here.  The requirements are to assign it on the session where the report is submitted for submission/webprint/test-.fec so that it reflects any changes made up to that point.  We handled this for `qualified_committee` by running an update before we kick off any .fec tasks:

```
async generate() {
    if (this.report instanceof Form3X) {
      const payload: Form3X = Form3X.fromJSON({
        ...this.report,
        qualified_committee: this.form3XService.isQualifiedCommittee(this.committeeAccount)
      });
      await firstValueFrom(this.form3XService.update(payload, ['qualified_committee']));
    }
    this.apiService.post(`/web-services/dot-fec/`, { report_id: this.report?.id }).subscribe(() => undefined);
    this.fileIsGenerated = true;
  }
```

You could tie in committee_name there too (especially since `this.committee_account` will be available in the scope.

toddlees (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
sasha-dresden: Ok, I switched it up so the committee information is only being set when the qualified_committee was also being set.

toddlees (changes requested):
working for me

toddlees (approved):

Issue #1889, opened 2024-04-27
Fix synchronous sort link between Form Type and Version columns on Reports table.


mjtravers:
The sorting arrows on the "Form type" and "Version" columns in the table both change when you click either one. They should operate separately. Unlink the sorting for columns "Form type" and "Version" on the reports list table:

![Image](https://github.com/fecgov/fecfile-web-app/assets/2971785/408ec4d8-1d53-46e4-8658-e64a6ada366d](https://github.com/fecgov/fecfile-web-app/assets/2971785/408ec4d8-1d53-46e4-8658-e64a6ada366))

### QA Notes ###
 
 null

### DEV Notes ###
 
 null

### Design ###
 
 null

 See full ticket and images here: [FECFILE-413](https://fecgov.atlassian.net/browse/FECFILE-413)

WiseQA:
QA and DEV reviewed this bug that was found after Sprint 41.

mjtravers:
Passes CR. Sending to QA.

WiseQA:
QA review verified the sorting arrows on the "Form type" and "Version" columns in the table both operate (sort) separately. Unlink the sorting for columns "Form type" and "Version" on the reports list table:

![Image](https://github.com/fecgov/fecfile-web-app/assets/96139912/83ffce96-c8b3-4a1d-95a9-2221a238b0d7)
___________________________________________________________________________________________________

Screenshot of selected sort for "Form type" column only

![Image](https://github.com/fecgov/fecfile-web-app/assets/96139912/77585158-4d8b-4e91-8556-8efc28aac2cd)

__________________________________________________________________________________________________

Screenshot of selected sort for "Version" column only

![Image](https://github.com/fecgov/fecfile-web-app/assets/96139912/5b334f96-08e9-4622-942f-5736f9c9e4e9)

__________________________________________________________________________________________________


QA retested sorting working as designed for each column in "Manage reports" table.

QA Review Completed.   Moved to Stage Ready.

PR #1910, opened 2024-05-01
feature/1889 - Move version_label from front end to back to allow for sorting

sasha-dresden:
Issue #1889 
API PR [PR868](https://github.com/fecgov/fecfile-web-api/pull/868)


mjtravers (approved):
That's great all that code can be removed now that the sort is on the back-end.

PR #865, opened 2023-02-10
Feature/349 create report popup

toddlees:

PR #1912, opened 2024-05-01
Feature/865 - Update to have transaction create/update only return string

sasha-dresden:
Issue: [#865](https://github.com/fecgov/fecfile-web-api/issues/865)
API PR [PR871](https://github.com/fecgov/fecfile-web-api/pull/871)

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/double-transaction-type-base.component.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/double-transaction-type-base.component.ts
@@ -112,7 +112,7 @@ export abstract class DoubleTransactionTypeBaseComponent
     })[0];
   }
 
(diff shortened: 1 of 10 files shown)


front-end/src/app/shared/components/transaction-type-base/double-transaction-type-base.component.ts
```diff
@@ -139,7 +139,7 @@ export abstract class DoubleTransactionTypeBaseComponent
     ];
     payload.children[0].reports = payload.reports;
 
-    this.processPayload(payload, navigationEvent);
+    return this.processPayload(payload, navigationEvent);
```
toddlees, line 142: love this addition of flow control!

Issue #1906, opened 2024-05-01
UI: Tables: Move or Add “Results per table:” to all tables


MitchellTCG:
Business Reason

  As the principal UX, I would like all tables to have continuity so users will have a similar experience throughout the site.

Tasks

  using the PrimeNG Table/Programmatic as a baseline

Please place the Results per table left justified under the data table (with 10px padding left) and the pagination justified to the right (with 10px padding to the right).

Move

  –

  Move the “Results per table” on the same line as the pagination on the following pages:

*   [x] Transactions in this report table
*   [x] Guarantors table

Add

  –

  Add“Results per table” to the table and place it on the same line as the pagination on the following pages:

*   [x] Manage reports table
*   [x] Manage Users table
*   [x] Manage contacts table
*   [x] Restore deleted contact(s) table
*   [x] Transaction history table inside the Edit contact modal

Icons

  –

  Update Icons

*   [x] Down Arrow
*   [x] Beginning
*   [x] Back One Page
*   [x] Forward One Page
*   [x] End

Wireframes

1.  Prime NG Reference

       ![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/66500f11-002c-41c8-be40-a3639cf26b2a](https://github.com/fecgov/fecfile-web-app/assets/25588182/66500f11-002c-41c8-be40-a3639cf26b2))

1.  Current state for Transactions in this report and Guarantors

       ![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/59135b58-0a6e-4741-830d-ef0c3d851ec7](https://github.com/fecgov/fecfile-web-app/assets/25588182/59135b58-0a6e-4741-830d-ef0c3d851ec))

1.  Current state for Manage reports, Manage users, Mange Contacts, Restore delete contact(s), and Transactions history

       ![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/24673b86-d268-4a0c-af2a-06bdaed40943](https://github.com/fecgov/fecfile-web-app/assets/25588182/24673b86-d268-4a0c-af2a-06bdaed4094))

1.  Ideal State

       ![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/512361ab-b2ad-4269-a68d-5e3ce297baa5](https://github.com/fecgov/fecfile-web-app/assets/25588182/512361ab-b2ad-4269-a68d-5e3ce297baa))

Assets

  ***Down Arrow***

[...]

PR #1931, opened 2024-05-14
Feature/1906 - Rework tables so all have same styling and move results to bottom

sasha-dresden:
Issue: #1906 
API: [PR889](https://github.com/fecgov/fecfile-web-api/pull/889)


front-end/src/app/shared/components/table/table.component.html
```diff
@@ -0,0 +1,131 @@
...
+        <td *ngIf="hasCheckbox">
+          <p-tableCheckbox [value]="item" [ariaLabel]="checkboxLabel!(item)"></p-tableCheckbox>
+        </td>
+        <ng-container *ngTemplateOutlet="body; context: { $implicit: item }"></ng-container>
+      </tr>
+    </ng-template>
+    <ng-template pTemplate="summary">
+      <div class="d-flex justify-content-between">
+        <div class="responsive-container">
+          <span class="paginator-text">Results per table:</span>
```
mjtravers: Should this be "Results per page:"?

front-end/src/app/shared/components/table/table.component.html
```diff
@@ -0,0 +1,131 @@
...
+          </p-paginator>
+        </div>
+      </div>
+    </ng-template>
+    <ng-template pTemplate="sorticon" let-sortOrder>
+      <app-table-sort-icon [sortOrder]="sortOrder"></app-table-sort-icon>
+    </ng-template>
+    <ng-template pTemplate="paginatorleft">
+      <div class="responsive-container">
+        <span>Results per table:</span>
```
mjtravers, line 100: Should this be "Results per page:"?

mjtravers (changes requested):
A small label change requested. Also, what is the default number of rows to show on a table, 5 or 10?

front-end/src/app/shared/components/table/table.component.html
sasha-dresden: Ah I didn't catch that. The old way was "table" and there wasn't a specific point to change it to "page" however all the wireframes had "page". Changed it to page.

sasha-dresden: As for the default number of rows, there isn't one. Some pages had their default as 10, others as 5.

mjtravers (approved):

Issue #1928, opened 2024-05-13
Fix nightly e2e tests


mjtravers:
The e2e tests have been failing the nightly run since May 6.

The issue appears to be Cypress not being able to find the email input on the legacy login page. The issue is not occurring locally, only on CircleCI.

Error Message:

  ![image]([https://github.com/fecgov/fecfile-web-app/assets/2971785/eda1056c-9074-469b-b36a-e7cb6be76fd0](https://github.com/fecgov/fecfile-web-app/assets/2971785/eda1056c-9074-469b-b36a-e7cb6be76fd0))

### QA Notes ###
 
 null

### DEV Notes ###
 
 null

### Design ###
 
 null

[FECFILE-394](https://fecgov.atlassian.net/browse/FECFILE-394)

lbeaufort:
On May 2, the e2e tests passed for the `release/sprint-41` branch. When I re-ran them today (5/15), they failed for the same commit: https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app?branch=release%2Fsprint-41. One difference between the failing and passing tests:

On the `Setup a remote Docker engine` step:
**Passing**
```
Created container accessible with:
  DOCKER_HOST=unix:///var/run/docker.sock
  DOCKER_MACHINE_NAME=localhost
Server Engine Details:
  Version:          24.0.9
  API version:      1.43 (minimum version 1.12)
```
**Failing**
```
Created container accessible with:
  DOCKER_HOST=unix:///var/run/docker.sock
  DOCKER_MACHINE_NAME=localhost
Server Engine Details:
  Version:          26.0.2
  API version:      1.45 (minimum version 1.24)
```

Pinning the version to 24 in the Circle config resulted in [passing e2e tests](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/5672/workflows/e53db6ad-04c0-42c9-910d-e03c5d2330c8/jobs/19601). The best guess I have so far is that the Docker API had a breaking change: https://docs.docker.com/engine/api/version-history/#v145-api-changes. Hotfix PRs in progress,

Additional follow-up ticket needed to get off the older docker version, ticket forthcoming.

mjtravers:
@lbeaufort The CircleCI config.yml changes look good but I noticed while code reviewing that there is a Snyk issue being raised on the PR branches. I do not have access to see what the issue is.

sasha-dresden:
[...]

PR #1942, opened 2024-05-16
feature/1928 Fixing e2e test issues

sasha-dresden:
Issue: #1928 
Laura's fixes resolved the problem with circle not being able to run e2e tests at all, but after they were running I noticed a couple that were failing, so I made some updates to resolve those.


Elaine-Krauss-TCG (approved):
image

Issue #1907, opened 2024-05-01
UI: Landing Page Redux: Update landing page for better responsive behavior


MitchellTCG:
Business Reason

  As the principal UX, I would like the landing page to be cleaned up and responsive for a better overall user experience

Tasks

  Official website banner

  –

  "An official website of the..." will stay the same

Navigation bar

  –

  the background-color will be change from grey to ***#FFFFFF***

*   [x] have 64 px of space between to bottom of the nav box and the Headline text

Headline text

  –

  ***At 1400 and 1200***

  Headline text stays the same _BUT will be on one line._ 

  ***Made for filers. Built to work anywhere.***

*   [x] Change font-size from 32px to ***36px***
*   [x] Center text

***At 992, 768, and 576***

  Headline text stays the same _AND will be on TWO line._ 

  **Made for filers. 

  Built to work anywhere.**

*   [x] Change font-size from 32px to ***36px***
*   [x] Center text

Paragraph text

–

  ***At 1400, 1200, and 992***

  Paragraph text will stay the same but each sentence will be on it's own line

  **FECFile is free web-based software for electronic filing. 

  It allows users to create, review, and submit reports from anywhere.**

*   [x] Change font from Karla to ***Gandhi Serif***
*   [x] Change font-size from 14px to 18px

***At 768 and 576***

  Paragraph text will stay the same and wrap inside a 480px div

  ***FECFile is free web-based software for electronic filing. It allows users to create, review, and submit reports from anywhere.***

*   [x] Change font from Karla to ***Gandhi Serif***
*   [x] Change font-size from 14px to 18px

Buttons

–

*   [x] Have 32px of vertical space between paragraph text and buttons
*   [x] change the font-weight of both buttons to ***font-weight:600***
*   [x] Sign up with Login.gov background color: #35BDBB and text is #212121

Learn More

–

*   [x] center text

Hero image

–

1.  at 1400xxl and 1200xl

*   [x] Have 64 px of vertical space between learn more url and hero image
*   [x] Image size is 800px wide
*   [x] 100px down from the top of the hero image change the background-color to ***#0BACFA***.
*   [x] There should be 64px of space between the bottom of the hero image and the footer

1.  at 992lg

*   [x] Have 64 px of vertical space between learn more url and hero image
*   [x] Image size is 600px wide
*   [x] 100px down from the top of the hero image change the background-color to ***#0BACFA***.
*   [x] There should be 64px of space between the bottom of the hero image and the footer

[...]

PR #1984, opened 2024-05-28
Feature/1907 - Login page redux

sasha-dresden:
#1907


Elaine-Krauss-TCG (approved):

front-end/src/app/layout/layout.component.ts
```diff
@@ -18,14 +18,29 @@ export enum BackgroundStyles {
...
 
   constructor(private router: Router) {
     super();
   }
+  ngAfterViewChecked(): void {
+    this.updateContentOffset();
+  }
+
+  updateContentOffset() {
+    const contentOffset = document.getElementById('content-offset');
```
mjtravers: Using document goes outside the Angular framework.
Would something like this work?

```
constructor(protected elementRef: ElementRef) ...

(<HTMLElement>this.elementRef.nativeElement).querySelector('#content-offset')
```

mjtravers (commented):
See comment about substituting out the DOM document call.

front-end/src/app/layout/layout.component.ts
sasha-dresden: Updated to use angular's renderer since I was already using the renderer in the Login Component. Updated it everywhere we were using document.getElementById which was just these two places and the DotFecService

mjtravers (approved):

PR #2042, opened 2024-07-10
hotfix/843 - Update 'Itemize' option to include when value was stored as null

sasha-dresden:
Issue: [FECFILE-1458](https://fecgov.atlassian.net/browse/FECFILE-1458)
The issue with why 'Itemize' wasn't showing up as an option was because the _itemize field was being stored in the sql view as NULL instead of False. And on the front-end we were specifically evaluating against false instead of falsy, which would include null or undefined.


mjtravers:
An alternative solution was found in the fecfile_web_api repo.
https://github.com/fecgov/fecfile-web-api/pull/957

Closing ticket without merging.

PR #2082, opened 2024-08-05
feature/429 - Remove alert for back end errors

sasha-dresden:
Issue [FECFILE-429](https://fecgov.atlassian.net/browse/FECFILE-429)

### DIFF
--- a/front-end/src/app/shared/interceptors/http-error.interceptor.spec.ts
+++ b/front-end/src/app/shared/interceptors/http-error.interceptor.spec.ts
@@ -49,7 +49,7 @@ describe('HttpErrorInterceptor', () => {
     const testIterceptor: HttpErrorInterceptor = TestBed.inject(HttpErrorInterceptor);
     const httpRequestSpy = jasmine.createSpyObj('HttpRequest', ['doesNotMatter'], { context: new HttpContext() });
     const httpHandlerSpy = jasmine.createSpyObj('HttpHandler', ['handle']);
-    spyOn(window, 'alert');
+    spyOn(console, 'log');
     httpHandlerSpy.handle.and.returnValue(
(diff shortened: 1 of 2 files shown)


dheitzer:
Do you know if the intention is to fail silently on server errors? It just seems like some indication to the user that something went wrong be good.  Maybe this is a ux discussion.

dheitzer (approved):

sasha-dresden:
> Do you know if the intention is to fail silently on server errors? It just seems like some indication to the user that something went wrong be good. Maybe this is a ux discussion.

Yea, I believe there was a little discussion about this previously in slack, but for this ticket it's just removing the alert. There should  eventually be a UX ticket to add an error page or some way to allow the user to know something went wrong. For example, to test this I tried going to a a transaction that didn't exist for the report. I should probably get a 404 page, but currently it just redirects you back to the landing page with any indication something went wrong in the console.

dheitzer:
> > Do you know if the intention is to fail silently on server errors? It just seems like some indication to the user that something went wrong be good. Maybe this is a ux discussion.
> 
> Yea, I believe there was a little discussion about this previously in slack, but for this ticket it's just removing the alert. There should eventually be a UX ticket to add an error page or some way to allow the user to know something went wrong. For example, to test this I tried going to a a transaction that didn't exist for the report. I should probably get a 404 page, but currently it just redirects you back to the landing page with any indication something went wrong in the console.

Thanks, makes sense.  Another scenario I tried was the contact lookup (I manually edited the code to raise an error).  In this case, the lookup just spins which is why I was wondering.

toddlees:
@dheitzer @sasha-dresden You're right.  There will be future tickets to handle errors in the UI

PR #2092, opened 2024-08-08
Feature/1518 - Replace all P-Dividers

Elaine-Krauss-TCG:
[Fecfile #1518](https://fecgov.atlassian.net/browse/FECFILE-1518?atlOrigin=eyJpIjoiY2JlMTM0YjI1MWM3NDkwOWJjNWJiNWFkZWNkYjMwZjYiLCJwIjoiaiJ9)

### DIFF
--- a/front-end/src/app/committee/committee-info/committee-info.component.html
+++ b/front-end/src/app/committee/committee-info/committee-info.component.html
@@ -1,5 +1,5 @@
 <h1>Account information</h1>
-<p-divider></p-divider>
+<hr type="solid" />
 <form id="form" p-disabled [formGroup]="form" class="p-fluid">
   <div class="grid">
     <div class="col-12">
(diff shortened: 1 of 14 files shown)


toddlees (changes requested):
I like it.  Who knows why the p-dividers went invisible.  Who cares.

Lets get rid of the css for them in theme.css

also looks like i missed a page or two in the list i gave shelly.  looks like the TransactionIndependentExpenditurePickerComponent has one that we'll want to convert as well

Elaine-Krauss-TCG:
> I like it. Who knows why the p-dividers went invisible. Who cares.
> 
> Lets get rid of the css for them in theme.css
> 
> also looks like i missed a page or two in the list i gave shelly. looks like the TransactionIndependentExpenditurePickerComponent has one that we'll want to convert as well

Ah, I searched for `<p-divider>` and I missed the one instance where a p-divider had a class.  Fixed

toddlees (approved):

PR #2097, opened 2024-08-09
feature/1483 - Limit the inclusion of children in transaction payloads to double and triple transaction sets

dheitzer:
https://fecgov.atlassian.net/browse/FECFILE-1483


front-end/src/app/shared/resolvers/transaction.resolver.ts
```diff
@@ -79,7 +79,11 @@ export class TransactionResolver {
   }
 
   resolveExistingTransaction(transaction: Transaction): Observable<Transaction | undefined> {
-    if (transaction.children) {
+    if (
+      transaction.children &&
```
sasha-dresden: Shouldn't this no longer include transaction.children in the if statement since it's no longer being passed from the backend. I just checked and it works ok because the frontend initializes it to an empty array, but it still doesn't feel like a needed check here. Also I just checked with a Conduit earmark receipt and ran into issues because the type was ScheduleATransactionTypes.CONDUIT_EARMARK_RECEIPT_DEPOSITED which was not included in the MultipleEntryTransactionTypes

dheitzer: @sasha-dresden good catches.  I have addressed these and have also added the PAC_CONDUIT_EARMARK memo types as well.

sasha-dresden: Awesome! Thanks. Oh, I just approved this but then I noticed the linter failed, might want to run the prettier fix command.

sasha-dresden (approved):

mjtravers (approved):

PR #2106, opened 2024-08-15
feature/1229 - Update filing frequency toggle to default based on the filing frequency of the committee

sasha-dresden:
Issue [FECFILE-1229](https://fecgov.atlassian.net/browse/FECFILE-1229)

### DIFF
--- a/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
+++ b/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
@@ -83,7 +83,7 @@ export class CreateF3XStep1Component extends DestroyerComponent implements OnIni
       .pipe(takeUntil(this.destroy$))
       .subscribe(([committeeAccount, existingCoverage]) => {
         this.committeeAccount = committeeAccount;
-        const filingFrequency = committeeAccount?.filing_frequency ?? 'Q';
+        const filingFrequency = committeeAccount?.filing_frequency === 'M' ? 'M' : 'Q';
         this.form.addControl('filing_frequency', new FormControl());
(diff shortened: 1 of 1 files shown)


front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -83,7 +83,7 @@ export class CreateF3XStep1Component extends DestroyerComponent implements OnIni
       .pipe(takeUntil(this.destroy$))
       .subscribe(([committeeAccount, existingCoverage]) => {
         this.committeeAccount = committeeAccount;
-        const filingFrequency = this.userCanSetFilingFrequency ? 'Q' : committeeAccount?.filing_frequency;
+        const filingFrequency = committeeAccount?.filing_frequency ?? 'Q';
```
toddlees: looks like the acs are a little more complicated:
`if Filing Frequency = Q, A or T, system will default to 'Quarterly Reporting'`

toddlees (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
sasha-dresden: Ah I see what you mean, we wouldn't want to set filingFrequency to 'T'. I updated this logic so now everything that isn't 'M' defaults to 'Q'. That way 'Q', 'A', and 'T' all go to 'Q', but also when it's undefined it goes to 'Q' as well.

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #2156, opened 2024-08-29
Feature/1549 - Update ESlint

Elaine-Krauss-TCG:
[Fecfile#1549](https://fecgov.atlassian.net/browse/FECFILE-1549)


front-end/src/app/committee/manage-committee/manage-committee.component.ts
```diff
@@ -64,6 +64,7 @@ export class ManageCommitteeComponent extends TableListBaseComponent<CommitteeMe
           life: 3000,
         });
       }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
     } catch (error) {
       return;
```
toddlees, line 68: looks like we can remove the identifier and the lint escape
```
} catch {
```
https://stackoverflow.com/a/56001361

toddlees (changes requested):

front-end/src/app/committee/manage-committee/manage-committee.component.ts
Elaine-Krauss-TCG: Great idea!  Done :+1:

toddlees (approved):

PR #2164, opened 2024-09-06
Patch - E2E Input Field Race Condition

Elaine-Krauss-TCG:


front-end/cypress/e2e/F24/reports-f24-independent-expenditures.cy.ts
```diff
@@ -48,13 +48,13 @@ describe('Form 24 Independent Expenditures', () => {
 
     PageUtils.clickButton('Save');
     PageUtils.clickLink('Independent Expenditure');
-    cy.contains(individualContactFormData.first_name).should('exist');
-    cy.contains(individualContactFormData.last_name).should('exist');
+    cy.get('#first_name').should('have.value', individualContactFormData.first_name);
```
toddlees, line 52: Looks like all of these are also clicking on an IE in the list and checking the transaction detail page.  Could we add the `cy.contains('Address').should('exist');` to all the places you replaced the contains so we know we're on the details page

toddlees (changes requested):

front-end/cypress/e2e/F24/reports-f24-independent-expenditures.cy.ts
Elaine-Krauss-TCG: Done

toddlees (approved):

PR #2190, opened 2024-09-16
feature/1611 - Switch select report dialog from 'contributor' to 'election'

sasha-dresden:
Issue [FECFILE-1611](https://fecgov.atlassian.net/browse/FECFILE-1611)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/select-report-dialog/select-report-dialog.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-list/select-report-dialog/select-report-dialog.component.ts
@@ -56,4 +56,8 @@ export class SelectReportDialogComponent implements OnInit {
     this.transaction = undefined;
     this.selectReportDialog?.nativeElement.close();
   }
+
+  get electContrib(): string {
+    return ReattRedesUtils.isReattribute(this.type) ? 'contributor' : 'election';
+  }
 }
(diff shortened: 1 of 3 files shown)


toddlees (changes requested):
This was missed in the writing of the ticket and I will address it.  For reattributions, we need to say contributor and for redesignations, we need to say election

sasha-dresden:
Updated code so it will use the appropriate text based off whether it's reattribution or redesignation.

front-end/src/app/reports/transactions/transaction-list/select-report-dialog/select-report-dialog.component.ts
```diff
@@ -56,4 +56,8 @@ export class SelectReportDialogComponent implements OnInit {
     this.transaction = undefined;
     this.selectReportDialog?.nativeElement.close();
   }
+
+  get electContrib(): string {
```
toddlees: It's a little tricky following the use and what's defined with these getters.  What do you think of these naming changes:
```
  // Label for action the user is trying to perform
  get actionLabel(): string {
    return ReattRedesUtils.isReattribute(this.type) ? 'reattribute' : 'redesignate';
  }

  // URL parameter to set the action type
  get urlParameter(): string {
    return ReattRedesUtils.isReattribute(this.type) ? 'reattribution' : 'redesignation';
  }

  // Label for the target of the action
  get actionTargetLabel(): string {
    return ReattRedesUtils.isReattribute(this.type) ? 'contributor' : 'election';
  }
```

toddlees (changes requested):

front-end/src/app/reports/transactions/transaction-list/select-report-dialog/select-report-dialog.component.ts
sasha-dresden: Good idea. Updated them with that. Also updated the spec to cover these a bit more fully.

toddlees (approved):

PR #2204, opened 2024-09-25
feature/1564 - Add poller for app updates

sasha-dresden:
Issue [FECFILE-1564](https://fecgov.atlassian.net/browse/FECFILE-1564)


front-end/src/app/shared/components/poller/poller.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    deploymentUrl += deploymentUrl.endsWith('/') ? 'index.html' : '/index.html';
+
+    this.pollerService.startPolling(deploymentUrl);
+    this.subscription = this.pollerService.isNewVersionAvailable$.subscribe((isNex) => {
+      this.isNewVersionAvailable = isNex;
+      if (this.isNewVersionAvailable) {
+        this.confirmationService.confirm({
+          key: 'refresh',
+          message:
+            'A new version of the application has been deployed! Would you like to refresh the page to receive the new app?',
```
toddlees: Good catch that this could be jarring for a user if they're in the middle of something.  This ticket is only to refresh when the versions don't match.  We're going to need to take this confirmation out or consult with design to get this part added to the ticket.

toddlees (changes requested):

front-end/src/app/shared/components/poller/poller.component.ts
sasha-dresden: Took out the confirmation. Now it instantly refreshes the page

Elaine-Krauss-TCG:
The Sonarcloud "10 New issues" thing gives me pause, but other than that this looks great

toddlees (approved):

PR #2208, opened 2024-09-27
feature/1505 update sec notice

dheitzer:
https://fecgov.atlassian.net/browse/FECFILE-1505

### DIFF
--- a/front-end/src/app/login/security-notice/security-notice.component.html
+++ b/front-end/src/app/login/security-notice/security-notice.component.html
@@ -16,45 +16,36 @@
       </ng-template>
 
       <div class="notice-content">
+        <strong>WARNING</strong>
         <p>
-          <strong>WARNING</strong>
-          <br />
           This version of the application is for testing new features and functionality and gathering user feedback. Any
           submitted reports do not constitute official filings with the Federal Election Commission (FEC). Any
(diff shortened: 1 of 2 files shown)


front-end/src/app/login/security-notice/security-notice.component.html
```diff
@@ -62,29 +63,27 @@ <h1>Security notification</h1>
...
-                [form]="form"
-                fieldName="security-consent-annual"
-                [formSubmitted]="formSubmitted"
-              ></app-error-messages>
-            </div>
           </div>
         </form>
         <div class="align-center">
-          <p-button class="security-notice-button" label="Decline and exit" (click)="loginService.logOut()"></p-button>
           <button
```
toddlees, line 63: These buttons need to be updated to match the wireframe.  hopefully we already have an example of that teal secondary and navy primary

toddlees (changes requested):
updated some css but the buttons need to be updated along with the check box

sasha-dresden (approved):

front-end/src/app/login/security-notice/security-notice.component.html
dheitzer: updated checkbox to match wireframe.  Also the wireframe was updated to match the requirements (and what was implemented).

PR #2266, opened 2024-10-17
feature/1600 - Add Security Notification page

sasha-dresden:
Issue [FECFILE-1600](https://fecgov.atlassian.net/browse/FECFILE-1600)


Elaine-Krauss-TCG (approved):

front-end/src/app/layout/layout.component.html
```diff
@@ -32,22 +32,16 @@
 
   <!-- Begin No Sidebar section -->
   <div *ngIf="!layoutControls.showSidebar" id="content-offset">
-    <div class="container">
```
toddlees, line 35: It looks like this change is having an effect on the header on wide views.  I don't think there is supposed to be a change to the header as a result of this ticket

As implemented:
image

On Dev:
image

toddlees (changes requested):

toddlees (approved):

PR #2274, opened 2024-10-21
feature/1654 - Setup Loan Repayment Received to override the contact_1 details from the parent transaction

sasha-dresden:
Issue [FECFILE-1654](https://fecgov.atlassian.net/browse/FECFILE-1654)

### DIFF
--- a/front-end/src/app/shared/models/transaction-types/LOAN_REPAYMENT_RECEIVED.model.ts
+++ b/front-end/src/app/shared/models/transaction-types/LOAN_REPAYMENT_RECEIVED.model.ts
@@ -5,8 +5,9 @@ import { SchATransaction, ScheduleATransactionTypeLabels, ScheduleATransactionTy
 import { STANDARD_CONTROLS, TransactionNavigationControls } from '../transaction-navigation-controls.model';
 import { AggregationGroups } from '../transaction.model';
 import {
+  ADDRESS_FIELDS,
+  COM_FIELDS,
   COMMITTEE,
-  COMMITTEE_FORM_FIELDS,
   ORGANIZATION_FORM_FIELDS,
 } from 'app/shared/utils/transaction-type-properties';
 import { TemplateMapKeyType } from '../transaction-type.model';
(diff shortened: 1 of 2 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/models/transaction-types/LOAN_REPAYMENT_RECEIVED.model.ts
```diff
@@ -4,14 +4,20 @@ import { SchATransactionType } from '../scha-transaction-type.model';
...
+} from 'app/shared/utils/transaction-type-properties';
+import { TemplateMapKeyType } from '../transaction-type.model';
 
 export class LOAN_REPAYMENT_RECEIVED extends SchATransactionType {
   formFields = ORGANIZATION_FORM_FIELDS;
   contactTypeOptions = COMMITTEE;
   title = LabelUtils.get(ScheduleATransactionTypeLabels, ScheduleATransactionTypes.LOAN_REPAYMENT_RECEIVED);
   schema = schema;
   override navigationControls: TransactionNavigationControls = STANDARD_CONTROLS;
+  override inheritedFields = [...COMMITTEE_FORM_FIELDS] as TemplateMapKeyType[];
```
toddlees: Ah this is a failure on my naming.  COMMITTEE_FORM_FIELDS has COMMON_FIELDS and AGGREGATE_FIELDS.  It will inherit the date and amount which we don't want.  I think you'll need to mix COM_FIELDS and ADDRESS_FIELDS.
`[...COM_FIELDS, ...ADDRESS_FIELDS]`

If you can think of a better name to clarify the typical committee based form (currently COMMITTEE_FORM_FIELDS) and JUST the committee specific fields (COM_FIELDS) we could make that improvement here

toddlees (changes requested):

front-end/src/app/shared/models/transaction-types/LOAN_REPAYMENT_RECEIVED.model.ts
sasha-dresden: Good catch. I didn't notice it was catching those extra fields. The names are a little confusing when you first come across them, but after working with them for a little while now, they make more sense, so I couldn't think of anything better for naming.

toddlees (approved):

PR #2282, opened 2024-10-22
feature/1118 - Allow user to override Form 3X cash-on-hand for the current year (6a)

dheitzer:
https://fecgov.atlassian.net/browse/FECFILE-1118
api pr: https://github.com/fecgov/fecfile-web-api/pull/1144


front-end/src/app/layout/header/header-links/header-links.component.html
sasha-dresden, line 1: There were changes in https://fecgov.atlassian.net/browse/FECFILE-1693 which this is reverting. Lines 25,26,27, and 29 should not be included.

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
```diff
@@ -0,0 +1,70 @@
...
+          </div>
+        </div>
+      </div>
+      <div class="grid justify-content-center">
+        <div class="col-3">
+          <button
+            type="button"
+            pButton
+            pRipple
+            [disabled]="!form.valid"
```
sasha-dresden: Should there also be a check here to see if the value is dirty? One of the requirements was that this button be disabled until a value is entered.

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    const currentYear = new Date().getFullYear();
+    this.yearOptions = Array.from({ length: this.numberOfYearOptions }, (value, index) =>
+      (currentYear - index).toString(),
+    );
+    this.yearFormControl.setValue(this.yearOptions[0]);
+  }
+
+  updateLine6a(): void {
+    if (this.form.valid) {
+      if (this.yearFormControl.value && this.newAmountFormControl.value) {
```
sasha-dresden: Not sure if this is desired behaviour, but it's not possible to update the value to 0 because if this.newAmountFormControl.value is 0, then this if statement is false.

sasha-dresden (changes requested):

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.ts
dheitzer: nice catch - updated to check for just null since it can't be undefined

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
dheitzer: good catch - updated

front-end/src/app/layout/header/header-links/header-links.component.html
dheitzer: removed these lines

sasha-dresden (approved):

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #2288, opened 2024-10-23
feature/1720 - Set link to prod link on prod, stage link everywhere else

sasha-dresden:
Issue [FECFILE-1720](https://fecgov.atlassian.net/browse/FECFILE-1720)

### DIFF
--- a/front-end/src/app/committee/committee-info/committee-info.component.ts
+++ b/front-end/src/app/committee/committee-info/committee-info.component.ts
@@ -41,17 +41,12 @@ export class CommitteeInfoComponent extends DestroyerComponent implements OnInit
     'custodian_name_full',
   ];
 
-  link: string;
-
   constructor(
     private store: Store,
     private fb: FormBuilder,
     private readonly changeDetectorRef: ChangeDetectorRef,
   ) {
     super();
-    this.link = environment.production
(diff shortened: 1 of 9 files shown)


front-end/src/app/committee/committee-info/committee-info.component.ts
```diff
@@ -40,12 +41,17 @@ export class CommitteeInfoComponent extends DestroyerComponent implements OnInit
...
+  link: string;
+
   constructor(
     private store: Store,
     private fb: FormBuilder,
     private readonly changeDetectorRef: ChangeDetectorRef,
   ) {
     super();
+    this.link = environment.production
+      ? 'https://webforms.stage.gov/webforms/form1/index.htm'
```
toddlees: can we make the link an environment variable rather than storing both in the component?

toddlees (changes requested):

front-end/src/app/committee/committee-info/committee-info.component.ts
sasha-dresden: I was on the fence about making it an environment variable but I thought maybe this was the only place this link would be used, but I do like it better as an environment variable. Updated

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #2322, opened 2024-10-31
feature/1629 - Attempt to fix date validation

sasha-dresden:
Issue [FECFILE-1629](https://fecgov.atlassian.net/browse/FECFILE-1629)

### DIFF
--- a/front-end/src/app/reports/f1m/main-form/main-form.component.html
+++ b/front-end/src/app/reports/f1m/main-form/main-form.component.html
@@ -251,6 +251,12 @@
             <div class="field">
               <label for="name">AFFILIATED DATE</label>
               <p-calendar
+                (onFocus)="calendarOpened = true"
+                (onClose)="
+                  calendarOpened = false;
+                  SchemaUtils.onBlurValidation(form.get('affiliated_date_form_f1_filed'), calendarOpened)
+                "
+                (onBlur)="SchemaUtils.onBlurValidation(form.get('affiliated_date_form_f1_filed'), calendarOpened)"
                 inputId="affiliated_date_form_f1_filed"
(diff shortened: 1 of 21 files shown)


front-end/src/app/reports/f1m/main-form/main-form.component.html
```diff
@@ -251,6 +251,12 @@ <h2>Affiliation</h2>
             <div class="field">
               <label for="name">AFFILIATED DATE</label>
               <p-calendar
+                (onFocus)="calendarOpened = true"
```
toddlees, line 254: could you make this an onFocus method in the component

front-end/src/app/reports/f1m/main-form/main-form.component.html
```diff
@@ -251,6 +251,12 @@ <h2>Affiliation</h2>
...
+                (onClose)="
+                  calendarOpened = false;
```
toddlees: could you make this an onClose method in the component where the form field name is the argument?

toddlees (changes requested):

front-end/src/app/reports/f1m/main-form/main-form.component.html
sasha-dresden: I left onFocus alone

front-end/src/app/reports/f1m/main-form/main-form.component.html
sasha-dresden: I updated onClose and onBlur to be a function which would move the SchemaUtils.onBlurValidation() function call to the ts instead of the html

toddlees (approved):

PR #2348, opened 2024-11-06
Update README.md local dev instructions

lbeaufort:
add `npm install` instructions, replace `ng serve` with `npx -p @angular/cli ng serve`

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,9 @@ From within the front-end directory, install packages with
 ```
 npm install
 ````
-
 and run the application with the command:
 ```
-ng serve
+npx -p @angular/cli ng serve
 ```
 to start a local server for the application. The front-end can then be accessed through your browser at port 4200.
 
--- a/README.md
+++ b/README.md
@@ -26,10 +26,15 @@ A Snyk authentication token is needed and should be set as the SNYK_AUTH_TOKEN e
 
 ### Running the Front-End locally
 
-From within the front-end directory, run the command:
-`
-    ng serve
-	`
+From within the front-end directory, install packages with 
+```
+npm install
+````
+
+and run the application with the command:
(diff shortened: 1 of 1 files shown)


dheitzer (approved):

lbeaufort:
@dheitzer thanks for your review! Do you think we need to add `npm i -g @angular/cli` too? A simple `npm install` wasn't sufficient for me.

dheitzer:
> @dheitzer thanks for your review! Do you think we need to add `npm i -g @angular/cli` too? A simple `npm install` wasn't sufficient for me.

Ah, I'm sorry I didn't notice the question.  Actually, I think we may want to change `ng serve` to `npx -p @angular/cli ng serve`.  This way, the installation is ephemeral and avoids the risk of a global @angular/cli becoming an outdated version.  What do you think? @toddlees @Elaine-Krauss-TCG @sasha-dresden curious if anyone else has any thoughts too (this is how I run locally).

lbeaufort:
> > @dheitzer thanks for your review! Do you think we need to add `npm i -g @angular/cli` too? A simple `npm install` wasn't sufficient for me.
> 
> Ah, I'm sorry I didn't notice the question. Actually, I think we may want to change `ng serve` to `npx -p @angular/cli ng serve`. This way, the installation is ephemeral and avoids the risk of a global @angular/cli becoming an outdated version. What do you think? @toddlees @Elaine-Krauss-TCG @sasha-dresden curious if anyone else has any thoughts too (this is how I run locally).

Thanks again @dheitzer - I tested `npx -p @angular/cli ng serve` and that worked well - happy to update the instructions unless anyone objects. I'll take care of this next week

dheitzer (approved):

Issue #1118, opened 2023-05-23
Allow user to enter Multistate Independent Expenditure on Schedule E


GreggMoreland:
Business Reason 
 

As a Committee Administrator, I will be able to enter a **”Multistate Independent Expenditure"** transaction on Schedule E of a Form 3X report so that I can accurately report on my committee's transactions.

 Acceptance Criteria 
 
**Given** I am a Committee Administrator editing an open Form 3X report
**Then** go to Add a disbursement
**Then** I can enter a transaction with the type “Multistate Independent Expenditure" under the "Independent Expenditures" Accordion link  
**Then** I will be required to enter the following information (unless otherwise noted):

 Payee 
#
-	**CONTACT TYPE**
Dropdown options: Organization, Individual
*Note - please default to Organization.

---
_if CONTACT TYPE = Organization_
-	**ORGANIZATION NAME** (Required)

---
_if CONTACT TYPE = Individual_
-	**LAST NAME** (Required)
-	**FIRST NAME** (Required)
-	**MIDDLE NAME** (Optional)
-	**PREFIX** (Optional)
-	**SUFFIX** (Optional)

# Address 
#
-	**STREET ADDRESS** (Required)
-	**APARTMENT, SUITE, ETC.** (Optional)
-	**CITY** (Required)
-	**STATE/PROVINCE** (Required)
-	**ZIP/POSTAL CODE** (Required)

# Expenditure information 
#
-	**DISSEMINATION DATE** (At least one date is required. If one is entered, the other is optional)
-	**DISBURSEMENT DATE** ((At least one date is required. If one is entered, the other is optional)
-	**MEMO ITEM** (Optional)
-	**AMOUNT** (Required, can be negative)
-	**CALENDAR YEAR-TO-DATE** (Required)
-- Amount Displayed will be the total of contributions that are:
--- For the selected ELECTION CODE/ELECTION YEAR and Candidate selected
--- within Aggregation Group of "INDEPENDENT_EXPENDITURE"
--- _from_ January 1 of report year _to_ the Disbursement Date* of the transaction. 
--- (**IF** Disbursement Date is not entered, then use Dissemination Date of the transaction- 
.

# Candidate information 
#
-	**SUPPORT/OPPOSE CODE** (Required, values Support, Oppose)
**And** the user will be required to select a Candidate using the LOOKUP, and upon selection the following will be populated:

-	**CANDIDATE ID** (Optional)
[...]

PR #2350, opened 2024-11-07
feature/1686 - Fix district not populating on candidate selection of no district House State

sasha-dresden:
Issue [FECFILE-1686](https://fecgov.atlassian.net/browse/FECFILE-1686)

### DIFF
--- a/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
+++ b/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
@@ -329,17 +329,6 @@ export class ContactDialogComponent extends DestroyerComponent implements OnInit
 
   updateContact(contact: Contact) {
     this.contact = contact;
-    if (
-      this.contact.candidate_office === CandidateOfficeTypes.HOUSE &&
-      this.contact.candidate_state &&
-      this.contact.candidate_district === ''
-    ) {
-      this.candidateDistrictOptions = LabelUtils.getPrimeOptions(
-        LabelUtils.getCongressionalDistrictLabels(this.contact.candidate_state),
-      );
-      this.contact.candidate_district = '00';
-    }
-
     this.form.patchValue(contact);
   }
 
--- a/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
+++ b/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
@@ -329,6 +329,13 @@ export class ContactDialogComponent extends DestroyerComponent implements OnInit
 
   updateContact(contact: Contact) {
     this.contact = contact;
+    if (this.contact.candidate_state && this.contact.candidate_district === '') {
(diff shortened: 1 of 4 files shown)


front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
```diff
@@ -329,6 +329,13 @@ export class ContactDialogComponent extends DestroyerComponent implements OnInit
 
   updateContact(contact: Contact) {
     this.contact = contact;
+    if (this.contact.candidate_state && this.contact.candidate_district === '') {
+      this.candidateDistrictOptions = LabelUtils.getPrimeOptions(
+        LabelUtils.getCongressionalDistrictLabels(this.contact.candidate_state),
+      );
+      this.contact.candidate_district = '00';
+    }
+
```
Elaine-Krauss-TCG: This is setting the candidate district to 00 for every Senate candidate, _buuuuuut_ it's immediately getting cleared out _somehow_.  Visually, I don't ever see a 00 in the district field for senators retrieved with Contact Lookup, but using my browser's debugger, the assignment on line #336 is getting run.

I'd recommend adding one further check to ensure that it's a House candidate, just to be safe.

Elaine-Krauss-TCG (changes requested):

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
sasha-dresden: Oh good catch. Limited it to just be done if the type is House.

Elaine-Krauss-TCG (approved):

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
toddlees: Looks like we're just addressing the symptom here.  When we get an alaska from the api, we're actually receiving a district.  https://github.com/fecgov/fecfile-web-app/commit/cc709bd32d566017974ac0360a36dd3715f9852d seems to be overwriting it.  Can you look into https://github.com/fecgov/fecfile-web-app/issues/1118 and see what problem matt was trying to address by setting it to '' when the district is '00' so that we can handle that scenario and still pull in district from candidates

toddlees (changes requested):

sasha-dresden:
Good point @toddlees I looked into it and I believe the issue Matt was trying to fix was that sometimes we would get data from FEC that would have a House candidate with a  candidate state of 'US' and a district of '00'. And since 'US' isn't an applicable state, we were clearing the state and the district. But the district clearing was too aggressive. It was clearing when the district was '00' which is acceptable.
I updated it so it would clear the district if the state was 'US' which I believe was the intention behind this original bug fix.
I also added in some tests to cover these two situations.

toddlees (approved):

PR #2370, opened 2024-11-15
Feature/1786 - Fix Date issues

sasha-dresden:
Issue [FECFILE-1786](https://fecgov.atlassian.net/browse/FECFILE-1786)

### DIFF
--- a/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
+++ b/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
@@ -88,18 +88,12 @@
       <ng-container *ngIf="isElectionReport()">
         <div class="election-info col-6 grid">
           <div class="col-12">
-            <div class="field">
-              <label for="date_of_election">ELECTION ON</label>
-              <p-calendar
-                (onFocus)="calendarOpened = true"
-                (onClose)="validateDate('date_of_election', false)"
-                (onBlur)="validateDate('date_of_election', calendarOpened)"
-                id="date_of_election"
-                name="date_of_election"
-                formControlName="date_of_election"
-              ></p-calendar>
-              <app-error-messages fieldName="date_of_election" [formSubmitted]="formSubmitted"></app-error-messages>
-            </div>
+            <app-calendar
+              fieldName="date_of_election"
+              [form]="form"
+              [formSubmitted]="formSubmitted"
+              label="ELECTION ON"
+            ></app-calendar>
           </div>
           <div class="col-12">
             <div class="field">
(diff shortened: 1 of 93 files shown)


front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -136,9 +134,9 @@ export class CreateF3XStep1Component extends DestroyerComponent implements OnIni
           isElectionYear,
           filingFrequency,
         );
-        this.form.patchValue({ coverage_from_date, coverage_through_date });
-      } else {
-        this.form.patchValue({ coverage_from_date: null, coverage_through_date: null });
+
+        this.coverageFromDate.control.setValue(coverage_from_date);
```
toddlees: Does this new approach not let us patchValues the way we did before?

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.html
```diff
@@ -7,42 +7,20 @@
...
-          <p-calendar
-            (onFocus)="calendarOpened = true"
-            (onClose)="validateDate(templateMap['date'], false)"
-            (onBlur)="validateDate(templateMap['date'], calendarOpened)"
-            inputId="date"
-            name="date"
-            [formControlName]="templateMap['date']"
-          ></p-calendar>
-          <app-error-messages
-            *ngIf="!templateMap['date2']"
```
toddlees, line 40: Looks like we need to not show an error if there's no date2

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
```diff
@@ -130,40 +124,22 @@ <h3 class="m-0">Covering period</h3>
...
-            name="coverage_from_date"
-            formControlName="coverage_from_date"
-          ></p-calendar>
-          <app-error-messages
-            [form]="form"
-            fieldName="coverage_from_date"
-            [formSubmitted]="formSubmitted"
-          ></app-error-messages>
-        </div>
+        <app-calendar
```
toddlees, line 127: validating the to date on from date change doesn't seem to be working anymore
image
On DEV:
image

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.ts
```diff
@@ -142,24 +142,30 @@ export class LoanTermsDatesInputComponent extends BaseInputComponent implements
...
     }
   }
 
+  /**
+   * Alternates between a date form control and a string form control.
+   * The new CalendarComponent replaces the string form control with a Date control.
+   * This happens on Init, which means when the ngIf condition is met.
+   * This means when returning to the string control we need to recreate the control.
+   * @param newDueDateSetting
+   */
```
toddlees, line 170: This conversion should no longer happen.  Switching the type should clear the value https://fecgov.atlassian.net/browse/FECFILE-1786

toddlees (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
sasha-dresden: After checking patchValues still works just fine. This was a bit of a leftover from when I was first trying to figure out how to handle losing validators when a new control is created. Returned it to the old patchValues. Also returned the patch value to null else condition.

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.html
sasha-dresden: Oh good catch. In order to resolve this and the custom error message for app-error-messages I needed to pass two additional fields to the calendar, but doing so resolved both issues.

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
sasha-dresden: Oh this was tricky. I thought I'd caught all of the validator issues when I directly copied the validators from the old control to the new one, but this showed a second area of problem. When the validator in question was taking in a second control and that second control no longer exists. Was able to fix this by passing the form and the field name to allow the validator to retrieve the correct control at validation time. This is something we'll want to be aware of in the future if we ever add any additional validators comparing two date controls.

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.ts
sasha-dresden: Ah, I totally misread that. I thought it was specifically talking about only the error messages being cleared, not the values themselves as well. I made some updates so now both the due date field and the interest rate field get set to empty when the associated setting field is changed.

toddlees:
works beautifully

toddlees (approved):

PR #2390, opened 2024-11-20
feature/1785 - Add trailing slash urls

sasha-dresden:
Issue [FECFILE-1785](https://fecgov.atlassian.net/browse/FECFILE-1785)
API: [PR1239](https://github.com/fecgov/fecfile-web-api/pull/1239)


toddlees (changes requested):
We can't assume that all endpoints need a trailing slash.
image

For now we can correct the urls that are not matching what they should be on the api

toddlees (approved):

PR #2442, opened 2024-12-06
Feature/1880 - Fix for missing date errors

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-1880)

Related PRs:

### DIFF
--- a/front-end/cypress/e2e/F3X/disbursements.cy.ts
+++ b/front-end/cypress/e2e/F3X/disbursements.cy.ts
@@ -97,17 +97,27 @@ describe('Disbursements', () => {
       '',
       'date_signed',
     );
-
     PageUtils.clickButton('Save');
+
+    // Check that fields saved correctly
     PageUtils.clickLink('Independent Expenditure');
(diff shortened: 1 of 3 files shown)


sasha-dresden (approved):

sasha-dresden:
Couple things I noticed while messing around with this and I think they are unrelated bugs. For Independent Expenditures where you have the disbursement and dissemination dates. If you have the version where error messages where showing up under both fields, the message under Disbursement Date says "This is a required field." However, with the new version that error message will never display. The logic is only looking to see if there's that second field and if there is, it will never show an error message under this one. It feels like the error message should still show when the dissemination date has no error messages. 

Secondly, the error message on that said "required" however I was able to save without a disbursement date, so it feels like it's not validating it properly.

Lastly, the memo code warning is erratic. If you select a dissemination date out of scope the warning pops up, but then if you tab through the disbursement date and leave it blank, the error goes away. Then if you tab into the dissemination date again, the error comes back.

toddlees (approved):

PR #2532, opened 2025-02-08
feature-1972 - Update to angular 19

sasha-dresden:
Issue [FECFILE-1972](https://fecgov.atlassian.net/browse/FECFILE-1972)


sasha-dresden:
So I thought it might be a good idea to classify some of these changes cause there's a lot....

- PrimeNG had quite a few big changes
    1. Big component name changes
        - Dropdown -> Select 
        - Calendar -> DatePicker 
        - OverlayPanel -> Popover 
        - Sidebar -> Drawer
    2. Accordion component changed structure
    3. Major style changes
        - generic .p-highlight replaced with component specific versions like .p-select-option-selected
        - .p-fluid replaced with a fluid property on each component
    4. Message/Toast components changed

- Angular moving away from modules to a standalone component structure, which meant all the module files went away and instead each individual component imports only what it needs.
- Angular testing had some changes that were deprecated and going away.
- Angular had a new inject() property that replaced the need to pass services in the constructor, which actually helped a lot with some of the inheritance constructors we had where we were declaring a service in one constructor just so it could be passed up to the parent where it was actually used.
- Angular introduced a replacement for *ngIf, *ngFor, and *ngSwitch, which definitely makes the code a lot more readable.


I think that's all the major things, some of these led to other changes like to the e2e tests requiring a bit of refactoring for things that were dependent upon the structure of a PrimeNG component which changed or some other changes I made trying to get it to pass SonarQube. 

Still didn't get full passing on SonarQube but I think it might actually be at a higher passing percentage than it was at before this. The few changes I couldn't get to resolve were outstanding issues that weren't related to these changes but since they were in a touched file they got flagged. Like the new route files which replaced the old routing module files all got flagged with duplication even though the lines being flagged weren't changed lines.

toddlees (approved):

toddlees:
Approved with the caveat that we'll be addressing the bugs in release-56 branch

PR #2570, opened 2025-02-28
Feature/2044 fix aggregates in transaction form

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2044)

Related PRs:


Elaine-Krauss-TCG:
The issue raised by SonarCloud is not an issue.  The 'await' used in the new unit test is actually essential, and the test doesn't work without it.

toddlees (approved):

PR #2590, opened 2025-03-06
FECFILE-1974: Feature Flag for Schedule F transactions

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-1974

Related PRs:


front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.ts
```diff
@@ -343,6 +346,10 @@ export class TransactionTypePickerComponent extends DestroyerComponent implement
     return !getTransactionTypeClass(transactionTypeIdentifier);
   }
 
+  showTransaction(transactionTypeIdentifier: string): boolean {
```
toddlees, line 349: That's the idea.  Since there's a large set of filtering rules in the getTransactionTypes() function, what do you think of filtering it out of the list there?  My thoughts are that we'd being doing filtering logic in less places, and they wouldn't be included in the angular template loop 
image

danguyf: Yeah, that's much better.  Let's do that.

toddlees (approved):

PR #2597, opened 2025-03-11
feature/1987 - Create two committee user guard and dialog

sasha-dresden:
Issue [FECFILE-1987](https://fecgov.atlassian.net/browse/FECFILE-1987)


Elaine-Krauss-TCG:
The guard only triggers on first logging in and on page refreshes, but arguably that's fine given #2082 is following this.  (Edit: 2082 does in fact fix this)

Side note: this breaks E2E testing

But the code looks great!  Other than the E2E stuff, I approve 👍

toddlees (changes requested):
needs to update e2e tests

sasha-dresden:
Updated e2e test to account for this.

front-end/src/app/shared/components/second-committee-admin-dialog/second-committee-admin-dialog.component.html
```diff
@@ -0,0 +1,42 @@
+<p-dialog
+  [visible]="memberService.isOnlyOne()"
```
toddlees: let's get rid of the X in the top right of the dialog.

front-end/src/app/shared/services/committee-member.service.ts
```diff
@@ -10,6 +13,33 @@ export class CommitteeMemberService implements TableListService<CommitteeMember>
...
+    loader: ({ request }) => {
+      if (request.committee) {
+        return this.getMembers();
+      }
+      return Promise.resolve([]);
+    },
+    defaultValue: [],
+  });
+
+  public readonly isOnlyOne = computed(() => {
```
toddlees: What do you think about a name that is more specific?  Like `needsSecondAdmin` or `hasOneAdmin`

front-end/src/app/shared/services/committee-member.service.ts
```diff
@@ -10,6 +13,33 @@ export class CommitteeMemberService implements TableListService<CommitteeMember>
...
+    if (
+      Roles[this.user$().role as keyof typeof Roles] !== Roles.COMMITTEE_ADMINISTRATOR ||
+      this.members$.value().length < 1
+    )
+      return false;
```
toddlees, line 27: I'm also getting unexpected results from this signal.  If i remove the other committee admin and trigger the guard, the members value is []

I realize this isn't really going to be exercised for the time being because the user should be adding the second CA on login, but we want to know that these resources are behaving how they should be

toddlees (changes requested):

front-end/src/app/shared/components/second-committee-admin-dialog/second-committee-admin-dialog.component.html
sasha-dresden: Oh good idea. It was nonfunctional but not showing it is good practice.

front-end/src/app/shared/services/committee-member.service.ts
sasha-dresden: Yea, naming things is hard. I like needsSecondAdmin much better.

front-end/src/app/shared/services/committee-member.service.ts
sasha-dresden: This was actually a bit of a tricky issue. The problem is dealing with signals which are synchronous but allow for reactive data, and with asynchronous promises. I had attempted to use resource. I did come up with a solution that did work with resource. I really liked that solution, however, after doing more research I realized resource is still in experimental status so I'm not sure they're quite ready to be used. My solution essentially added an asynchronous version of reload so you can call whenever you needed to get fresh data right now.

But since resource is experimental I came up with a version that just uses signals. The member signal now gets updated whenever the getMembers, addMember, or deleteMember promises are called. So to get the latest you would just use getMembers() and as soon as that is done, the signal is updated. This is the same process we had used with RXJS BehaviourSubjects.

toddlees (approved):

PR #2603, opened 2025-03-12
Feature/2082

sasha-dresden:
Issue [FECFILE-2082](https://fecgov.atlassian.net/browse/FECFILE-2082)

API PR: [PR1385](https://github.com/fecgov/fecfile-web-api/pull/1385)

### DIFF
--- a/front-end/cypress/e2e/pages/loginPage.ts
+++ b/front-end/cypress/e2e/pages/loginPage.ts
@@ -69,9 +69,10 @@ function loginDotGovLogin() {
   cy.get('.committee-list .committee-info').first().click();
   cy.wait('@ActivateCommittee');
 
+  // Wait for the reports page to load
+  cy.contains('Manage reports').should('exist');
+
   // Create second create admin after logging in to make pop-up go away
-  cy.visit('/dashboard');
-  cy.wait(500);
   const alias = PageUtils.getAlias('');
   cy.get(alias)
     .find('[data-cy="second-committee-email"]')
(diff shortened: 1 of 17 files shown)


Elaine-Krauss-TCG:
I can change the role of the second committee member to Manager, which I don't think I should be able to do

sasha-dresden:
Oh that's a very good point. I updated it so you don't get access to the dropdown at all of an administrator if there's only two. I suppose if we ever add additional features to the kebab, like if you could edit name or something then we might need to readdress this in the future, but for now, with only these two options, not showing it seems the right choice.

front-end/cypress/e2e/pages/loginPage.ts
```diff
@@ -67,7 +68,20 @@ function loginDotGovLogin() {
   cy.wait('@GetCommitteeAccounts');
   cy.get('.committee-list .committee-info').first().click();
   cy.wait('@ActivateCommittee');
-  //cy.visit('/dashboard');
+
+  // Create second create admin after logging in to make pop-up go away
+  cy.visit('/dashboard');
+  cy.wait(500);
```
Elaine-Krauss-TCG: We've been phasing out fixed-time wait calls, so I took a few minutes to make an alternative solution.  I've opened a PR into this PR [here](https://github.com/fecgov/fecfile-web-app/pull/2610)

Elaine-Krauss-TCG (approved):
The modal works great, and you can't delete or modify the second admin anymore.  I've left a comment regarding a wait() call in the E2E script, but I don't consider it a deal-breaker

sasha-dresden:
Merged that PR into this.

PR #2605, opened 2025-03-13
feature/2085 - Set calendar to not reset control

sasha-dresden:
Issue [FECFILE-2085](https://fecgov.atlassian.net/browse/FECFILE-2085)

Made 2 primary changes.
1. All of the SubscriptionFormControls that we create can handle both string and Date formats. This allows us to no longer need to recreate the Date controls in the CalandarComponent. We also don't need to toggle between a Date control and a string control in the Loan Terms Due Date field.
2. Used the schema to determine if the type should be a date control or not. I was a little iffy on this one, this was a second way I was thinking of handling the creation of the SubscriptionFormControl so it would know beforehand if it is a Date or string type. However it wouldn't resolve the issue with the Due Date field. I left it in because we were previously setting the Date types to update on 'submit' so I left that functionality in.


Elaine-Krauss-TCG (approved):

toddlees (approved):
Great solution to our problem and simplification of our code

PR #2606, opened 2025-03-14
feature/2097 - Update contact changes logic to account for Senate candidates having '00' or null as district

sasha-dresden:
Issue [FECFILE-2097](https://fecgov.atlassian.net/browse/FECFILE-2097)

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
@@ -73,9 +73,6 @@ export class TransactionContactUtils {
   ): any[] {
     return Object.entries(contactConfig)
       .map(([field, property]: string[]) => {
-        // For Senate, 00 is used for district but not displayed in form, treated as NULL. It cannot be changed.
-        if (field === 'candidate_district' && form.get(templateMap.candidate_office)?.value === 'S') return undefined;
-
         const contactValue = contact[property as keyof Contact];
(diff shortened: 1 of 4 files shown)


front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
```diff
@@ -73,6 +73,9 @@ export class TransactionContactUtils {
   ): any[] {
     return Object.entries(contactConfig)
       .map(([field, property]: string[]) => {
+        // For Senate, 00 is used for district but not displayed in form, treated as NULL. It cannot be changed.
```
toddlees: Good find!  Looks like this is coming from the openfec data.  Let's fix this closer to the source.  What about in onFecApiCandidateLookupDataSelect().  We're doing something similar where we set the district to '' when the state is US.  Same idea but we could also include if the office is `S` to the condition

This would also fix the case where someone makes the contact using a lookup on the contact page, it won't result in us saving 00 in the district

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
sasha-dresden: Oh good point! Just adding a simple or condition there does the trick.

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #2638, opened 2025-04-02
Feature/form 3

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-1979
VAL:https://github.com/fecgov/fecfile-validate/pull/346
APP:https://github.com/fecgov/fecfile-web-app/pull/2638
API:https://github.com/fecgov/fecfile-web-api/pull/1416


front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -0,0 +1,217 @@
...
+  ],
+})
+export class CreateF3Step1Component extends FormComponent implements OnInit {
+  private readonly store = inject(Store);
+  private readonly form3Service = inject(Form3Service);
+  private readonly messageService = inject(MessageService);
+  protected readonly router = inject(Router);
+  private readonly activatedRoute = inject(ActivatedRoute);
+  readonly formProperties: string[] = [
+    'report_type_category',
```
sasha-dresden, line 57: I'm not sure if this is a difference between Form 3 and 3X. In Form 3X there is filing_frequency. I don't know if that just doesn't apply to Form 3 or was just missed.
These two are so similar I almost want to make them use inheritance, but that brings its own pros and cons. Since it's just two I can understand not going for inheritance.

front-end/src/app/reports/f3/routes.ts
```diff
@@ -0,0 +1,112 @@
...
+  },
+  {
+    path: 'create/step1/:reportId',
+    title: 'Create a report',
+    component: CreateF3Step1Component,
+    canActivate: [ReportIsEditableGuard],
+    runGuardsAndResolvers: 'always',
+  },
+  // {
+  //   path: 'summary/:reportId',
```
sasha-dresden, line 38: Will we eventually get these added in? Otherwise we might want to delete these commented out routes.

front-end/src/app/reports/report-list/report-list.component.ts
```diff
@@ -85,6 +85,8 @@ export class ReportListComponent extends TableListBaseComponent<Report> implemen
     }
 
     switch (item.report_type) {
+      case ReportTypes.F3:
```
sasha-dresden, line 88: Maybe it would be a good idea to combine F3, F3X, and F24 cases since they all route the same?

front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
toddlees: Yep no filing frequency in F3.  They are super similar.  I'm thinking we can either wait until F3L or something comes around and reevaluate then, or we could even just take a look at those forms and plan from there

front-end/src/app/reports/f3/routes.ts
toddlees: Yep we will.  I drew the line at summary because I figured that'd be it's own big chunk for next sprint

front-end/src/app/reports/report-list/report-list.component.ts
toddlees: I like it!

front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.ts
```diff
@@ -93,7 +93,15 @@ export class TransactionTypePickerComponent extends DestroyerComponent {
 
   transactionGroups: Signal<TransactionGroupTypes[]> = computed(() => {
     if (this.category() === 'disbursement') {
-      if (this.report().report_type === ReportTypes.F3X) {
+      if (this.report().report_type === ReportTypes.F3) {
+        return [
+          ScheduleBTransactionGroups.OPERATING_EXPENDITURES,
+          ScheduleBTransactionGroups.CONTRIBUTIONS_EXPENDITURES_TO_REGISTERED_FILERS, // includes ScheduleFTransactionTypes.*
```
sasha-dresden, line 99: So this isn't related to this ticket, but I just noticed this comment and I looked into ScheduleBTransactionGroups.CONTRIBUTIONS_EXPENDITURES_TO_REGISTERED_FILERS and it does look like it is now a mix of Schedule B and Schedule F items. I think we might want to correct this, though I don't think this is the ticket for that.

toddlees: Yeah that came up when we added it to the accordion.  I think since the groups are no longer a 1 to 1 with schedules, they should probably be redefined outside the context of a schedule.

toddlees: @sasha-dresden https://fecgov.atlassian.net/browse/FECFILE-2129

lbeaufort (approved):
Exciting work! 🎉

front-end/src/app/shared/utils/reatt-redes/reattributed.utils.ts
```diff
@@ -11,7 +11,9 @@ export class ReattributedUtils {
       if (transaction.report_ids?.includes(activeReportId as string)) {
         transaction.contribution_purpose_descrip = 'See reattribution below.';
       } else {
-        transaction.contribution_purpose_descrip = `(Originally disclosed on ${transaction.getForm3X()?.report_code_label ?? ''}.) See reattribution below.`;
+        const report_code: string =
+          transaction.getForm3X()?.report_code_label ?? transaction.getForm3()?.report_code_label ?? '';
```
sasha-dresden, line 15: Hmm, this feels a bit clunky having 2 sets of nullish coalescing. It works just fine and is perfectly readable. I'd just mark this down as another spot we might want to consider when we get more forms especially form 3 variations. Any more than 2 and we might want to just create a specific helper function on the transaction or something to say get any form 3 type form.

toddlees: agreed! I think this is also a place where we can abstract the the depth of forms into a common location.

PR #2648, opened 2025-04-04
2128 - Move labels for component, fix phone E2E tests

lbeaufort:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2128

I'm not sure how these tests were passing in Circle.

**Note**
I noticed `telephone` was missing some labels in the html 

```
<input _ngcontent-ng-c3418061334="" type="tel" aria-labelledby="telephone-label" class="iti__tel-input" autocomplete="off" data-intl-tel-input-id="0" style="padding-left: 75px;">

<input _ngcontent-ng-c1258949998="" type="text" pinputtext="" id="zip" formcontrolname="zip" class="p-inputtext p-component ng-untouched ng-pristine ng-invalid" ng-reflect-name="zip" pc38="">
```

### DIFF
--- a/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
+++ b/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
@@ -197,7 +197,7 @@
         <div class="col-6">
           <div class="field">
             <label id="telephone-label" for="telephone">TELEPHONE (OPTIONAL)</label>
-            <app-fec-international-phone-input> </app-fec-international-phone-input>
+            <app-fec-international-phone-input formControlName="telephone"> </app-fec-international-phone-input>
             <app-error-messages
               [form]="form"
               fieldName="telephone"
--- a/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
+++ b/front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
@@ -197,8 +197,7 @@
         <div class="col-6">
           <div class="field">
             <label id="telephone-label" for="telephone">TELEPHONE (OPTIONAL)</label>
-            <app-fec-international-phone-input id="telephone" formControlName="telephone" labelName="telephone-label">
-            </app-fec-international-phone-input>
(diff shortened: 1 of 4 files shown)


front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
```diff
@@ -197,8 +197,7 @@ <h3>Address</h3>
         <div class="col-6">
           <div class="field">
             <label id="telephone-label" for="telephone">TELEPHONE (OPTIONAL)</label>
-            <app-fec-international-phone-input id="telephone" formControlName="telephone" labelName="telephone-label">
```
Elaine-Krauss-TCG, line 200: We also use these parameters in the committee info component for the treasurer's telephone data, but I can see from this that they definitely need to be passed through to the input.

sasha-dresden: Yes, we will also need to update that file. Additionally the formControlName needs to remain on the outer component like this:

`<app-fec-international-phone-input  formControlName="telephone" >`

If it isn't there it would break the onBlur function. I can't test cause apparently this is a mac issue? But I think the issue is that the browser is complaining about a lack of id on the actual input. Since we use different id's in different places, we might want to pass the id as an input parameter like this:
 <app-fec-international-phone-input id="telephone" formControlName="telephone">
 
 And in the ts file
` @Input() id='telephone';`
 
 And then on the <input /> make the id like this 
 `[id]="id"`

sasha-dresden (changes requested):

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.html
Elaine-Krauss-TCG: Good catches!  I'll implement them

lbeaufort:
@sasha-dresden and @Elaine-Krauss-TCG thank you both for your help! I pulled down Elaine's changes and the tests pass on my local. I'll wait for @toddlees to test as well.

danguyf (approved):

PR #2653, opened 2025-04-04
Feature/2107 - Fix error messages being too wide on report coverage dates

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2107)

Related PRs:


sasha-dresden (commented):
This works. But I might recommend to move this up to the form element. As it stands now, the calendar expanding is what breaks this, so putting the max-width limit where you did works, but if in the future we add a calendar outside there we could run into the same issue again. 

Somethng like this?
```
#form {
    max-width: 780px;
    width: 780px;
  }
```
  
Another question is break points. This works well when screen width is at least 800px but what about when it's less. Usually we set specific breakpoints like this:

```
#form {
  max-width: 360px;
  width: 360px;
}

@media (min-width: 576px) {
  #form {
    max-width: 540px;
    width: 540px;
  }
}

@media (min-width: 768px) {
  #form {
    max-width: 720px;
    width: 720px;
  }
}

@media (min-width: 992px) {
  #form {
    max-width: 780px;
    width: 780px;
  }
}
```

danguyf (approved):
Tested successfully.

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.scss
```diff
@@ -6,3 +6,29 @@
...
+@media (min-width: 768px) {
+  #form {
+    max-width: 720px;
+    width: 720px;
+  }
+}
+
+@media (min-width: 992px) {
+  #form {
+    max-width: 780.4px;
```
lbeaufort, line 31: Thanks @Elaine-Krauss-TCG! 

Should this be `780px` to match @sasha-dresden's suggestion, or is the extra `.4` on purpose?

Elaine-Krauss-TCG: I put it at 780.4px, since that was the exact maximum width that I had observed prior to these changes.  I could reduce the width of my browser, and the form would stay at 780.4px until I crossed a threshold (I did not measure what the threshold was).

sasha-dresden (approved):

PR #2668, opened 2025-04-11
feature/2098 - Fix tsconfig breaking e2e

sasha-dresden:
Accidentally broke e2e tests with a tsconfig update. This fixes it. 

Issue [FECFILE-2098](https://fecgov.atlassian.net/browse/FECFILE-2098)


danguyf (approved):
Approving this PR because it makes sense in itself.  I see two other branches with "2098" in the name and I assume that one of the others is meant to address the issue reported in FECFILE-2098, as this one does not so far as I understand.

PR #2697, opened 2025-04-30
FECFILE-2084: hide dashboard

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2084

Related PRs:

Side question: Is using a guard for this semantically impure? Is there a better way that's as simple?


dheitzer (dismissed):
Looks nice! Only suggestion I would make would be to add a show-dashboard.guard.spec.ts to get unit test coverage on that guard.

dheitzer:
RE:  Is using a guard for this semantically impure?

I would say a potentially better solution would be to add[ this block](https://github.com/fecgov/fecfile-web-app/blob/a896431845be7546697c8ceb230a33d61dd8e92b/front-end/src/app/routes.ts#L28) dynamically to the ROUTES array based on the feature flag.  This would prevent /dashboard from being routable at all.

front-end/cypress/e2e/pages/loginPage.ts
```diff
@@ -89,8 +89,6 @@ function loginDotGovLogin() {
       }
     });
   cy.contains('Welcome to FECfile').should('not.exist');
-  cy.visit('/dashboard');
```
toddlees, line 92: good call! not only do we not want to go to the dashboard, but the previous line is in fact sufficient to know for certain that the entirety of this workflow is complete

toddlees (commented):
Everything looks great.  testing locally

toddlees (approved):

PR #2709, opened 2025-05-02
FECFILE-2166: Send to /login if user is not authenticated.

danguyf:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2166

Related PRs:

After testing a few options I decided that this was the safest.

Test this in less than 30 minutes by editing [this line](https://github.com/fecgov/fecfile-web-api/blob/develop/django-backend/fecfiler/settings/base.py#L54).  (I removed the "* 60" to get it to fail in 30s.)

### DIFF
--- a/front-end/src/app/shared/services/login.service.ts
+++ b/front-end/src/app/shared/services/login.service.ts
@@ -24,7 +24,7 @@ export class LoginService extends DestroyerComponent {
 
   public logOut() {
     this.store.dispatch(userLoginDataDiscardedAction());
-    if (!this.isLoggedIn()) {
+    if (!this.isLoggedIn() || !this.userIsAuthenticated()) {
       this.router.navigate(['/login']);
     } else {
       window.location.href = environment.loginDotGovLogoutUrl;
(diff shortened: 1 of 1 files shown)


dheitzer (dismissed):
Nice sleuthing @danguyf.  Tangentially, as part of this ticket I think we should see if we can combine isLoggedIn and userIsAuthenticated into a single function unless we have a good reason for them to be separate.  

If we do in-fact need separate functions, I think we should at least rename one to better describe what it's checking since these functions sound like they should both be returning the same result based on their names (but in the case of the 403 error, they're returning different results). 

Maybe we can sync up next week.

danguyf:
> Tangentially, as part of this ticket I think we should see if we can combine isLoggedIn and userIsAuthenticated into a single function unless we have a good reason for them to be separate.

That is an excellent point.  After further investigation, this was the only place that isLoggedIn was used.  While the difference between it and userIsAuthenticated seems notable if nothing is using it there's no sense keeping it.  I have removed it.

toddlees (approved):

PR #2715, opened 2025-05-05
Feature/2136 - Add CSP report logging

sasha-dresden:
Issue [FECFILE-2136](https://fecgov.atlassian.net/browse/FECFILE-2136)

A few notes. I switched from using nginx to using openresty. This is also provided through the cloudfoundry nginx buildpack. This was done so I could use the lua nginx module, which was needed in order to access the request body inside nginx.
Unfortunately, the nginx buildpack from cloudfoundry did not have "sub_filter" so I needed to find an alternate way to get the same effect. This is what the header_filter_by_lua_block and body_filter_by_lua_block should be achieving. But yes, with this nginx.conf file I was able to get it to log the csp report to cloud.gov in the web app.
image


deploy-config/front-end-nginx-config/nginx.conf
```diff
@@ -1,7 +1,7 @@
 worker_processes 1;
 daemon off;
 
-error_log stderr;
+error_log stderr info;
```
toddlees: Is there a way we can raise the level of the log message being produced in the csp rather than lower the level that gets logged to stderr?

deploy-config/front-end-nginx-config/nginx.conf
```diff
@@ -11,28 +11,77 @@ http {
...
+
+    set $nonce $request_id;
+
+    # CSP headers
     location = /index.html {
       add_header Cache-Control no-cache;
       add_header Access-Control-Allow-Origin null;
-      # Ensure any updates here are reflected in the local serve in angular.json
-      add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'nonce-$request_id'; style-src 'self' 'nonce-$request_id'; connect-src 'self' https://dev-api.fecfile.fec.gov https://stage-api.fecfile.fec.gov https://test-api.fecfile.fec.gov https://api.fecfile.fec.gov; frame-ancestors 'none'; object-src 'none'";
+      add_header Report-To '{"group":"csp-endpoint","max_age":10886400,"endpoints":[{"url":"/csp-report"}]}';
```
toddlees: It seems like using the Report-To header is deprecated.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Report-To

Same with report-uri in the CSP. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/report-uri

Let's use `report-to` in the CSP https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/report-to

deploy-config/front-end-nginx-config/nginx.conf
```diff
@@ -11,28 +11,77 @@ http {
...
+      }
     }
+
+    # CSP report handling using Lua
+    location = /csp-report {
+      content_by_lua_block {
+        ngx.req.read_body()
+        local csp_report = ngx.req.get_body_data()
+        if csp_report then
+          ngx.log(ngx.INFO, 'CSP Report: ', csp_report)
```
toddlees: let's make these level ERROR

toddlees (changes requested):

sasha-dresden:
Updated log level to error and got rid of report-to header. Used report-to directive, which required the reporting-endpoints header. I thought perhaps we should leave report-uri as a fallback for the browsers that don't use report-to yet?
Confirmed with Todd report-uri is out!

image

toddlees:
Updated api-url to be an environment variable and added comments.
Thank you for figuring out all this hard stuff.  It was a learning experience for me definitely!

I have a small amount of hope that we could get the sub module working.  Maybe we can ask the cloudfoundry slack if they know what's up

toddlees (approved):

PR #2716, opened 2025-05-07
Feature/588 1131

dheitzer:
Ticket link:https://fecgov.atlassian.net/browse/FECFILE-588 and https://fecgov.atlassian.net/browse/FECFILE-1131

Related PRs:


front-end/src/app/reports/submission-workflow/submit-report-status.component.ts
```diff
@@ -15,18 +15,19 @@ import { LongDatePipe } from '../../shared/pipes/long-date.pipe';
   selector: 'app-report-summary',
   templateUrl: './submit-report-status.component.html',
   styleUrls: ['./submit-report-status.component.scss'],
-  imports: [Card, NgOptimizedImage, ButtonDirective, Ripple, LongDatePipe],
+  imports: [NgOptimizedImage, ButtonDirective, Ripple, LongDatePipe],
 })
 export class SubmitReportStatusComponent extends DestroyerComponent implements OnInit {
+  reportStatus = ReportStatus;
```
toddlees: We find ourselves having to do this a lot with enums.  I wonder if naming it `reportStatuses` or `reportStatusEnum` would make it clearer that this variable is holding the enum options rather than the status of the report

front-end/src/app/reports/submission-workflow/submit-report-status.component.html
```diff
@@ -1,133 +1,100 @@
...
+          pButton
+          href="https://www.fec.gov/help-candidates-and-committees/filing-reports/upload-errors/"
+          target="_blank"
+          rel="noopener"
+          class="p-button-secondary learn-more-button"
+        >
+          Learn more about your submission error
+        </a>
+        <h3 class="assistance-heading">Electronic Filing Office</h3>
+        <p class="assistance-text">1-800-424-9530, menu option 4 202-694-1307 efiletechsupport&#64;fec.gov</p>
```
toddlees: ac had a typo,  the mockup has this saying "menu option 4, or 202..." rather than "menu option 4 202..."

toddlees (changes requested):
Looks great! One other change is that the "Review" button on the reports page needs to go to the status page instead of the transactions page when the status is failure

front-end/src/app/reports/submission-workflow/submit-report-status.component.html
toddlees: new update.  we're making it match https://www.fec.gov/contact/
Shannon is updating the wireframes.
So the text stays as you've got it and we just need to add some newlines in between

front-end/src/app/reports/submission-workflow/submit-report-status.component.ts
dheitzer: yep makes sense - updated to reportStatusEnum.

front-end/src/app/reports/submission-workflow/submit-report-status.component.html
dheitzer: updated to match the latest wireframe.

dheitzer:
> Looks great! One other change is that the "Review" button on the reports page needs to go to the status page instead of the transactions page when the status is failure

nice catch - updated and added some unit tests for coverage of this bug.

toddlees (approved):

PR #2739, opened 2025-05-16
FECFILE-2218: cookie collision!

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2218
  
Related PRs:
https://github.com/fecgov/fecfile-web-api/pull/1493

### DIFF
--- a/front-end/src/app/shared/services/login.service.ts
+++ b/front-end/src/app/shared/services/login.service.ts
@@ -37,6 +37,12 @@ export class LoginService extends DestroyerComponent {
   }
 
   public async retrieveUserLoginData(): Promise<void> {
+    if (!environment.hasOwnProperty("ffapiTimeoutCookieName")) {
+      console.error("The ffapi_timeout cookie name environment variables is not set.");
+    } else if (!this.cookieService.check(environment.ffapiTimeoutCookieName)) {
+      console.error("The ffapi_timeout cookie is not set.");
+    }
+    
     return this.usersService.getCurrentUser().then((userLoginData) => {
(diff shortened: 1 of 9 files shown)


front-end/src/app/shared/services/login.service.ts
```diff
@@ -53,6 +53,6 @@ export class LoginService extends DestroyerComponent {
   }
 
   public userIsAuthenticated() {
-    return new Date() < new Date(parseInt(this.cookieService.get('ffapi_timeout')) * 1000);
+    return new Date() < new Date(parseInt(this.cookieService.get(environment.ffapiTimeoutCookieName)) * 1000);
```
dheitzer, line 62: similar to API comment, we may want to add a default cookie name here in case the environment var isn't set.  Otherwise, parseInt may fail and break all api requests in the guard.

dheitzer (dismissed):
small request, otherwise looks good!

front-end/src/app/shared/services/login.service.ts
toddlees: I think this is one where failing loudly on missing the configuration is what we want.

We should add a check, though that throws a descriptive error like "api timeout cookie name not defined"

toddlees (changes requested):

front-end/src/app/shared/services/login.service.ts
danguyf: In the case of the web app, it will not compile if the cookie name definition is missing.  I have added a check for it, and a check for whether it is defined, for good measure, though.

toddlees (approved):

PR #2763, opened 2025-05-23
feature/2191 - Update mergeReducer for ngrx localStorageSync to prevent cypress converting zipcodes into dates.

sasha-dresden:
Issue [FECFILE-2191](https://fecgov.atlassian.net/browse/FECFILE-2191)

Underlying issue was that, for some reason I couldn't identify, in cypress, when serializing/deserializing from local storage, the zip code on the committee account was being treated as a date leading to validation issues when that zip was used on the autopopulated section of form 1M creation.

Solution was to override NGRX mergeReducer to force zip codes to be strings.

### DIFF
--- a/front-end/src/main.ts
+++ b/front-end/src/main.ts
@@ -61,8 +61,6 @@ function localStorageSyncReducer(reducer: ActionReducer<AppState>): ActionReduce
     keys: ['committeeAccount', 'singleClickDisabled', 'userLoginData', 'activeReport'],
     storageKeySerializer: (key) => `fecfile_online_${key}`,
     rehydrate: true,
-    mergeReducer,
-    restoreDates: false,
   })(reducer);
 }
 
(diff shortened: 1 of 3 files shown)


front-end/src/main.ts
```diff
@@ -130,3 +132,20 @@ bootstrapApplication(AppComponent, {
     provideHttpClient(withInterceptorsFromDi()),
   ],
 }).catch((err) => console.log(err));
+
```
toddlees: I think https://fecgov.atlassian.net/browse/FECFILE-2234 fixed the zipcode issue without needing to change the merge reducer.  It was a bug with our dependency.  

We should be able to enable the tests without this code change now

toddlees (changes requested):

front-end/src/main.ts
sasha-dresden: Yep. It works just fine with the new package. Reverted changes so it's now just enabling the tests.

toddlees (approved):

PR #2769, opened 2025-05-27
feature/2259 - Force button click

sasha-dresden:
Issue [FECFILE-2259](https://fecgov.atlassian.net/browse/FECFILE-2259)

### DIFF
--- a/front-end/cypress/e2e/pages/loginPage.ts
+++ b/front-end/cypress/e2e/pages/loginPage.ts
@@ -55,6 +55,7 @@ function getLoginIntervalString(sessionDur: number): string {
 }
 
 function loginDotGovLogin() {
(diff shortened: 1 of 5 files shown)


toddlees (changes requested):
It might be the save success toast getting in the way.  Instead of clicking through what we don't know is in the way, let's add a step to loginPage that dismisses the toast after adding a second committee admin

sasha-dresden:
> It might be the save success toast getting in the way. Instead of clicking through what we don't know is in the way, let's add a step to loginPage that dismisses the toast after adding a second committee admin

Yep! That did it. Just told it to click the toast away and it worked without needing to use force.

toddlees (approved):

PR #2787, opened 2025-05-30
Feature/2145 - 2147: 508 Fixes

sasha-dresden:
Issues:
[FECFILE-2145](https://fecgov.atlassian.net/browse/FECFILE-2145)
[FECFILE-2146](https://fecgov.atlassian.net/browse/FECFILE-2146)
[FECFILE-2147](https://fecgov.atlassian.net/browse/FECFILE-2147)

I worked all of these as one unit as the fix for the selects only needed to be done once, but was needed for each of them.


danguyf:
Is `<select> @for` always preferable to `<p-select>`?

front-end/src/app/committee/committee-info/committee-info.component.html
```diff
@@ -64,14 +64,11 @@ <h2>Address</h2>
...
         <label for="state">STATE/PROVINCE</label>
-        <p-select
-          inputId="state"
-          formControlName="state"
-          [options]="stateOptions"
-          [autoDisplayFirst]="false"
-          appendTo="body"
-          ariaLabel="State Dropdown"
-        ></p-select>
+        <select id="state" formControlName="state">
```
dheitzer, line 67: Do we need an ariaLabel on this still? Same with all other `<select>`

dheitzer:
> Is `<select> @for` always preferable to `<p-select>`?

I'm also wondering if you think there's anyway we can continue using p-select? seems like maintaining all the styling manually might not be ideal.

front-end/src/app/shared/components/inputs/input-number/input-number.component.html
```diff
@@ -1,45 +1,48 @@
...
-    (paste)="onPaste($event)"
-    (click)="onInputClick()"
-    (focus)="onInputFocus($event)"
-    (blur)="onInputBlur($event)"
-    [attr.data-pc-section]="'input'"
-    [autofocus]="autofocus"
-  />
-</span>
+<div class="field">
+  <label [for]="inputId()">{{ label() }}</label>
```
dheitzer, line 2: my concern about calling inputId() and label() methods in the template is that change detection will trigger continuous execution of these functions (even if they are just getters).  Not the end of the world if they don't do much, but do you think we could/should avoid defining these as functions? (also other places in this CR).

front-end/src/styles.scss
```diff
@@ -517,3 +517,28 @@ table {
   font-size: 1.5rem;
   margin-left: 0.5rem;
 }
+
+select {
```
dheitzer: I think it would be ideal to avoid having to style select ourselves if there's a reasonable way to use p-select.

front-end/src/styles.scss
```diff
@@ -517,3 +517,28 @@ table {
...
+  font-size: 1rem;
+  font-weight: 400;
+  -webkit-appearance: none; /* Chrome, Safari, Opera */
+  -moz-appearance: none; /* Firefox */
+  appearance: none;
+  background: none;
+  background-color: transparent;
+  background-image: url("data:image/svg+xml,%3Csvg%20xmlns='http://www.w3.org/2000/svg'%20viewBox='0%200%2014%2014'%20fill='%23a1a1aa'%20aria-hidden='true'%3E%3Cpath%20d='M7.01744%2010.398C6.91269%2010.3985%206.8089%2010.378%206.71215%2010.3379C6.61541%2010.2977%206.52766%2010.2386%206.45405%2010.1641L1.13907%204.84913C1.03306%204.69404%200.985221%204.5065%201.00399%204.31958C1.02276%204.13266%201.10693%203.95838%201.24166%203.82747C1.37639%203.69655%201.55301%203.61742%201.74039%203.60402C1.92777%203.59062%202.11386%203.64382%202.26584%203.75424L7.01744%208.47394L11.769%203.75424C11.9189%203.65709%2012.097%203.61306%2012.2748%203.62921C12.4527%203.64535%2012.6199%203.72073%2012.7498%203.84328C12.8797%203.96582%2012.9647%204.12842%2012.9912%204.30502C13.0177%204.48162%2012.9841%204.662%2012.8958%204.81724L7.58083%2010.1322C7.50996%2010.2125%207.42344%2010.2775%207.32656%2010.3232C7.22968%2010.3689%207.12449%2010.3944%207.01744%2010.398Z'/%3E%3C/svg%3E");
```
dheitzer: Unfortunately the use of the `data:` scheme is a csp violation with no good policy workaround.  I ran into this during our csp work and found [this discussion](https://security.stackexchange.com/a/95011/208408) helpful to understand.  

Basically, the solution is to generate the svg as an asset and link to it like [here](https://github.com/fecgov/fecfile-web-app/blob/d0578659c0d70f14ccc5143cc6201452603382d6/front-end/src/app/layout/banner/banner.component.scss#L99)

dheitzer (changes requested):
just a few thoughts

front-end/src/app/committee/committee-info/committee-info.component.html
sasha-dresden: No we don't need arialabel though I suppose it doesn't hurt. Because we have a label and this is now a <select> it can retrieve the accessible name directly from the label, so the accessible name would essentially be "STATE/PROVINCE" whereas I believe the aria label would override it.

front-end/src/app/shared/components/inputs/input-number/input-number.component.html
sasha-dresden: oh these are signals not getters. So they do NOT trigger on change detection. Signal _only_ recompute when their underlying value changes. Because these are inputs that means if the input changed would be the only time it recalculates. Essentially you could think of it like this.

```
labelSubject = new BehaviorSubject<string|undefined>(undefined)
this.labelSubject.subscribe(label => this.label = label)
```
This means the label won't trigger recompilation until the input changes. The other nice thing is you can derive values from them as well, so for example if you wanted to create a combined label that takes the label and the input id you could do something like this:
`readonly idLabel = computed(() => `${this.label()}-${this.inputId()}`);`
And this new one would only recompute when either of those change and it does so automatically.

front-end/src/app/committee/committee-info/committee-info.component.html
dheitzer: good catch! works for me.

front-end/src/app/shared/components/inputs/input-number/input-number.component.html
dheitzer: oh! I still need to get used to the signal syntax.  Sounds great!

sasha-dresden:
AHA I found a way to do it! The issue with <p-select> was that instead of using an an input type element (like input or select) for <p-select> instead it uses <span>. So no matter what I did with p-select if was failing, so I figured just using <select> was a solid solution. However I have been messing around with it today and I found out that if I use a template with an <input> for the form label to latch on to.
Since this is so different I just made a different branch. [PR2802](https://github.com/fecgov/fecfile-web-app/pull/2802)

PR #2811, opened 2025-06-05
feature/1953 - Move security notification consent text into environment variable

sasha-dresden:
Issue [FECFILE-1953](https://fecgov.atlassian.net/browse/FECFILE-1953)

### DIFF
--- a/front-end/src/app/login/security-notice/dev-notice.component.ts
+++ b/front-end/src/app/login/security-notice/dev-notice.component.ts
@@ -1,0 +1,37 @@
+import { Component } from '@angular/core';
+
+@Component({
+  template: `<h1>Security notification</h1>
+    <hr type="solid" class="top-line" />
+    <h2>Warning</h2>
+    <p>
+      This version of the application is for testing new features and functionality and gathering user feedback. Any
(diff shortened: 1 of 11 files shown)


danguyf (approved):
Putting the HTML in the environment files bothers me a little but the alternatives that I've come up with are similarly bothersome so why not?

toddlees (changes requested):
I'm inclined to avoid defining the html in the env files as well.  What do you think of storing a key in the environment object and in the consent file, mapping the key show a specified chunk of html and not showing the others?

sasha-dresden:
> Putting the HTML in the environment files bothers me a little but the alternatives that I've come up with are similarly bothersome so why not?

> I'm inclined to avoid defining the html in the env files as well. What do you think of storing a key in the environment object and in the consent file, mapping the key show a specified chunk of html and not showing the others?

So I talked with @toddlees about this and I think I came up with a cleaner solution. Instead of storing it in the environment, I used dynamic component loading so it will use a different component based on environment. Currently just set to prod/not-prod but could very easily be expanded as necessary. This stores each version in its own angular component.

toddlees (approved):

PR #2814, opened 2025-06-06
feature/1926 - Updated select button styling to match styleguide

sasha-dresden:
Issue [FECFILE-1926](https://fecgov.atlassian.net/browse/FECFILE-1926)

Updated to match [reference](https://fec-pattern-library.app.cloud.gov/components/detail/horizontal.html).
Key changes:
- Border width changed from 3px to 2px.
- Inner padding locked at  8px 20px
- Width of toggle buttons set by content rather than number of buttons
- removed focus ring that appeared when you clicked an option
- Removed left border from right option when selecting left option and removed right border from left option when selecting right option

### DIFF
--- a/front-end/src/assets/styles/theme.css
+++ b/front-end/src/assets/styles/theme.css
@@ -1655,9 +1655,14 @@ p-selectbutton.ng-dirty.ng-invalid > .p-selectbutton > .p-button {
   transition: height 0.15s;
 }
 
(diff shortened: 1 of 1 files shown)


Elaine-Krauss-TCG (approved):

danguyf (approved):
Look good, I can tab through them, they are keyboard selectable.

PR #2820, opened 2025-06-10
feature/478 - Add cookie check service to check if user has cookie access. Also make sure basic page loads if no cookies

sasha-dresden:
Issue [FECFILE-478](https://fecgov.atlassian.net/browse/FECFILE-478)


front-end/src/app/shared/services/cookie-check.service.ts
```diff
@@ -0,0 +1,18 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class CookieCheckService {
+  areCookiesEnabled(): boolean {
+    try {
+      document.cookie = 'cookietest=1';
+      const cookiesEnabled = document.cookie.indexOf('cookietest=') !== -1;
+      document.cookie = 'cookietest=1; expires=Thu, 01-Jan-1970 00:00:01 GMT';
+      return cookiesEnabled;
```
danguyf: Firefox allows you to disable local storage independently of cookies, and conversely when cookies are disabled still allows local storage access for whitelisted sites.

Should we also test localstorage within the try block just to be safe, then?
> localStorage.setItem('localtest', 'localtest');
> localStorage.removeItem('localtest');

front-end/src/main.ts
```diff
@@ -39,9 +39,19 @@ import { AppComponent } from './app/app.component';
...
+  loginService: LoginService,
+  router: Router,
+  cookieCheckService: CookieCheckService,
+): () => Promise<void> {
   return async () => {
+    if (!cookieCheckService.areCookiesEnabled()) {
+      router.navigate(['/cookies-disabled']);
+      return;
+    }
+
```
danguyf, line 54: This is the bit I was going for in my first attempt but failed to arrive at!  I was trying to put it elsewhere; this makes so much more sense.

front-end/src/app/shared/services/cookie-check.service.ts
sasha-dresden: Hmm, I'm not quite sure how to test that. It seems to work perfectly fine as long as cookies are enabled.

danguyf:
So much cleaner and clearer!

There is some styling in the Design Notes to be incorporated, and displaying the logo (but not the logout) and the footer without the grey bar like the security consent page.  I'll give that a shot; if you beat me to it then it's still a good learning exercise for me.

sasha-dresden:
> So much cleaner and clearer!
> 
> There is some styling in the Design Notes to be incorporated, and displaying the logo (but not the logout) and the footer without the grey bar like the security consent page. I'll give that a shot; if you beat me to it then it's still a good learning exercise for me.

Oh that's what I get for just quickly checking the ticket description and seeing the first image and basing it off of that when that was the *old* version. Updated to match the new version presented below it.

front-end/src/app/layout/layout.component.html
```diff
@@ -35,7 +35,12 @@
 
   <!-- Begin No Sidebar section -->
   @else {
-    <div id="content-offset" #contentOffset>
+    <div
+      id="content-offset"
+      #contentOffset
+      [style.marginBottom]="isCookiesDisabled ? '-100px' : '0px'"
+      [style.marginTop]="isCookiesDisabled ? '100px' : '0px'"
+    >
```
danguyf, line 43: I don't love this but after fiddling for a little while I don't have a better solution.

danguyf (approved):

Elaine-Krauss-TCG (approved):

PR #2842, opened 2025-06-13
feature/2308 - Fix layout issue with login page and sidebar pages

sasha-dresden:
Issue [FECFIL-2308](https://fecgov.atlassian.net/browse/FECFILE-2308)

Every time I think I have it I find something else. So the reason I must've taken that column flex-direction off was because it broke the login page, but I eventually found the problem was because the internal element wasn't growing to fill when it was set to column, so the blue background portion wasn't stretching as it should have. I got that fixed. 
And as I was messing around with all the other pages to make sure nothing us was broken I noticed 2 other things.
1. The scrollbar fix wasn't happening at the right time. I needed to  move the variable into the _variables sheet where it belonged. And tell it to run after the view is checked. (This fix is mostly for the blue background of the login page so it adjust the width of the full width to account for the scrollbar which varies from browser to browser.)
2. The 64px minimum padding that's supposed to be on every page was missing on the pages with a sidebar now. That was easily fixed by setting the grow div to have a minimum height of 64px.

### DIFF
--- a/front-end/src/app/layout/layout.component.html
+++ b/front-end/src/app/layout/layout.component.html
@@ -24,7 +24,7 @@
           </div>
         </div>
         <div class="flex-grow-1 flex flex-column">
-          <div class="flex-grow-1" [style.minHeight]="'64px'"></div>
+          <div class="flex-grow-1 footer-spacer"></div>
           <app-footer [showSidebar]="true"></app-footer>
         </div>
       </div>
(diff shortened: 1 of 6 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/layout/layout.component.html
```diff
@@ -24,7 +24,7 @@
           </div>
         </div>
         <div class="flex-grow-1 flex flex-column">
-          <div class="flex-grow-1"></div>
+          <div class="flex-grow-1" [style.minHeight]="'64px'"></div>
```
danguyf: What are your thoughts on putting this in the scss file rather than as an inline style?

danguyf (approved):

PR #2843, opened 2025-06-13
feature/2315 - Make labels in parenthesis non-bold

sasha-dresden:
Issue [FECFILE-2315](https://fecgov.atlassian.net/browse/FECFILE-2315)

Most were rather straightforward. Only a couple that were not simply wrapping in the new paren-label class.
- PurposeDescriptionLabelSuffix, which was rather straightforward because all options were in parenthesis. 
- A couple classes which could be passed a label, which I resolved by adding a second input for a secondary label
- The memo text label which was the most complicated fix, but the logic for it is probably simpler than it once was.


danguyf (approved):

dheitzer (approved):
looks great

PR #2845, opened 2025-06-18
[Sprint 65] feature/1670 - Update F24 with name field

sasha-dresden:
Issue [FECFILE-1670](https://fecgov.atlassian.net/browse/FECFILE-1670)

API PR: [PR1526](https://github.com/fecgov/fecfile-web-api/pull/1526)


dheitzer (approved):
looks great

sasha-dresden:
Updated with some feedback from Todd.

- ‘Type’ changed to ‘Name’.
- Updated so Form24 subLabel uses the new name field.
  - This updates so the name will show in multiple different places, not just the report list.
- Update Add to Form 24 dialog, so you choose which Form 24 to add the transaction to based on the name of the Form24, instead of a generated label based on the create date.
- I also did some cleanup of the components to be more signal based.
  - More reactive and less usage of getters which cause constant calls on the render loop.

Elaine-Krauss-TCG (approved):

sasha-dresden:
Fixed an issue where if you try to save a Form 24 and it fails to process the save, it would close out of the dialog with no warning.

PR #2846, opened 2025-06-18
Feature/533 - Extends C2 Guarantor entity types to include Committees

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-533)

Related PRs: [App](https://github.com/fecgov/fecfile-web-app/pull/2846), [API](https://github.com/fecgov/fecfile-web-api/pull/1530)


front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
```diff
@@ -27,12 +27,21 @@
 
 @if (transaction?.contact_1) {
   @switch (entityType) {
-    @case (ContactTypes.ORGANIZATION || ContactTypes.COMMITTEE) {
```
Elaine-Krauss-TCG, line 30: Unfortunately, we can't use @case to match against multiple values at one time.  In this situation, the switch was comparing the value of `entityType` to `true` which is the result of the expression `(ORGANIZATION || COMMITTEE)`

danguyf (approved):
Code changes look good.

Updates tested successfully following steps in Description.

I think that the UI/UX for "Save" vs "Save and Add Another" has room for improvement but that is outside the scope of this task.

sasha-dresden (approved):

dheitzer (approved):

toddlees (approved):

PR #2850, opened 2025-06-24
FECFILE-2303: Updated angular to 19.2.15 with necessary dependencies …

danguyf:
…and possibly a few optional angular dependencies.

Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2303

Related PRs:
N/A


Elaine-Krauss-TCG (approved):

sasha-dresden (commented):
I feel like we should update all the angular dependencies to the latest 19 if we're updating so for example
@angular/platform-browser-dynamic, @angular/router, or @angular/animations

danguyf:
> I feel like we should update all the angular dependencies to the latest 19 if we're updating so for example @angular/platform-browser-dynamic, @angular/router, or @angular/animations

I had that same thought but ran into dependency issues when updating some Angular packages to the latest so eventually I backed out those few.  I could take another stab at it and see how far forward I can take those?

danguyf:
Okay, I found the solution and updated all of the angular packages now.

sasha-dresden (approved):

PR #2865, opened 2025-07-01
[Sprint 66] feature/2328 - Append datepicker to body to prevent overlay issue when datepicker is near bottom of the component

sasha-dresden:
Issue [FECFILE-2328](https://fecgov.atlassian.net/browse/FECFILE-2328)


danguyf (approved):
Reproduced original issue (-- though for me it only occurred when the field is in the top half of the screen; at the bottom it displayed above and was not clipped --) and then tested this fix successfully.

Elaine-Krauss-TCG (approved):

PR #2870, opened 2025-07-03
Feature/2333 - Allow Guarantor deletion

sasha-dresden:
Issue [FECFILE-2333](https://fecgov.atlassian.net/browse/FECFILE-2333)

Also addressed some e2e issues I noticed as I was writing the e2e test for this. I fixed an issue with some of the tests using format that would work under 8.5 but not 8.4

I also fixed an issue introduced by moving Independent Expenditure under the Contributions. There are still a few other issues but might want an actual ticket to look into those.


dheitzer (approved):

danguyf (approved):
Tested locally and successfully.

PR #2877, opened 2025-07-08
feature/2335 - Add scrollToTop function called when accordion selection changes

sasha-dresden:
Issue [FECFILE-2335](https://fecgov.atlassian.net/browse/FECFILE-2335)


toddlees (changes requested):
do we need to also do this for reattribution redesignation

danguyf (approved):
The test case given in the linked ticket works for me.

sasha-dresden:
> do we need to also do this for reattribution redesignation

The way I set this up it will run the scrollToTop function whenever the accordion value changes for any component that descends from the DoubleTransactionTypeBaseComponent which does include the ReattRedesTransactionTypeBaseComponent.

toddlees (approved):

PR #2883, opened 2025-07-10
feature/2242 - Update Save and Continue font color to black. Also update SaveAndCancel to use updated output function. Update Create Form 99 with several 508 compliance updates.

sasha-dresden:
Issue [FECFILE-2242](https://fecgov.atlassian.net/browse/FECFILE-2242)

### DIFF
--- a/front-end/src/app/shared/components/inputs/address-input/address-input.component.html
+++ b/front-end/src/app/shared/components/inputs/address-input/address-input.component.html
@@ -58,7 +58,7 @@
     </div>
     <div class="col-3">
       <div class="field">
-        <label for="state">STATE/PROVINCE</label>
+        <span id="state-label" class="span-label">STATE/PROVINCE</span>
         <p-select
           inputId="state"
           [formControlName]="stateFieldName"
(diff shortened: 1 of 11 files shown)


front-end/src/app/shared/components/inputs/address-input/address-input.component.html
```diff
@@ -58,7 +58,7 @@
     </div>
     <div class="col-3">
       <div class="field">
-        <label for="state">STATE/PROVINCE</label>
+        <span id="state-label" class="span-label">STATE/PROVINCE</span>
         <p-select
```
danguyf, line 62: The `p-select` which follows will now need an `ariaLabelledBy`, won't it?

sasha-dresden: No, so the wierd thing about this is that <p-select> is not treated as a regular html input. Since it doesn't make use of one of the defined html input-type elements input, select, textarea, etc. So because of that there's no concept of labelling it, which is why this was getting flagged as orphaned labels in the first place.

Elaine-Krauss-TCG (approved):

danguyf (approved):

PR #2926, opened 2025-08-04
feature/2239 - Cleanup some dead code in the header links. Add tabindex for the <li> elements to allow them to be tabbable.

sasha-dresden:
Issue [FECFILE-2239](https://fecgov.atlassian.net/browse/FECFILE-2239)


toddlees (changes requested):
I see you made changes to the other header links. I can tab into report and contacts but hitting enter doesn't navigate

dheitzer (approved):

toddlees (approved):

PR #2928, opened 2025-08-04
feature/2241 - Update Create Report Dialog with Custom header actions. Update Create F24 component labels.

sasha-dresden:
Issue [FECFILE-2241](https://fecgov.atlassian.net/browse/FECFILE-2241)

### DIFF
--- a/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
+++ b/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
@@ -9,9 +9,8 @@
           icon="pi pi-times"
           class="p-dialog-header-icon p-link"
           (click)="dialog.close($event)"
-        >
-          <span class="sr-only">Close</span>
-        </button>
+          aria-label="Close"
+        ></button>
       </div>
     </div>
   </ng-template>
--- a/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
+++ b/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
@@ -1,17 +1,7 @@
-<p-dialog #dialog [(visible)]="dialogVisible" [style]="{ width: '512px' }" [closable]="false" [draggable]="false">
+<p-dialog #dialog [(visible)]="dialogVisible" [style]="{ width: '512px' }" [draggable]="false" closeAriaLabel="Close">
   <ng-template #header>
-    <div class="custom-header" [id]="dialog.ariaLabelledBy">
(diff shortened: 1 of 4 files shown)


Elaine-Krauss-TCG (approved):

danguyf (approved):

front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
```diff
@@ -1,7 +1,21 @@
...
+      <h1>Create a new report</h1>
+      <div class="p-dialog-header-actions">
+        <button
+          pButton
+          type="button"
+          icon="pi pi-times"
+          class="p-dialog-header-icon p-link"
+          (click)="dialog.close($event)"
+        >
+          <span class="sr-only">Close</span>
```
dheitzer: Is there any reason we wouldn't want to use aria-label instead? seems like it is the preferred approach for buttons based on discussions I've seen of aria-label vs sr-only.

dheitzer (dismissed):

front-end/src/app/reports/form-type-dialog/form-type-dialog.component.html
sasha-dresden: This is frustrating. I could have _sworn_  I had tried aria-label and it didn't work. And I thought the issue was because the problem is that the button is considered empty. Here is the aria warning.

Errors
Empty button
What It Means
A button is empty or has no value text.
Why It Matters
When navigating to a button, descriptive text must be presented to screen reader users to indicate the function of the button.
How to Fix It
Place text content within the <button> element or give the <input> element a value attribute.

So it sounds like the issue is that there is absolutely no text inside the button. Icons apparently don't count. So I had put that sr-only text in there so it can be read by the screen reader. 

But I decided to double check aria-label and it worked this time. Maybe I didn't try, maybe I accidentally used aria-labelledby instead. Ugh so frustrating. But yes. aria-label worked.

sasha-dresden: I just looked at the comment on 2357 and decided to try closeAriaLabel here and it worked. Which is incredibly frustrating because closeAriaLabel did NOT work on the confirmationDialog.

toddlees: thank you for trying it again. That's really frustrating that it's not clear why it didn't work before, but i'm glad you found the simplest way to make it work

toddlees (approved):

PR #2931, opened 2025-08-06
feature/2392 - Update tables to use signals

sasha-dresden:
Issue [FECFILE-2392](https://fecgov.atlassian.net/browse/FECFILE-2392)


danguyf (approved):
This appears to be a significant reduction in complexity. I tested locally and did not run into any issues.

PR #2952, opened 2025-08-14
Feature/2330 - Create a printFormErrors utility function to print errors in the form to console

sasha-dresden:
Issue [FECFILE-2330](https://fecgov.atlassian.net/browse/FECFILE-2330)

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/triple-transaction-type-base.component.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/triple-transaction-type-base.component.ts
@@ -137,7 +137,7 @@ export abstract class TripleTransactionTypeBaseComponent
 
   override isInvalid(): boolean {
     blurActiveInput(this.childForm_2);
-    if (this.childForm_2.invalid) printFormErrors(this.form);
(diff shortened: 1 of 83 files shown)


front-end/src/app/shared/components/transaction-type-base/triple-transaction-type-base.component.ts
```diff
@@ -137,6 +137,7 @@ export abstract class TripleTransactionTypeBaseComponent
 
   override isInvalid(): boolean {
     blurActiveInput(this.childForm_2);
+    if (this.childForm_2.invalid) printFormErrors(this.form);
```
danguyf: This should be `printFormErrors(this.childForm_2)` I think?

sasha-dresden: Good catch! Fixed it there and an the double transaction as well.

danguyf (approved):

Elaine-Krauss-TCG (approved):

toddlees (approved):
works as expected: 
image

PR #2963, opened 2025-08-25
feature/2244 -Update ngrx-store-localstorage to latest 19 version

sasha-dresden:
Issue [FECFILE-2244](https://fecgov.atlassian.net/browse/FECFILE-2244)

Added overrides to prevent an issue with the dependency checker being a bit aggressive and flagging things that aren't issues.


toddlees (changes requested):
I ran circle on just changing the localstorage package and didn't see the issue with rollup.  I'm inclined to see that as a circle caching issue as opposed to something we need to change on our package.json.  Can you remove the overrides?
https://github.com/fecgov/fecfile-web-app/compare/test-npm?expand=1

sasha-dresden:
> I ran circle on just changing the localstorage package and didn't see the issue with rollup. I'm inclined to see that as a circle caching issue as opposed to something we need to change on our package.json. Can you remove the overrides? https://github.com/fecgov/fecfile-web-app/compare/test-npm?expand=1

Took the overrides out.

danguyf (approved):
The code works for me, and makes sense. I'm not sure about the failed dependency check, though.

toddlees (approved):

PR #2966, opened 2025-08-26
feautre/2423 - Fixes for e2e test failures

sasha-dresden:
Issue [FECFILE-2423](https://fecgov.atlassian.net/browse/FECFILE-2423)

There were a few different reasons for failures. 
First was fairly easy to resolve. When creating multiple Reports using the new direct API setup system, I had set up solid Promise completion when there was just a single report but not multiple. I updated the F3XSetup so it now has an option to pass in a list of Reports to create instead of only taking a single option. One is still returned as the working report.

Slightly more complicated was the issue with Individual loan testing. Some of those were failing when we had guarantors. At first I thought maybe the issue was related to the all the Toasts that popped up and covering the dropdown that needed to be clicked so I created a function to close out the toasts. That wasn't the issue but I left it in anyways since it might be useful. The issue seemed to be because we had three tables loading asynchronously and with the slight added delay from having guarantors that somehow threw off the timing just enough that the page refreshed content right as the Edit item kebab was trying to be selected. Fix was to add intercepts and wait for all three tables to load before trying to open the edit page. While working on this I noticed we were actually double calling all of those tables, so I did a fix to address that so it'd only fetch once on page load.


front-end/cypress/e2e/users.cy.ts
```diff
@@ -8,8 +8,9 @@ describe('Manage Users', () => {
   });
 
   it('should create a user', () => {
+    cy.intercept('GET', 'http://localhost:8080/api/v1/committee-members/').as('GetMembers');
     UsersPage.create(userFormData);
-    cy.wait(500);
+    cy.wait('@GetMembers');
```
toddlees, line 13: EXCELLENT

toddlees (changes requested):
Im not sure which change caused it, but now if i change number of rows to 5 on my reports page, clicking the different pages doesn't change the reports that show up
image

sasha-dresden:
You're right about the code changes. I wanted to address the double list API calls because it was leading to issues with the kebab button clicks getting interrupted when the data refreshed. I did find a workaround that did it with _only_ e2e code changes. Some of it is pretty nice, other parts, I had to double up on the cy.wait calls to make sure it waited on ALL the list calls to finish. I think we should make a ticket to address the doubled list API calls, but that should be the focus of a different ticket and some of these changes will require the e2e tests to be updated for when they only need to wait for a single API list call to be made. I tried to limit the number of places this doubled waits happen by moving more code into utility functions for repeatability.

toddlees (approved):

PR #2984, opened 2025-09-10
feature/2458 - Update so outer contact gets cleared when saved

sasha-dresden:
Issue [FECFILE-2458](https://fecgov.atlassian.net/browse/FECFILE-2458)

The issue was that the contact info wasn't being cleared on save. So when you left and returned it was re-saving it.


danguyf (approved):
Looks good. Local testing succeeded.

Elaine-Krauss-TCG (approved):

toddlees:
Accurate and effective fix.  However, we had to revert the merge that caused it.  We need to keep this in mind when we reimplement https://fecgov.atlassian.net/browse/FECFILE-2408

PR #2989, opened 2025-09-12
feature-1655 - Create Searchable Select Wrapper around PrimeNG select

sasha-dresden:
Issue [FECFILE-1655](https://fecgov.atlassian.net/browse/FECFILE-1655)


danguyf (approved):
Tested locally and working for me.  Repeatedly hitting a letter steps through the states starting with that letter.  Quickly typing the first few letters of a state (e.g. "M-I-S") jumps to the first state matching.  Slowly typing a string of letters jumps to the first state with each (e.g. Maine, Idaho, South Dakota).

dheitzer (dismissed):
Have we considered instead overriding [the p-select `searchOptions` function] (maybe in our `app-select` component)(https://github.com/primefaces/primeng/blob/ae09202fa9a046ad95c16ac39451fb39a47e5637/packages/primeng/src/select/select.ts#L1773C5-L1773C18) with the logic to handle repeated character inputs? I think something like this could help:

```
import { Select } from 'primeng/select';
export class SelectComponent extends Select {
  searchOptions(event: any, char: any): boolean {
    if (this.searchValue.split('').every(char => char === input[0]) {
      // set custom optionIndex 
      // call this.changeFocusedOptionIndex(event, optionIndex);
      // return true
  } else {
    super.searchOptions(event, char);
  }
}

```

front-end/src/app/contacts/contact-list/contact-list.component.ts
```diff
@@ -28,7 +30,9 @@ import { ContactService, DeletedContactService } from 'app/shared/services/conta
     ContactDialogComponent,
     DeletedContactDialogComponent,
     LabelPipe,
+    SelectModule,
   ],
+  providers: [SearchableSelectComponent],
```
toddlees: I don't believe the select is used in this component

toddlees (changes requested):
David's suggestion is enticing.  Any reason extending the component wouldn't let you reuse the code in the way he suggests?

toddlees:
Ultimately, It's functioning exactly as intended!  Even if we don't get the simplification @dheitzer suggested, this is looking really good

front-end/src/app/contacts/contact-list/contact-list.component.ts
sasha-dresden: That was my first thought and what I really wanted to do, to just extend it and override it that way. The main issue is that extending a component means you need to provide a new html template and I did _not_ want to do that. I suppose I could've copied and pasted the entire template html code from their codebase, but it's quite a bit and I was hoping to avoid being as fragile to changes in their code base as possible. Linking into just searchOptions was already a very strong tie but it was just a single point whereas overriding the entire html file meant we'd need to copy and paste it every time they make breaking updates.

toddlees (approved):

PR #3019, opened 2025-09-19
feature/2379 - Add loan agreement notification

sasha-dresden:
Issue [FECFILE-2379](https://fecgov.atlassian.net/browse/FECFILE-2379)


danguyf:
Tested locally and this is functionally complete.  My only quibble is that the space between the new notification and the existing text box needs to be greater.

image

sasha-dresden:
> Tested locally and this is functionally complete. My only quibble is that the space between the new notification and the existing text box needs to be greater.
> 
> image

Oh! Good catch! Got that added in so it matches the spacing that existed previously (1rem)
I also noticed the text wasn't centered in the grey box so I fixed that as well.

danguyf (approved):
Looks and works great.

toddlees (approved):
Works great

PR #3027, opened 2025-09-23
feature/2477 - Replace PrimeNG Textarea with custom AutoResize Directive

sasha-dresden:
Issue [FECFILE-2477](https://fecgov.atlassian.net/browse/FECFILE-2477)


danguyf (approved):
I confess that, after tracing through this off and on over two days I do not completely grok some of it, I'm making what I think are reasonable assumptions about what the changes are doing or why they were made.  I have tested it and it does what was requested.

The font size change within the textarea was a little jarring at first, though.  If that wasn't intentional then I expect it's an easy change.

dheitzer:
This does seem like a ton of work/changes just to make this not jump when it's resizing automatically.  From what I can tell, there's not an obvious alternative if we want the text area to resize automatically as they type.

Do we know if this would be acceptable to the user instead?
image

If we simply set autoResize to false, we get a scrollbar when they exceed the size of the box.  This prevents the page from getting longer as well.

toddlees (changes requested):
The font size will need to be updated.  But what I also noticed while testing is shelly's comment that this only happens on new reports.  I confirmed that the issue doesn't happen when the report is existing.  My best guess is that it has to do with the side bar.  Could this be a clue into a solution that doesn't require changing the directive?

sasha-dresden:
> The font size will need to be updated. But what I also noticed while testing is shelly's comment that this only happens on new reports. I confirmed that the issue doesn't happen when the report is existing. My best guess is that it has to do with the side bar. Could this be a clue into a solution that doesn't require changing the directive?

So I looked into this and it IS happening on the Edit screen as well, it's just much much less often. I was only able to get it to trigger when I was scrolling up from the bottom and only if I did it at a moderate speed. Very annoying. I also tried putting the sidebar on the Create screen and it caused it to behave just like the Edit screen with far fewer jumps. However I can't figure out why it would happen so much more often on one than the other. It doesn't make sense and unfortunately it's really hard to debug. When I run the debugger I can put it in a state where I know it will happen and in a state where I know it won't happen and both behave exactly the same until they return the height and one keeps it at the top and one returns to the previous scroll location with no clear understanding why. My guess is that some wierd css layout stuff is happening. 
There's also an issue that they have it happening on AfterViewChecked. One of the things I did to fix my version was only having it run the resize script when the input of the textarea changed, this prevented it from running constantly in the background. I think they run theirs constantly to allow for the case where the user manually resized the textarea, since that wouldn't trigger an input change. However it behaves very oddly when you try to do that, so I just turned off manual resizing.

Elaine-Krauss-TCG (approved):
Works like a charm!

toddlees (approved):
Styling fix addresses difference from original.  Given the simplicity of the primeng directive we're replacing, this isn't a big risk.  Even trying to chase down the behavior of the new report scenario (not edit scenario) is flawed because when it's new, the primeng component still jumps, just not as bad.

PR #3031, opened 2025-09-24
Feature/2478 - Update Edit Report for Form 3 and Form 3X to more accurately match existing data on load

sasha-dresden:
Issue [FECFILE-2478](https://fecgov.atlassian.net/browse/FECFILE-2478)

### DIFF
--- a/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
+++ b/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
@@ -112,12 +112,13 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
   private readonly isElectionYear = computed(() => F3ReportTypeCategories.ELECTION_YEAR === this.reportTypeCategory());
   readonly isElectionReport = computed(() => electionReportCodes.includes(this.reportCode()));
   private readonly coverages = computed(() => {
-    const coverageDatesFunction = getCoverageDatesFunction(this.reportCode());
-    if (!coverageDatesFunction) return undefined;
     const report = this.report();
+    const coverageDatesFunction = getCoverageDatesFunction(this.reportCode());
     if (this.form.pristine && report) {
       return [report.coverageDates!['coverage_from_date'], report.coverageDates!['coverage_through_date']];
     }
+
+    if (!coverageDatesFunction) return undefined;
     return coverageDatesFunction(this.thisYear, this.isElectionYear(), 'Q');
   });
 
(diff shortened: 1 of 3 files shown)


danguyf (approved):
Tested locally and successfully.

front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -114,6 +114,10 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
   private readonly coverages = computed(() => {
     const coverageDatesFunction = getCoverageDatesFunction(this.reportCode());
     if (!coverageDatesFunction) return undefined;
+    const report = this.report();
```
toddlees, line 115: I think we need to move this above the coverage dates check because in the scenario that your report has no coverage date funciton, we still want to pull the existing dates.  Ex.  make your report a termination report and set the dates.  when you go back to the dates page, it needs to show those dates

toddlees (changes requested):

front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
sasha-dresden: Oh good point! Got that fixed so it checks the report and if the form is pristine first before the coverageDatesFunction check

Elaine-Krauss-TCG (approved):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -124,7 +124,12 @@ export class CreateF3XStep1Component extends FormComponent implements OnInit {
 
   private readonly isElectionYear = computed(() => F3xReportTypeCategories.ELECTION_YEAR === this.reportTypeCategory());
```
toddlees, line 130: If I pick a non-election year code and come back, that toggle is not set to non-election.  Looks like its getting signaled the same way as something like filing frequency which is working

toddlees (changes requested):

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
sasha-dresden: Ah good catch, yea, I added the same fix I did for the coverage dates to here and it worked perfectly.

toddlees (approved):

PR #3037, opened 2025-09-25
Update memo code close action to use the dialog close and cleanup the dialog reference open status to be a signal for better two-way connections

sasha-dresden:
Issue [FECFILE-2515](https://fecgov.atlassian.net/browse/FECFILE-2515)


toddlees (approved):
works great

PR #3070, opened 2025-10-02
feature/2522 - Remove category code from conduit earmark receipt

sasha-dresden:
Issue [FECFILE-2522](https://fecgov.atlassian.net/browse/FECFILE-2522)


danguyf (approved):
Tested locally successfully.  Impressed that it's that simple.  (There's no code on the other end looking to save CATEGORY_CODE?)

PR #3073, opened 2025-10-02
Feature/2519 - Update MEMOs to all use memo_code = true and disabled/readonly. 

sasha-dresden:
Issue [FECFILE-2519](https://fecgov.atlassian.net/browse/FECFILE-2519)

### DIFF
--- a/front-end/src/app/shared/models/transaction-types/common-types/schb_memo.model.ts
+++ b/front-end/src/app/shared/models/transaction-types/common-types/schb_memo.model.ts
@@ -1,0 +1,7 @@
+import { SchBTransactionType } from '../../schb-transaction-type.model';
+
+export abstract class SchBMemo extends SchBTransactionType {
+  override mandatoryFormValues = {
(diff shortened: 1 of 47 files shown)


danguyf (approved):
Successfully tested locally.

If there is a way to do this in fewer edits I could not come up with it.

front-end/src/app/shared/models/transaction-types/common-types/schb_memo.model.ts
```diff
@@ -0,0 +1,7 @@
+import { SchBTransactionType } from '../../schb-transaction-type.model';
+
+export abstract class SchBMemo extends SchBTransactionType {
```
toddlees: Maybe It's worth being consistent with the other common type class names and calling this SCHEDULE_B_MEMO.  Not a blocker but I think it would make interpreting it's usage easier because it's like the other common types

toddlees (approved):

PR #3076, opened 2025-10-02
FECFILE-2533: refund to refunds

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2533

Related PRs:
N/A


sasha-dresden (approved):
Looks good. I was worried the e2e test might break but it doesn't look like we even have Disbursement Refunds in the e2e test path yet.

PR #3087, opened 2025-10-03
Feature/2500 - Default Election/Non-Election based on year/month

sasha-dresden:
Issue [FECFILE-2500](https://fecgov.atlassian.net/browse/FECFILE-2500)

As a note you can test this in the browser by running this script in your console:
```
const mockDate = new Date(2024, 1, 1); 
Date = class extends Date {
  constructor(dateString) {
    if (dateString) {
      super(dateString);
    } else {
      super(mockDate);
    }
  }
};
```

Just replace the new Date with whatever date you want to test. 
After running the script in your console, navigate to the Create Form 3X page and it will behave as if the current date is whatever you set your mockDate to.

### DIFF
--- a/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
+++ b/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
@@ -123,6 +123,10 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
   });
 
   public thisYear = new Date().getFullYear();
+  readonly defaultReportTypeCategory =
+    (this.thisYear % 2 === 0) !== (new Date().getMonth() === 0)
+      ? F3ReportTypeCategories.ELECTION_YEAR
+      : F3ReportTypeCategories.NON_ELECTION_YEAR;
   reportCodeLabelMap?: { [key in ReportCodes]: string };
 
   readonly reportId = injectParams('reportId');
@@ -134,7 +138,7 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
 
   constructor() {
     super();
-    this.form.patchValue({ form_type: F3FormTypes.F3N, report_type_category: this.reportTypeCategories[0] });
+    this.form.patchValue({ form_type: F3FormTypes.F3N, report_type_category: this.defaultReportTypeCategory });
     this.form.controls['coverage_from_date'].addValidators([Validators.required]);
     this.form.controls['coverage_through_date'].addValidators([
       Validators.required,
--- a/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
+++ b/front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
@@ -123,10 +123,7 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
   });
 
   public thisYear = new Date().getFullYear();
-  readonly defaultReportTypeCategory =
-    (this.thisYear % 2 === 0) !== (new Date().getMonth() === 0)
-      ? F3ReportTypeCategories.ELECTION_YEAR
-      : F3ReportTypeCategories.NON_ELECTION_YEAR;
+  readonly defaultReportTypeCategory = this.getDefaultTypeCategory();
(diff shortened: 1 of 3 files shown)


sasha-dresden:
Oh also, I did a bit of cleanup to the Create Form 3X component which is contained to the first commit. It mostly just added some comments for sections and rearranged code for easier reading.
The only real functional change is in the following commits and is the same as what's done for Create Form 3 where I didn't reorganize code.

danguyf (approved):
Local testing successful

front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -123,6 +123,10 @@ export class CreateF3Step1Component extends FormComponent implements OnInit {
   });
 
   public thisYear = new Date().getFullYear();
+  readonly defaultReportTypeCategory =
```
dheitzer: Does this ticket include updating Form 3 default coverage date logic as well? If so, I'm thinking the ticket should be updated to reflect this (especially for QA purposes).

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
```diff
@@ -74,11 +76,21 @@ export class CreateF3XStep1Component extends FormComponent implements OnInit {
     'state_of_election',
     'form_type',
   ];
-
   readonly form: FormGroup = this.fb.group(SchemaUtils.getFormGroupFieldsNoBlur(this.formProperties, f3xSchema), {
     updateOn: 'blur',
   });
+  readonly reportTypeCategories = [F3xReportTypeCategories.ELECTION_YEAR, F3xReportTypeCategories.NON_ELECTION_YEAR];
+  readonly defaultReportTypeCategory =
+    (this.year % 2 === 0) !== (new Date().getMonth() === 0)
```
dheitzer: Can we rewrite this conditional to be more verbose/explicit? It was tough for me to understand what it is trying to do without referencing the requirements.  I'm thinking a slight tweak would be more clear for me:

```
const currMonth = new Date().getMonth();
if (this.year % 2 === 0) {
  defaultReportTypeCategory = currMonth === 0 ?  F3xReportTypeCategories.NON_ELECTION_YEAR : F3xReportTypeCategories.ELECTION_YEAR;
} else {
  defaultReportTypeCategory = currMonth === 0 ?  F3xReportTypeCategories.ELECTION_YEAR: F3xReportTypeCategories.NON_ELECTION_YEAR ;
}
```

dheitzer (changes requested):
just a couple thoughts

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
sasha-dresden: Yea, I guess I was trying to be a little too clever there. Just kept realizing I could reduce the logic further and further till I got to that beautiful single line of conditions. haha
I went ahead and turned it into a function and made it much more explicit.

front-end/src/app/reports/f3/create-workflow/create-f3-step1.component.ts
sasha-dresden: Good point. We keep running into this situation. I talked with @toddlees about this and he's going to talk with @AureliaKhorsand about it. I believe in this situation both should match. 
What I'd really like to do is make an abstract class that sits above these to keep them in sync.

dheitzer: yeah that makes a lot of sense.  I thought the two should probably be in sync but didn't see anything explicitly mentioned in the ticket about Form 3.

front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.ts
dheitzer: As I was trying to write the code above, I kept doing the same thing ha.  It's an impressive little bit of code you wrote actually.

function looks nice - thanks!

dheitzer (approved):

sVmsepi0l:
test comment for exalate sync

PR #3090, opened 2025-10-06
feature/2536 - Move custom manage report code padding adjustment to theme so it will also apply to Manage Contact

sasha-dresden:
Issue [FECFILE-2536](https://fecgov.atlassian.net/browse/FECFILE-2536)

### DIFF
--- a/front-end/src/assets/styles/theme.css
+++ b/front-end/src/assets/styles/theme.css
@@ -2846,7 +2846,7 @@ p-treeselect.p-treeselect-clearable .p-treeselect-clear-icon {
   background: #fff;
   color: var(--p-content-color);
   border-bottom: 2px solid #000000;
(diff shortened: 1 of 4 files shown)


front-end/src/assets/styles/theme.css
```diff
@@ -2846,7 +2846,7 @@ p-treeselect.p-treeselect-clearable .p-treeselect-clear-icon {
   background: #fff;
   color: var(--p-content-color);
   border-bottom: 2px solid #000000;
-  padding: 48px 1rem 8px 0;
+  padding: 48px 0 8px 0;
```
dheitzer, line 2849: just a note that this removes the 16px margin for every data table (which I'm assuming is desired).

dheitzer (approved):

danguyf (approved):
Tested and working

PR #3102, opened 2025-10-08
feature/2537 - fix orphaned and missing label on Lookups Application wide

max-zaremba-tcg:
Ticket link: [FECFILE-2537](https://fecgov.atlassian.net/browse/FECFILE-2537)

Related PRs:


front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
```diff
@@ -25,6 +25,7 @@
     <div class="pl-0">
       <p-autoComplete
         id="searchBox"
+        [inputId]="`searchBox`"
```
sasha-dresden: You don't need to wrap it like that, you could just say inputId="searchBox"

sasha-dresden (changes requested):
This is good. It takes care of the look up on the transaction page. However you also need to take care of the one in the ContactLookupComponent. Just do the same fix there. For testing that page, you can do so by going to the /contacts and clicking Add Contact. Then switch to either committee or candidate.

Elaine-Krauss-TCG (approved):
I tested locally and tried to look for anything that might have been overlooked.  I looked at Earmark Receipt, Contribution to Candidate, Loan By Committee + Guarantor, and both versions of an F1M report.  Using the WAVE extension on Firefox, I found no issues related to the contact lookup fields.

sasha-dresden (approved):
Looks good! Very thorough in finding every single place!

toddlees (approved):
confirmed the labels are attached

PR #3105, opened 2025-10-08
feature/1997 - Merge Step 1 and 2 Submit Report

sasha-dresden:
Issue [FECFILE-1997](https://fecgov.atlassian.net/browse/FECFILE-1997)


danguyf (approved):
Code looks good, page looks good, spot checked CSS.

Elaine-Krauss-TCG (commented):
The code's great, and in manual testing it works like a charm, but E2E tests are broken (at least Amendment is).  This should be a quick fix, and I'm not against doing it in a separate ticket.

danguyf (changes requested):
The e2e tests need to be updated. As Elaine points out, there is at least one that uses the old `/step-x` URL schema.

sasha-dresden:
> The code's great, and in manual testing it works like a charm, but E2E tests are broken (at least Amendment is). This should be a quick fix, and I'm not against doing it in a separate ticket.

Yea I checked submit but not amendment. It was hard to figure out since the e2e tests were broken because of a different issue. Looks like amendment is the only area it was using step. However I did a search and found a few more routes using step that I missed so I fixed those as well.

Elaine-Krauss-TCG (approved):

danguyf (approved):

danguyf (approved):

PR #3109, opened 2025-10-09
feature/2566 - Change STATE/PROVINCE to STATE/TERRITORY

sasha-dresden:
Issue [FECFILE-2566](https://fecgov.atlassian.net/browse/FECFILE-2566)


danguyf (approved):

front-end/src/app/shared/models/contact.model.ts
```diff
@@ -53,7 +53,7 @@ export enum ContactFields {
   street_1 = 'Street Address',
   street_2 = 'Apartment, Suite, Etc.',
   city = 'City',
-  state = 'State/Province',
+  state = 'State/TERRITORY',
```
dheitzer: Super minor but I'm wondering if this should be camelcase to match the rest of the fields.  It seems we only display this in the update audit dialog (which is lowercased anyway), but I thought for consistency here.

dheitzer (approved):
Minor optional change suggested.

front-end/src/app/shared/models/contact.model.ts
sasha-dresden: Good catch. Fixed.

PR #3124, opened 2025-10-14
feature/2435 - Deduplicate table requests

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2435

Related PRs:


front-end/src/app/contacts/contact-list/contact-list.component.html
```diff
@@ -67,4 +67,6 @@ <h1 class="m-0">Manage contacts</h1>
   (savedContact)="saveContact($event)"
 />
 
-<app-deleted-contact-dialog [(visible)]="restoreDialogIsVisible" (contactsRestored)="refreshTable()" />
+@if (restoreDialogIsVisible) {
```
dheitzer, line 70: deleted-contact-dialog was making a request for deleted contact count even before it was opened (this was in addition to a deleted contact count request for the contact table)..

front-end/src/app/contacts/contact-list/contact-list.component.ts
```diff
@@ -66,10 +66,6 @@ export class ContactListComponent extends TableListBaseComponent<Contact> implem
     { field: 'occupation', label: 'Occupation' },
   ];
 
-  ngOnInit() {
-    this.checkForDeletedContacts();
```
dheitzer, line 70: This is not needed as it is already being called in loadTableItems for contact-list.

front-end/src/app/contacts/deleted-contact-dialog/deleted-contact-dialog.component.ts
```diff
@@ -35,7 +34,6 @@ export class DeletedContactDialogComponent extends TableListBaseComponent<Contac
       this.selectedItems.set([]);
       if (this.visible()) {
         this.first.set(0);
-        this.loadTableItems({ first: 0, rows: this.rowsPerPage() });
```
dheitzer, line 38: This was an extraneous call

front-end/src/app/shared/components/table/table.component.html
```diff
@@ -12,6 +12,7 @@
     [currentPageReportTemplate]="currentPageReportTemplate()"
     [sortField]="sortField()"
     [lazy]="true"
+    [lazyLoadOnInit]="false"
```
dheitzer, line 15: this defaults to "true" which was causing extra loadTableData calls/requests.  We are already calling loadTableItems in our table-list-base component constructor.

sasha-dresden (dismissed):
This looks great! I really like your solution to essentially just turn off that init lazy load and using the distinctUntilChanged.
Works perfect in usage, however, all the e2e tests are going to be broken until we fix them up to take out the waiting for multiple responses. For example look at the ReportListPage class and it has a few places where it's waiting for multiple responses.

dheitzer:
> This looks great! I really like your solution to essentially just turn off that init lazy load and using the distinctUntilChanged. Works perfect in usage, however, all the e2e tests are going to be broken until we fix them up to take out the waiting for multiple responses. For example look at the ReportListPage class and it has a few places where it's waiting for multiple responses.

Ohh, great catch! I completely forgot about updating those.

dheitzer:
> This looks great! I really like your solution to essentially just turn off that init lazy load and using the distinctUntilChanged. Works perfect in usage, however, all the e2e tests are going to be broken until we fix them up to take out the waiting for multiple responses. For example look at the ReportListPage class and it has a few places where it's waiting for multiple responses.

fixed

danguyf (approved):
I dig the explanatory comments in the PR, that's very helpful.

I'm excited that this change lets us remove all the duplicate waits.

I love that the solution is so direct.

toddlees (approved):
works beautifully

PR #3125, opened 2025-10-15
feature/2512 - Allow user to click "+" in "+ Create new contact" to create new contact & fix tabbing

max-zaremba-tcg:
Ticket link: [FECFILE-2512](https://fecgov.atlassian.net/browse/FECFILE-2512)

Related PRs:


front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
```diff
@@ -62,9 +62,23 @@
     <div class="add-contact-container">
       or
       <span [style.verticalAlign]="'middle'"
-        ><img alt="Plus" src="assets/img/new_contact_blue_plus.svg" height="20px" width="20px"
+        ><img
+          tabindex="0"
```
sasha-dresden: So, instead of putting the click and tabindex on both the image and the link, if you move everything up to the span like this:
```
<span
        class="flex align-items-center gap-1"
        tabindex="0"
        (click)="onCreateNewContactSelect()"
        (keydown.enter)="onCreateNewContactSelect()"
        (keydown.space)="onCreateNewContactSelect()"
      >
        <img alt="Plus" src="assets/img/i-plus-circle.svg" height="20px" width="20px" />
        <a>Create a new contact</a>
      </span>
```
That should prevent a bit of duplication and give them same effect. Also when you're tabbing through it will treat it as a single tab instead of two separate entries to tab through.

front-end/src/app/shared/components/contact-lookup/contact-lookup.component.scss
```diff
@@ -75,16 +75,17 @@
...
 }
 
 .add-contact-container span {
   padding: 0 0.3em;
+  cursor: pointer;
 }
 
 .add-contact-container a {
-  border-bottom: none;
+  border-bottom: 1px dotted;
```
sasha-dresden: You can just remove this and the default <a> underlining effect should take place.

sasha-dresden (changes requested):
Looks pretty good. I've requested a couple changes to make things a little cleaner and a teensy bit more keyboard friendly.

front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
sasha-dresden: Oh if sonar complains about an anchor with nothing on it (I never know when sonar is going to complain half the time but this seems like something it might complain about) you could replace the anchor with a span styled to look the same:
`<span class="border-none border-bottom-1 border-dotted p-0">Create a new contact</span>`

sasha-dresden (approved):

danguyf (approved):

PR #3138, opened 2025-10-17
feature/2534 - Update Committee Name found on most pages in the app

max-zaremba-tcg:
Ticket link: [FECFILE-2534](https://fecgov.atlassian.net/browse/FECFILE-2534?search_id=5201bdc1-06db-4cbf-9afe-5c84d6f344fb)

Related PRs:

### DIFF
--- a/front-end/src/app/layout/committee-banner/committee-banner.component.html
+++ b/front-end/src/app/layout/committee-banner/committee-banner.component.html
@@ -1,21 +1,17 @@
 @if (committeeID()) {
-  <div class="grid">
-    <div class="col-12 pt-0">
-      <div class="committee-name">{{ committeeName() }}</div>
-      <div class="grid">
-        <div class="col-12 committee-info" aria-label="Committee Status">
-          <div
-            class="pi pi-circle-on status-circle"
-            [style.color]="frequencyColor()"
-            [attr.aria-label]="committeeStatus()"
-          ></div>
-          <div class="sub-banner-item">{{ committeeFrequency() }}</div>
-          |
-          <div class="sub-banner-item">{{ committeeTypeLabel() }}</div>
-          |
-          <div class="sub-banner-item">ID: {{ committeeID() }}</div>
-        </div>
-      </div>
+  <div class="pt-0">
+    <div class="committee-name">{{ committeeName() }}</div>
+    <div class="committee-info" aria-label="Committee Status">
+      <div
+        class="pi pi-circle-on status-circle"
+        [style.color]="frequencyColor()"
+        [attr.aria-label]="committeeStatus()"
+      ></div>
+      <div class="sub-banner-item">{{ committeeFrequency() }}</div>
+      |
+      <div class="sub-banner-item">{{ committeeTypeLabel() }}</div>
+      |
+      <div class="sub-banner-item">ID: {{ committeeID() }}</div>
     </div>
   </div>
 }
--- a/front-end/src/app/layout/committee-banner/committee-banner.component.scss
+++ b/front-end/src/app/layout/committee-banner/committee-banner.component.scss
@@ -10,7 +10,6 @@
   line-height: 1.2;
   font-size: 36px;
   text-align: left;
-  margin-left: -8px;
   font-weight: 700;
 }
 
@@ -22,7 +21,6 @@
 .status-circle {
   font-size: 16px;
   margin-right: 5px;
-  margin-left: -16px;
   vertical-align: middle;
 }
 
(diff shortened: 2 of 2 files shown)


danguyf (approved):
This results in the requested change everywhere that I tested.

This is probably the most succinct change to accomplish this, but I have an aversion to negative margins when an alternative exists.  I would be happier with a solution that removed the left margin from the container instead; it seems safer and more semantically pure.

Approving this despite those reservations.

dheitzer (dismissed):
I always try to avoid using negative margin/padding because it's usually a smell that indicates an offset added somewhere else that is causing the misalignment.  In this case, I found that padding: .5rem added in the col-12 class [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L3) and [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L6) were causing the offset.  Removing these classes seemed to solve the issue.  Happy to tag up if you want to discuss further.

image

danguyf:
> I always try to avoid using negative margin/padding because it's usually a smell that indicates an offset added somewhere else that is causing the misalignment. In this case, I found that padding: .5rerm added in this class [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L3) and [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L6) were causing the offset. Removing these classes seemed to solve the issue. Happy to tag up if you want to discuss further.
> 
> image

My thought was that class col-12 might be used elsewhere, now or in the future, but that class committee-info could be changed and a new class could be added and styled in place of modifying col-12.

dheitzer:
> > I always try to avoid using negative margin/padding because it's usually a smell that indicates an offset added somewhere else that is causing the misalignment. In this case, I found that padding: .5rerm added in this class [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L3) and [here](https://github.com/fecgov/fecfile-web-app/blob/cdd63289920f7d94e234781f214901cd66a04c71/front-end/src/app/layout/committee-banner/committee-banner.component.html#L6) were causing the offset. Removing these classes seemed to solve the issue. Happy to tag up if you want to discuss further.
> > image
> 
> My thought was that class col-12 might be used elsewhere, now or in the future, but that class committee-info could be changed and a new class could be added and styled in place of modifying col-12.

Well, col-12 class is actually a bootstrap grid class so if we simply remove it in those two places the alignment seems to look correct to me after.

danguyf:
> > My thought was that class col-12 might be used elsewhere, now or in the future, but that class committee-info could be changed and a new class could be added and styled in place of modifying col-12.
> 
> Well, col-12 class is actually a bootstrap grid class so if we simply remove it in those two places the alignment seems to look correct to me after.

Yeah, I hate to modify the default bootstrap classes just in case someone else comes in expecting them to work normally.

sasha-dresden:
> > > My thought was that class col-12 might be used elsewhere, now or in the future, but that class committee-info could be changed and a new class could be added and styled in place of modifying col-12.
> > 
> > 
> > Well, col-12 class is actually a bootstrap grid class so if we simply remove it in those two places the alignment seems to look correct to me after.
> 
> Yeah, I hate to modify the default bootstrap classes just in case someone else comes in expecting them to work normally.

Yea don't modify the col-12. In this case, we don't even need the col-12 or the grids, just drop them so you'll end up with something like this:
```
<div class="pt-0">
    <div class="committee-name">{{ committeeName() }}</div>
    <div class="committee-info" aria-label="Committee Status">
      <div
        class="pi pi-circle-on status-circle"
        [style.color]="frequencyColor()"
        [attr.aria-label]="committeeStatus()"
      ></div>
      <div class="sub-banner-item">{{ committeeFrequency() }}</div>
      |
      <div class="sub-banner-item">{{ committeeTypeLabel() }}</div>
      |
      <div class="sub-banner-item">ID: {{ committeeID() }}</div>
    </div>
  </div>
  ```

danguyf:
Moved back on In Progress, @max-zaremba-tcg .

danguyf (approved):

sasha-dresden (approved):

PR #3140, opened 2025-10-17
Feature/2402 - Cleanup Manage Committee table columns

sasha-dresden:
Issue [FECFILE-2402](https://fecgov.atlassian.net/browse/FECFILE-2402)

API[PR1750](https://github.com/fecgov/fecfile-web-api/pull/1750)

This also cleaned up some of the table templates to be a bit simpler and less repetitive. 
Unfortunately until all tables are converted over to this format we will have a bit of lingering code but I marked what should be removed when the final table is converted over to using Column Definitions instead of splitting some headers in the table component and some in the template.


danguyf (changes requested):
I love the clean-up and refactoring.

ACs working in my testing:
* table widths
* column widths
* column visibility
* putting " (You)" after the first user

I am not seeing the requested changes to:
* table and header padding
* header font size
* header font weight
* bolding the first user
* bottom pagination padding

sasha-dresden:
Oh good catch @danguyf ! This was actually my 3rd take on this ticket and it looks like I fixed it in the previous version but totally forgot in this one. Simply copied fixes over.

danguyf (approved):

Elaine-Krauss-TCG (approved):

PR #3141, opened 2025-10-17
FECFILE-2569: remove efiling password validation

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2569

Related PRs:
N/A

### DIFF
--- a/front-end/cypress/README.md
+++ b/front-end/cypress/README.md
@@ -35,7 +35,7 @@ export CELERY_WORKER_STORAGE="local"
 
 With the environment variables set, run `ng e2e` (or `npx -p @angular/cli ng e2e`) to execute the end-to-end tests via [Cypress](https://www.cypress.io/).
 
-To run in headless mode, run the command: ng e2e --headless
+To run in headless mode, run the command: `ng e2e --headless`
 
 ## E2E tests in CircleCI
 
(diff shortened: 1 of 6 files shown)


dheitzer (dismissed):
@danguyf I think we still want validation on this field.  I think the idea for this ticket was simply not to trigger the validation error when they click the 'eye' icon to view the obfuscated password.

danguyf:
> @danguyf I think we still want validation on this field. I think the idea for this ticket was simply not to trigger the validation error when they click the 'eye' icon to view the obfuscated password.

@dheitzer While your interpretation sounds entirely sensible, I checked with Aurelia and Todd and confirmed that we do, in fact, want to remove validation on this field.  The thinking is that the user is not create an efiling password here, that was done on FEC, all the user is doing here is typing in that existing password, therefore we should not be attempting to validate it.

dheitzer:
> > @danguyf I think we still want validation on this field. I think the idea for this ticket was simply not to trigger the validation error when they click the 'eye' icon to view the obfuscated password.
> 
> @dheitzer While your interpretation sounds entirely sensible, I checked with Aurelia and Todd and confirmed that we do, in fact, want to remove validation on this field. The thinking is that the user is not create an efiling password here, that was done on FEC, all the user is doing here is typing in that existing password, therefore we should not be attempting to validate it.

That makes sense @danguyf thanks for circling back.  I'll CR this one accordingly then.

toddlees (changes requested):
We can now remove the passwordValidator in the validator.utils

Testing works great!

danguyf:
> We can now remove the passwordValidator in the validator.utils

Done!

toddlees (approved):

PR #3142, opened 2025-10-17
Feature/729 - Form should jump to error on failed validation.

sasha-dresden:
Issue [FECFILE-729](https://fecgov.atlassian.net/browse/FECFILE-729)


danguyf (approved):
This works in my testing.

danguyf:
Pinging @sasha-dresden that a merge conflict has cropped up since yesterday. Looks trivial!

sasha-dresden:
> Pinging @sasha-dresden that a merge conflict has cropped up since yesterday. Looks trivial!

Oh thanks for the alert. Got it addressed.

front-end/src/app/shared/components/destroyer.component.ts
Elaine-Krauss-TCG, line 1: 🎉 🎉 🎉

Elaine-Krauss-TCG (dismissed):
The code is beautiful, and it runs super well locally.  I did run through all the E2E tests, and the Users test is newly broken by this PR.  It passes on develop and failed twice in a row in this branch because it can't find the submit button.

Other than that, I wholeheartedly approve! 🎉

sasha-dresden:
> The code is beautiful, and it runs super well locally. I did run through all the E2E tests, and the Users test is newly broken by this PR. It passes on develop and failed twice in a row in this branch because it can't find the submit button.
> 
> Other than that, I wholeheartedly approve! 🎉

Oh thank goodness the e2e tests are working again! Looks like the issue was due to the fact we weren't fully validating the form for editing users (because email wasn't changing) but essentially we were always submitting that with an invalid form (I wonder if it would be better to use two separate forms here) so I fixed it up so it updates the email validation based on whether it's a new user (requires unique email) or an edited user.

front-end/src/app/reports/f1m/main-form/main-form.component.ts
```diff
@@ -248,18 +247,10 @@ export class MainFormComponent extends MainFormBaseComponent implements OnInit,
...
-      printFormErrors(this.form);
-      this.store.dispatch(singleClickEnableAction());
-      return;
-    }
-
+  public override async submit(jump: 'continue' | void): Promise<void> {
     const confirmed = await this.getConfirmations();
     // if every confirmation was accepted
-    if (confirmed) super.save(jump);
+    if (confirmed) super.submit(jump);
```
toddlees, line 251: ~~It seems like since we didn't use the `submitForm` method, `validateForm` doesn't get run~~
Upon looking closer it actually seems like its not running the validateForm from the form component because it's overridden in MainFormBaseComponent

toddlees (changes requested):

front-end/src/app/reports/f1m/main-form/main-form.component.ts
sasha-dresden: After we talked about it and looking into it I was able to get rid of the MainForms override of the validateForm function. Only place I override it now is for accordion transactions so it can deal with the transition.
Also updated the FormControl's validation to use that more accurate form statuses and then also after the next render to make sure the DOM is updated with the form errors. I think I got rid of all the setTimeouts this way.

toddlees (approved):

PR #3154, opened 2025-10-21
Revert BDD AND E2E fixes

toddlees:
In an effort to get a clean slate work bdd into the pipeline with less things in flux, we're removing it from the sprint.

This also fixes a number of e2e that broke but were not caught during the e2e outage


front-end/.gitignore
```diff
@@ -38,7 +38,3 @@ testem.log
 # System Files
 .DS_Store
 Thumbs.db
-
-# Cypress files
```
sasha-dresden, line 42: I don't think we should remove this from gitignore. I ran my e2e test locally and now I have a ton of files that it wants to commit

front-end/cypress/e2e/F3X/aggregate-calculation.cy.ts
```diff
@@ -82,7 +82,7 @@ describe('Tests transaction form aggregate calculation', () => {
 
       // Tests changing the second transaction's contact
       cy.get('[id=aggregate]').should('have.value', '$25.00');
-      cy.get('[id="searchBox"]').type('A');
+      cy.get('[id="searchBox"]').first().type('A');
```
sasha-dresden: This does fix the e2e test, but we should probably address it at the root as well. 
ContactLookup and ContactSearch have a <p-autoComplete> which uses searchBox both as the id and the inputId. It should _just_ be the inputId.

front-end/.gitignore
toddlees: good call!  it was part of a revert, but let's keep it

front-end/cypress/e2e/F3X/aggregate-calculation.cy.ts
toddlees: I'll make the edit

PR #3159, opened 2025-10-21
feature/2589 - Update formatting of the name of the Form 24 during creation.

sasha-dresden:
Issue [FECFILE-2589](https://fecgov.atlassian.net/browse/FECFILE-2589)

### DIFF
--- a/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.scss
+++ b/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.scss
@@ -2,3 +2,33 @@
   margin-top: 32px;
   width: fit-content;
 }
+
+#f24-input-container {
+  border: 2px solid var(--fec-grey-alt);
+  display: flex;
+  border-radius: var(--border-radius);
+}
+#f24-input-container:focus-within {
+  outline: 0 none;
+  outline-offset: 0;
+  box-shadow: var(--focus-ring);
+  border-color: var(--p-primary-color);
+}
+
+#f24-input-container:hover {
+  border-color: var(--surface-border);
+}
+
+#f24-input-container > span {
+  border: 0;
+  background: var(--p-content-background);
+  align-content: center;
+  font-weight: bold;
+  padding: 0 0 0 15px;
+  text-wrap-mode: nowrap;
+}
+
+#form24Name {
+  border: 0;
+  box-shadow: none;
+}
--- a/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.scss
+++ b/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.scss
@@ -5,27 +5,29 @@
 
 #f24-input-container {
   border: 2px solid var(--fec-grey-alt);
-  display: flex;
   border-radius: var(--border-radius);
-}
-#f24-input-container:focus-within {
-  outline: 0 none;
-  outline-offset: 0;
-  box-shadow: var(--focus-ring);
-  border-color: var(--p-primary-color);
-}
+  display: flex;
+  align-items: baseline;
 
-#f24-input-container:hover {
-  border-color: var(--surface-border);
-}
+  &:focus-within {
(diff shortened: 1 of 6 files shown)


danguyf (changes requested):
This ticket has quite the tangle of ACs.  A few comments (and at least one question) based on my interpretation of them.

* When entering text into the Report Name field the text baseline does not line up between the 24/48 prefix and the text entry.
* When I submit a report it is saved with the Report Name that I entered only -- missing the "24-Hour:"/"48-Hour:" prefix.  I believe this is a bug.
* The Report Name is validated for uniqueness as it is entered whereas the ACs say that it should only be validated once the "Start building report" button is checked. I prefer it the way you have implemented but I may be in the minority there.
* According to the QA Notes (but not spelled out as clearly in the ACs), the label "Edit Name" should be "Report Name" when the report is being created.

sasha-dresden:
> This ticket has quite the tangle of ACs. A few comments (and at least one question) based on my interpretation of them.
> 
> * When entering text into the Report Name field the text baseline does not line up between the 24/48 prefix and the text entry.
> * When I submit a report it is saved with the Report Name that I entered only -- missing the "24-Hour:"/"48-Hour:" prefix.  I believe this is a bug.
> * The Report Name is validated for uniqueness as it is entered whereas the ACs say that it should only be validated once the "Start building report" button is checked. I prefer it the way you have implemented but I may be in the minority there.
> * According to the QA Notes (but not spelled out as clearly in the ACs), the label "Edit Name" should be "Report Name" when the report is being created.

Ah, you're correct. I misinterpreted the AC. I thought the "24-Hour:" section was purely decorative, but re-reading the AC's you are definitely correct. I updated it so that gets included both for the validation check and for what gets saved.

As for the validation, I didn't touch the validation logic (well just to modify the name being compared). The code doing the check was already happening from the first request. So I imagine that's the way it should be. I'll verify, but I expect this method to be what stays.

danguyf (changes requested):
This issue with report title prefix is fixed and my question about the validation has been addressed.

That leave the following two quibbles:
* When entering text into the Report Name field the text baseline does not line up between the 24/48 prefix and the text entry.  This is because the prefix is karla but the input is the `p.inputtext` default; if you change one or the other this is addressed.
* According to the Design and QA Notes sections, the label "Edit Name" should be changed to "Report Name".

front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.scss
```diff
@@ -2,3 +2,33 @@
...
+
+#f24-input-container > span {
+  border: 0;
+  background: var(--p-content-background);
+  align-content: center;
+  font-weight: bold;
+  padding: 0 0 0 15px;
+  text-wrap-mode: nowrap;
+}
+
```
danguyf, line 32: I think adding something like `#f24-input-container .p-inputtext { font-family: var(--karla); }` would address the text baseline mismatch.

sasha-dresden: Oh that's a really good catch because the default text for the input was `-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'` which is definitely wrong, never should be that. I updated it to be Karla which is our sans-serif font. The baseline issue, man you have a good idea to notice that. I had centered the span, rather than baseline so I updated to use baseline and I also noticed the line height was different so I fixed that as well.

sasha-dresden:
> * According to the Design and QA Notes sections, the label "Edit Name" should be changed to "Report Name".

I'm losing my mind. That was like the _very_ first thing I changed. I guess I must've accidentally undone it when I switched to using an input group. Got it fixed now.

danguyf (approved):

Elaine-Krauss-TCG (approved):

PR #3162, opened 2025-10-22
Fecfile-2598 and 2584 (e2e-nightly test suite and reusable test-api)

sVmsepi0l:
Ticket link: 
https://fecgov.atlassian.net/browse/FECFILE-2584  (test-api command)
https://fecgov.atlassian.net/browse/FECFILE-2598  (new cypress/nightly suite)
https://fecgov.atlassian.net/browse/FECFILE-2453  (re-worked users.manage.feature to *.cy.ts)

Related PRs:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,6 +249,10 @@ parameters:
     type: boolean
     default: false
 
+  is-nightly-e2e-test:
+    type: boolean
+    default: false
+
 workflows:
   primary:
     when:
@@ -280,6 +284,11 @@ workflows:
     when: << pipeline.parameters.is-triggered-e2e-test >>
     jobs:
       - e2e-test
+
+  triggered-nightly-e2e-test:
+    when: << pipeline.parameters.is-nightly-e2e-test >>
+    jobs:
+      - e2e-nightly
 
   nightly:
     triggers:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ orbs:
 
 # Reusable Commands see https://circleci.com/docs/reference/reusing-config/
 commands:
-  test-api:
+  spin-up-api-containers:
     description: "Setup and start fecfile-web-api for E2E tests"
     steps:
       - setup_remote_docker:
@@ -146,7 +146,7 @@ jobs:
     docker:
       - image: cimg/node:lts-browsers
     steps:
-      - test-api
+      - spin-up-api-containers
       - run:
           name: execute e2e tests
           command: |
@@ -174,7 +174,7 @@ jobs:
     docker:
       - image: cimg/node:lts-browsers
     steps:
-      - test-api
+      - spin-up-api-containers
       - run:
           name: execute nightly e2e tests
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,48 @@ orbs:
   node: circleci/node@7.1.1
   python: circleci/python@3.2.0
 
+# Reusable Commands see https://circleci.com/docs/reference/reusing-config/
+commands:
+  test-api:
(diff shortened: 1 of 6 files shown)


danguyf (approved):
Slick!  I love the re-usable command.

sVmsepi0l:
> Slick! I love the re-usable command.

Thanks, glad you like it!

front-end/cypress.config.ts
```diff
@@ -2,7 +2,7 @@ import { defineConfig } from 'cypress';
 
 export default defineConfig({
   defaultCommandTimeout: 10000,
-  projectId: 'x5egpz',
+  projectId: '6um8hh',
```
toddlees: I don't think we should be changing this.

.circleci/config.yml
```diff
@@ -8,6 +8,48 @@ orbs:
   node: circleci/node@7.1.1
   python: circleci/python@3.2.0
 
+# Reusable Commands see https://circleci.com/docs/reference/reusing-config/
+commands:
+  test-api:
```
Elaine-Krauss-TCG: I would recommend updating this name to be more expressive, maybe like "spin-up-api-containers" or "setup-containers-for-testing" or something like that.

front-end/cypress.config.ts
```diff
@@ -26,5 +26,9 @@ export default defineConfig({
   },
   e2e: {
     baseUrl: 'http://localhost:4200',
+    specPattern: ['cypress/e2e/**/*.cy.ts', 'cypress/nightly/**/*.cy.ts'],
+    experimentalRunAllSpecs: true,
```
toddlees: we're not using any of these experiemental features as far as i can tell.  that prompt feature looks insane though!  Looks like it only works on cloud.

toddlees (changes requested):

front-end/cypress/nightly/users/users.manage.cy.ts
```diff
@@ -0,0 +1,147 @@
...
+      cy
+        .get('body')
+        .find(
+          "[data-cy='email-error'], [aria-live='assertive'], [role='alert'], .p-message-error, .p-error, .p-inline-message-error"
+        )
+        .first();
+
+    UsersPage.goToPage();
+    PageUtils.clickButton('Add user');
+    cy.wait(150);
```
Elaine-Krauss-TCG: Waiting with fixed timing has caused a lot of race conditions in our E2E tests in the past.  They run great locally, but in the slow, slow CircleCI environment, they'll sometimes fail and sometimes pass.  This will need to be replaced with waiting for some element of the Add User pop-up to be visible or with some other deterministic solution.

Elaine-Krauss-TCG (changes requested):
This looks great, and this has great code coverage for the tests!  I have a small suggestion for the CircleCI config, and there's one very quick, very simple change that needs to be made in the E2E test itself.  Otherwise, this is great stuff 🎉

sVmsepi0l:
> This looks great, and this has great code coverage for the tests! I have a small suggestion for the CircleCI config, and there's one very quick, very simple change that needs to be made in the E2E test itself. Otherwise, this is great stuff 🎉

Awesome, thanks for the feedback! I'll make those changes real quick and push them

dheitzer (changes requested):
looks like there's a few sonarcloud low issues that might be easy to address.

sVmsepi0l:
> looks like there's a few sonarcloud low issues that might be easy to address.

On it, good catch

Elaine-Krauss-TCG (approved):
Tested it now that the wait(150) command has been removed, and it works just fine~

dheitzer (approved):

toddlees (approved):
Awesome stuff Joshua, this level of detail is going to really help us

PR #3185, opened 2025-10-27
feature/2544 - 24px gutter on the left and right sides of the Left Nav

max-zaremba-tcg:
Ticket link: [FECFILE-2544](https://fecgov.atlassian.net/browse/FECFILE-2544)

Related Tickets:
[FECFILE-2543](https://fecgov.atlassian.net/browse/FECFILE-2543)
[FECFILE-2559](https://fecgov.atlassian.net/browse/FECFILE-2559)

### DIFF
--- a/front-end/src/app/layout/sidebar/menus/f1m/f1m-menu.component.html
+++ b/front-end/src/app/layout/sidebar/menus/f1m/f1m-menu.component.html
@@ -8,7 +8,7 @@
       </div>
     </div>
   </div>
-  <div class="pl-1 pt-2">
+  <div>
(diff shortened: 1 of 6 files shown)


sasha-dresden (changes requested):
This looks good, however I noticed there was one other spot with wierd margin coming into play. If you look at the `<div class="p-panelmenu-panel">`

there's a margin applied which is adding some extra margins on the left and right which isn't desired. You can find where it's happening in the theme.css file
```
.p-panelmenu .p-panelmenu-panel {
  margin: 0 0.5rem 0 0.25rem;
}
```

sasha-dresden (approved):

PR #3186, opened 2025-10-27
Feature/1933 - Update Report List to show separate tables for each form type.

sasha-dresden:
Issue [FECFILE-1933](https://fecgov.atlassian.net/browse/FECFILE-1933)

One of the things I changed to accomplish this was by adding a generic type to the ReportService since it would apply to something extending Report.  However this meant there were a few more points that needed to be updated to make use of the Genericized version of ReportService.

I also included the updates from the ColumnDefinition version of the Table Component from [FECFILE-2042](https://github.com/fecgov/fecfile-web-app/pull/3140)

### DIFF
--- a/front-end/cypress/e2e/F3X/reports-f3x.cy.ts
+++ b/front-end/cypress/e2e/F3X/reports-f3x.cy.ts
@@ -14,9 +14,10 @@ describe('Manage reports', () => {
   it('Create a Quarterly Election Year report', () => {
     ReportListPage.createF3X();
     ReportListPage.goToPage();
-    cy.get('tr').should('contain', 'Form 3X');
-    cy.get('tr').should('contain', 'GENERAL (12G)');
-    cy.get('tr').should('contain', `04/01/${currentYear} - 04/30/${currentYear}`);
+    cy.get('[data-cy="form3x-list-component').as('form3xTable');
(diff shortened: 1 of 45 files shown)


Elaine-Krauss-TCG:
image

Form 99's don't have anything in the Type column, but other than that, this works brilliantly

danguyf (commented):
The ACs say that the pagination default needs to be changed from 10 to 5.  Otherwise this looks great!

Elaine-Krauss-TCG (changes requested):
I left a comment about the Form 99 type value being missing, but that's my only concern.  As soon as that's resolved, I approve.

sasha-dresden:
> Form 99's don't have anything in the Type column, but other than that, this works brilliantly

Ah good catch, I accidentally used the report_code_label instead of the form's subLabel (which usually uses the report_code_label  except for Form 99). Fixed that. 
And also addressed the 5 report default pagination.

danguyf (approved):

Elaine-Krauss-TCG (approved):
There are some Sonarcloud issues that you'll be able to wrap up in like 30 seconds, but I'm not sweating over that.

sasha-dresden:
> There are some Sonarcloud issues that you'll be able to wrap up in like 30 seconds, but I'm not sweating over that.

Got those fixed and also added some padding I noticed got disappeared when I merged in develop.

PR #3194, opened 2025-10-28
Feature/2579

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2579

Related PRs:

### DIFF
--- a/front-end/src/app/layout/layout.component.html
+++ b/front-end/src/app/layout/layout.component.html
@@ -1,7 +1,12 @@
 <div id="page-container" [class]="[layoutControls().backgroundStyle]">
   <app-banner #bannerRef />
   @if (layoutControls().showHeader) {
-    <div class="header-container" [style]="{ paddingBottom: layoutControls().showSidebar ? '0' : '64px' }">
+    <div
+      class="header-container"
+      [style]="{
+        paddingBottom: layoutControls().showSidebar ? '64px' : '128px',
+      }"
+    >
       <app-header [headerStyle]="layoutControls().headerStyle" />
     </div>
   }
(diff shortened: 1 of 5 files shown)


front-end/src/app/layout/layout.component.html
```diff
@@ -1,7 +1,12 @@
 <div id="page-container" [class]="[layoutControls().backgroundStyle]">
   <app-banner #bannerRef />
   @if (layoutControls().showHeader) {
-    <div class="header-container" [style]="{ paddingBottom: layoutControls().showSidebar ? '0' : '64px' }">
+    <div
+      class="header-container"
+      [style]="{
+        paddingBottom: layoutControls().showSidebar ? '64px' : '128px',
```
sasha-dresden: Everything looks good but I was noticing some discrepancies between top spacing when scrolled to the top. I think what we want is for this div to have a fixed 110px padding (this is the height of the 30px USA banner, and the  80px FECFILE banner.
Then update the content-offset to this: [style.marginTop]="isCookiesDisabled() ? '164px' : '64px'"
I believe this gets us to the correct top spacing for both sidebar and non-sidebar pages.

sasha-dresden (changes requested):

front-end/src/app/layout/layout.component.html
Elaine-Krauss-TCG: This was really helpful, thanks

sasha-dresden (approved):

max-zaremba-tcg (approved):
I tested the scroll on multiple pages and everything seems to look good based on the ticket requirements.

danguyf (approved):

PR #3197, opened 2025-10-29
feature/2343 - Allow user to edit Form 24 unique name

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2343

Related PRs:


danguyf:
I found an issue where the new name is not immediately accounted for in subsequent form name validation checks.

If I create an F24 "24-Hour: The Original Name", then rename it to "24-Hour: The New Name", then go to create a new F24 it will warn me that "24-Hour: The Original Name" is already in use and not warn me that "24-Hour: The New Name" is in use -- but will throw an error if I try to create it.

If I then name that new form "24-Hour: The Old Name" this is accepted (good), at which point I see "24-Hour: The New Name" and "24-Hour: The Old Name" (good).  If I go to create a new F24 it will not let me name it either of those things (good) and will allow me to name it "24-Hour: The Original Name" (good).

But if I rename "24-Hour: The Old Name" to "24-Hour: The Oldest Name" and then go to create a new F24, it will warn me that "24-Hour: The New Name" is taken (good) and that "24-Hour: The Old Name" is taken (incorrect) and will allow me to try to name my new form "24-Hour: The Oldest Name" (incorrect) but will error (good).

front-end/src/app/shared/utils/validators.utils.ts
```diff
@@ -225,3 +226,21 @@ export class CommitteeMemberEmailValidator implements AsyncValidator {
...
   }
 }
+
+@Injectable({ providedIn: 'root' })
+export class F24UniqueNameValidator implements AsyncValidator {
+  protected readonly form24Service = inject(Form24Service);
+  async validate(control: AbstractControl): Promise<ValidationErrors | null> {
+    if (control.value) {
+      const reports = await this.form24Service.getAllReports();
+      const existingNames = reports.map((report) => (report as Form24)?.name?.toLowerCase() ?? '') ?? [];
```
danguyf: I assume there is some simple way to give this a kick on `updateName` to tell it to repopulate `existingNames`.

sasha-dresden:
Few things. Most of which are due to my code changes causing some major conflicts.
1. You might want to merge in develop since 1933, which was merged into develop like 30 minutes after you created this PR really shifted around the Report Table action buttons. Should actually end up a lot cleaner for you though. Just move everything from ReportList to the new Form24List
``` 
readonly renameF24DialogVisible = signal(false);
form24ToUpdate?: Form24;
constructor() {
  super();
  this.rowActions.push(new TableAction('Rename', this.renameForm24.bind(this))); // You don't even need to worry about checking type now!
}

public renameForm24(item: Form24): void {
  this.form24ToUpdate = item;
  this.renameF24DialogVisible.set(true);
}
  ```

2. Ticket 729 was also merged in and had massive updates to the FormComponent. Sorry, all my tickets created big changes for this one.
Now all Forms go through the same validation process. The key part is forms need to call the submitForm() function. And the part where it actually submits the form is called submit()

3. You might want to have the successful submit emit something to tell the f24 table to refresh the table with the new name.

**Comment by dheitzer** [comment]
> I found an issue where the new name is not immediately accounted for in subsequent form name validation checks.
> 
> If I create an F24 "24-Hour: The Original Name", then rename it to "24-Hour: The New Name", then go to create a new F24 it will warn me that "24-Hour: The Original Name" is already in use and not warn me that "24-Hour: The New Name" is in use -- but will throw an error if I try to create it.
> 
> If I then name that new form "24-Hour: The Old Name" this is accepted (good), at which point I see "24-Hour: The New Name" and "24-Hour: The Old Name" (good). If I go to create a new F24 it will not let me name it either of those things (good) and will allow me to name it "24-Hour: The Original Name" (good).
> 
> But if I rename "24-Hour: The Old Name" to "24-Hour: The Oldest Name" and then go to create a new F24, it will warn me that "24-Hour: The New Name" is taken (good) and that "24-Hour: The Old Name" is taken (incorrect) and will allow me to try to name my new form "24-Hour: The Oldest Name" (incorrect) but will error (good).

@danguyf it sounds like the issue you're having each time is with F24 creation form if I'm not mistaken.  This ticket only deals with the F24 update form which is currently separate.  The creation code also does the same validation but uses a different mechanism as found [here](https://github.com/fecgov/fecfile-web-app/blob/75b4e4ce2bf8666d074f67ece1a66f9fa7ea8141/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.ts#L41).  I spoke with @toddlees and we may want to try to replace the creation name validation with the update validation.utils implementation.  @toddlees @sasha-dresden thoughts?

**Comment by sasha-dresden** [comment]
> > I found an issue where the new name is not immediately accounted for in subsequent form name validation checks.
> > If I create an F24 "24-Hour: The Original Name", then rename it to "24-Hour: The New Name", then go to create a new F24 it will warn me that "24-Hour: The Original Name" is already in use and not warn me that "24-Hour: The New Name" is in use -- but will throw an error if I try to create it.
> > If I then name that new form "24-Hour: The Old Name" this is accepted (good), at which point I see "24-Hour: The New Name" and "24-Hour: The Old Name" (good). If I go to create a new F24 it will not let me name it either of those things (good) and will allow me to name it "24-Hour: The Original Name" (good).
> > But if I rename "24-Hour: The Old Name" to "24-Hour: The Oldest Name" and then go to create a new F24, it will warn me that "24-Hour: The New Name" is taken (good) and that "24-Hour: The Old Name" is taken (incorrect) and will allow me to try to name my new form "24-Hour: The Oldest Name" (incorrect) but will error (good).
> 
> @danguyf it sounds like the issue you're having each time is with F24 creation form if I'm not mistaken. This ticket only deals with the F24 update form which is currently separate. The creation code also does the same validation but uses a different mechanism as found [here](https://github.com/fecgov/fecfile-web-app/blob/75b4e4ce2bf8666d074f67ece1a66f9fa7ea8141/front-end/src/app/reports/form-type-dialog/create-f24/create-f24.component.ts#L41). I spoke with @toddlees and we may want to try to replace the creation name validation with the update validation.utils implementation. @toddlees @sasha-dresden thoughts?

I think that would be a good idea. The create version does a much more limited API hit since it's only refreshing on load and on create, but if someone else created one while you were in the process...could be an issue.  Since the validation only happens on blur the API hit should be fairly minor and provide more up-to-date results.

**Comment by dheitzer** [comment]
> Few things. Most of which are due to my code changes causing some major conflicts.
> 
> 1. You might want to merge in develop since 1933, which was merged into develop like 30 minutes after you created this PR really shifted around the Report Table action buttons. Should actually end up a lot cleaner for you though. Just move everything from ReportList to the new Form24List
> 
> ```
> readonly renameF24DialogVisible = signal(false);
> form24ToUpdate?: Form24;
> constructor() {
>   super();
>   this.rowActions.push(new TableAction('Rename', this.renameForm24.bind(this))); // You don't even need to worry about checking type now!
> }
> 
> public renameForm24(item: Form24): void {
>   this.form24ToUpdate = item;
>   this.renameF24DialogVisible.set(true);
> }
> ```
> 
> 2. Ticket 729 was also merged in and had massive updates to the FormComponent. Sorry, all my tickets created big changes for this one.
>    Now all Forms go through the same validation process. The key part is forms need to call the submitForm() function. And the part where it actually submits the form is called submit()
> 3. You might want to have the successful submit emit something to tell the f24 table to refresh the table with the new name.

Thanks for merging develop into this branch @sasha-dresden - I believe that takes care of 1 and 2 here.
for # 3, the changes I made in f24-list component ensure that it gets refreshed on dialogclose:
```

effect(() => {
      if (!this.renameF24DialogVisible() && this.form24ToUpdate) {
        this.form24ToUpdate = undefined;
        this.refreshTable();
      }
    });
```

sasha-dresden (changes requested):
Works great!
A few minor styling issues to match The ACs.

1. It looks like there's supposed to be a light box behind the modal so you can't navigate or do anything besides interact with the modal.
    - You can do this by adding: [modal]="true" to the p-dialog
2.  Looks like there should be a dotted one underline to the RENAME link. It looks like you have one but it's the exact same color as the background.

dheitzer:
> Works great! A few minor styling issues to match The ACs.
> 
> 1. It looks like there's supposed to be a light box behind the modal so you can't navigate or do anything besides interact with the modal.
>    
>    * You can do this by adding: [modal]="true" to the p-dialog
> 2. Looks like there should be a dotted one underline to the RENAME link. It looks like you have one but it's the exact same color as the background.

good catch and thanks for the tip on the modal! pushed both these fixes.

sasha-dresden (approved):
Looks great!

danguyf (approved):

danguyf:
Two approvals now, but do you want to knock out the lint failure before merge?

PR #3206, opened 2025-10-30
Feature/2249 - Split Disbursement picker by Form 3X vs Form 3

sasha-dresden:
Issue [FECFILE-2249](https://fecgov.atlassian.net/browse/FECFILE-2249)


front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.ts
```diff
@@ -206,64 +214,99 @@ export class TransactionTypePickerComponent extends DestroyerComponent {
...
-            ScheduleFTransactionTypes.COORDINATED_PARTY_EXPENDITURE,
-            ScheduleFTransactionTypes.COORDINATED_PARTY_EXPENDITURE_VOID,
-            ScheduleETransactionTypes.INDEPENDENT_EXPENDITURE,
-            ScheduleETransactionTypes.INDEPENDENT_EXPENDITURE_VOID,
-            ScheduleETransactionTypes.MULTISTATE_INDEPENDENT_EXPENDITURE,
-            ScheduleETransactionTypes.INDEPENDENT_EXPENDITURE_CREDIT_CARD_PAYMENT,
-            ScheduleETransactionTypes.INDEPENDENT_EXPENDITURE_STAFF_REIMBURSEMENT,
-            ScheduleETransactionTypes.INDEPENDENT_EXPENDITURE_PAYMENT_TO_PAYROLL,
-          ];
+          if (isF3) {
```
toddlees: At this point would it make sense to define what transaction types are available for each form in reports.model and then filter the complete list of each group by whether or not they're current report type uses it?

Since we're here, I think this switch statement has maybe overstayed it's welcome 😆.  Looking at my original commit, i think it was asking too much of the humble switch anyways.  

I noticed you refactored the groups.  Would it make sense to take that to its logical conclusion and have the value of the groups be the actual list of transaction types in that group?  Then here we don't have to have any conditionals, we just reference the list of types defined by the group in question

toddlees (changes requested):

front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.ts
sasha-dresden: This was a good idea. I refactored this moving the transaction group/type logic into the form models themselves. This significantly reduced the complexity of those switch statements. Essentially the only logic left in there is for dealing with PTY vs PAC restrictions and limiting based on environment or if a debt id is present.

I also moved the reports and form models into a subdirectory of the models class for better organization. And I did create a BaseForm3 abstract model which includes duplicate information between the two. I know we don't have a perfect idea of what a Form3 will entail, but for now this includes the bare minimum of what's similar and we could easily move things around as we get more information.

front-end/src/app/reports/submission-workflow/submit-report.component.ts
```diff
@@ -187,7 +187,7 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
       ...SchemaUtils.getFormValues(this.form, this.activeReport().schema, this.formProperties),
     });
 
-    if (payload instanceof Form3X || payload instanceof Form3) {
+    if (payload instanceof BaseForm3) {
       if (this.form.controls['change_of_address']) {
         payload.change_of_address = this.form.value.change_of_address;
         payload.confirmation_email_1 = this.form.value.confirmation_email_1;
```
toddlees: We dont want to conditionally set the confirmation_email_1/2 based on change of address.

I also found a bug with the following lines where we overwrite any change of address info with the existing committee info https://fecgov.atlassian.net/browse/FECFILE-2654

Elaine-Krauss-TCG (approved):

toddlees (changes requested):

front-end/src/app/shared/models/reports/base-form-3.ts
```diff
@@ -0,0 +1,163 @@
...
+    Receipt.REFUNDS,
+    Receipt.OTHER,
+  ];
+  @Exclude()
+  protected loanTransactionGroup = [LoansAndDebts.LOANS, LoansAndDebts.DEBTS];
+
+  @Exclude()
+  protected receiptTransactionMap = new Map<ReceiptType, TransactionTypes[]>([
+    [
+      Receipt.CONTRIBUTIONS_FROM_INDIVIDUALS_PERSONS,
```
toddlees: This is getting there.  In my mind, we're repeating ourselves here by defining a combination of group x transaction type x report type

the way i see the configuration is:

F3_TRANSACTION_TYPES = [] of transaction types
F3X_TRANSACTION_TYPES = [] of transaction types

Groups:
CONTRIBUTIONS_FROM_INDIVIDUALS_PERSONS = {label: string, transaction_types: [] of transaction types in that group}

Then when the group is picked in the picker, we already do some filtering on the list before spitting it out, we could just add `filter out types not in the current report's list of types`

front-end/src/app/shared/models/schc1-transaction.model.ts
```diff
@@ -78,11 +78,11 @@ export class SchC1Transaction extends Transaction {
   }
 }
 
-export enum ScheduleC1TransactionGroups {
-  LOAN_AGREEMENTS = 'LOAN AGREEMENTS',
-}
-
-export type ScheduleC1TransactionGroupsType = ScheduleC1TransactionGroups.LOAN_AGREEMENTS;
+export const ScheduleC1TransactionGroups = {
```
toddlees: I think if the other groups live in transaction-group.ts, these should too

toddlees (changes requested):

front-end/src/app/reports/submission-workflow/submit-report.component.ts
sasha-dresden: Fixed the issue with the confirmation email being based on change of address. Didn't do anything with the other issue to be handled by the new bug ticket

front-end/src/app/shared/models/schc1-transaction.model.ts
sasha-dresden: That makes sense. Moved.

front-end/src/app/shared/models/reports/base-form-3.ts
sasha-dresden: Ah I think I see what you're saying. I was trying to avoid massive transaction lists cause they get a little mind numbing to look at it, but I added a little bit of comments to break it up. I think I have a good setup of what you're looking for in this latest version.

front-end/src/app/shared/models/schc2-transaction.model.ts
```diff
@@ -40,11 +40,11 @@ export class SchC2Transaction extends Transaction {
   }
 }
 
-export enum ScheduleC2TransactionGroups {
-  LOAN_GUARANTORS = 'LOAN GUARANTORS',
-}
-
-export type ScheduleC2TransactionGroupsType = ScheduleC2TransactionGroups.LOAN_GUARANTORS;
+export const ScheduleC2TransactionGroups = {
```
toddlees: this is another one that can be moved

toddlees: Or actually, is this ever even used?  probably not because you can't make one from the picker

toddlees (approved):
I removed the c2 group

PR #3214, opened 2025-11-03
feature/2545 - Create a NEW modal component

max-zaremba-tcg:
Ticket link: [FECFILE-2545](https://fecgov.atlassian.net/browse/FECFILE-2545)

Related PRs:


front-end/src/app/reports/form-type-dialog/form-type-dialog.component.scss
```diff
@@ -52,7 +48,7 @@ label,
 }
 
 .dropdown {
-  width: 400px !important;
+  width: 450px !important;
```
sasha-dresden: Why are you changing this dropdown?

sasha-dresden:
Merged in develop as there were some breaking changes due to [FECFILE-729](https://fecgov.atlassian.net/browse/FECFILE-729)

front-end/src/app/reports/form-type-dialog/form-type-dialog.component.scss
sasha-dresden: I'm a little confused by why you made this change

front-end/src/app/shared/components/committee-member-dialog/committee-member-dialog.component.html
```diff
@@ -1,62 +1,40 @@
...
+  [submitLabel]="member() ? 'Change' : 'Add'"
+  [(visible)]="detailVisible"
+  (submitForm)="submitForm()"
+>
+  <form id="form" [formGroup]="form" [class.ng-submitted]="formSubmitted">
     <div class="dialog-body">
       <div class="grid">
         @if (member()) {
-          <div class="col-12">
+          <div class="col-12 no-padding">
```
sasha-dresden: What you have here is fine, but one of the things I've been trying to do is simplify our html code around all the unnecessary uses of "grid" and "col-12" which lead to some odd padding and html bloat. I'm going to submit a PR into this branch which shows a cleaner alternative. It wasn't necessarily part of this ticket but should make things cleaner.

front-end/src/app/shared/components/dialog/dialog.component.html
```diff
@@ -0,0 +1,35 @@
...
+  <div class="dialog-header">
+    <h1 class="dialog-title" id="dialog-title">{{ title() }}</h1>
+    <a (click)="visible.set(false)" aria-label="Cancel">
+      <img src="assets/img/i-x-circle.svg" />
+    </a>
+  </div>
+  <ng-content></ng-content>
+  <div class="dialog-footer">
+    <hr type="solid" />
+    <div class="grid">
```
sasha-dresden: Another place where we could simplify the html

front-end/src/app/shared/components/committee-member-dialog/committee-member-dialog.component.html
```diff
@@ -1,62 +1,40 @@
-<dialog #dialog tabindex="-1" aria-labelledby="dialog-title">
-  <div class="dialog-header">
-    <h1 class="dialog-title" id="dialog-title">{{ member() ? 'Edit' : 'Add' }} user</h1>
-    <button type="button" class="btn-close" (click)="dialog.close()" aria-label="Cancel"></button>
-  </div>
-  <form id="form" [formGroup]="form" [class.ng-submitted]="formSubmitted" (ngSubmit)="submitForm()">
+<app-dialog
+  tabindex="-1"
+  aria-labelledby="dialog-title"
+  [title]="member() ? 'Edit' : 'Add user'"
```
sasha-dresden: Just noticed this should be 'Edit user' : 'Add user'.  Or you could do it like the old version where it's (edit or Add )+ user

front-end/src/app/shared/components/committee-member-dialog/committee-member-dialog.component.ts
```diff
@@ -66,16 +53,9 @@ export class CommitteeMemberDialogComponent extends FormComponent implements OnC
     return Roles[member.role as keyof typeof Roles];
   });
 
-  readonly dialog = viewChild.required<ElementRef>('dialog');
-
-  ngAfterViewInit() {
-    this.dialog().nativeElement.addEventListener('close', () => this.detailClose.emit());
-  }
-
   ngOnChanges(): void {
```
sasha-dresden: I just realized we could move this to an effect instead of ngOnChanges. Would make more sense there as it is now a signal.

front-end/src/app/shared/components/dialog/dialog.component.scss
```diff
@@ -0,0 +1,91 @@
+dialog::backdrop {
+  background: rgba(0, 0, 0, 0.3);
+}
+
+h1 {
+  color: #212121;
+  font-size: 26px;
+  font-family: var(--gandhi-bold);
+  height: 2.2em;
```
sasha-dresden: Based on the ACs I _think_ this works for the correct spacing between the h1 and the body of the dialog, but I think it might be better to set the height to the font-size and the margin-bottom to 32px. Easier to definitely tell it's 32px spacing then.

front-end/src/app/shared/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  templateUrl: './dialog.component.html',
+  styleUrl: './dialog.component.scss',
+})
+export class DialogComponent {
+  readonly title = input.required<string>();
+  readonly submitLabel = input('Save');
+  readonly visible = model.required<boolean>();
+  readonly userAdded = output<string>();
+  readonly submitForm = output<void>();
+  readonly detailClose = output<void>();
```
sasha-dresden: I just realized we don't need the detailClose output any longer because we just use the visible model.

danguyf (approved):

front-end/src/assets/styles/theme.css
```diff
@@ -668,6 +668,11 @@ p-chips.p-chips-clearable .p-chips-clear-icon {
   padding-right: 0;
   padding-left: 1rem;
 }
+
+.dialog-body .p-select-label {
```
sasha-dresden: So, all these style changes here _feel_ correct. It looks to match the provided screen very well, but none of the ACs mention changing the inputs or labels in this manner. I do prefer the look of these over the previous, especially with the smaller font size. Let's check with Mitchell to make sure.

front-end/src/app/shared/components/dialog/dialog.component.scss
```diff
@@ -0,0 +1,91 @@
...
+  justify-content: space-between;
+}
+
+.field .dialog-label {
+  font-family: var(--karla);
+  font-weight: bold;
+  font-size: 14px;
+  color: #212121;
+  margin-bottom: 8px;
+  text-transform: uppercase;
```
sasha-dresden, line 58: This was causing the dropdown options to be all uppercase.

front-end/src/app/shared/components/dialog/dialog.component.scss
```diff
@@ -0,0 +1,91 @@
...
+  padding-top: 0;
+  margin-top: 16px;
+}
+
+.option {
+  font-size: 14px;
+}
+
+// Both of these are necessary to keep the dropdowns displaying consistently
+:host ::ng-deep p-select > div > p-overlay > div {
```
sasha-dresden: I think this isn't required any longer. Getting rid of it left it looking exactly the same.

sasha-dresden (changes requested):
This looks pretty good. There were a couple styling issues and an issue with the title but otherwise looked great. 
I have a [PR3222](https://github.com/fecgov/fecfile-web-app/pull/3222) into this branch that address these issues. I do like your styling but we should check with Mitchell before making some of those changes since they weren't in the ACs.

sasha-dresden (approved):

PR #3223, opened 2025-11-05
FECFILE-2159: .fec file naming schema

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2159

Related PRs:
N/A

### DIFF
--- a/front-end/src/app/shared/services/dot-fec.service.ts
+++ b/front-end/src/app/shared/services/dot-fec.service.ts
@@ -13,7 +13,7 @@ export interface Download {
   isComplete: boolean;
   // guaranteed unique filename on server
   name: string;
-  // Human-friendly filename for download (e.g. YYYY-MM-DD_SHORTENEDFORMTYPE-FORMNAME)
+  // Human-friendly filename for download (e.g. SHORTENEDFORMTYPE-FORMNAME_YYYY-MM-DD-HHMM)
   filename: string;
   report: Report;
 }
@@ -49,8 +49,11 @@ export class DotFecService {
       report_id: report.id,
     });
 
-    // current date as YYYY-MM-DD
-    const dateStr = new Date().toISOString().slice(0, 10);
+    // current date as YYYY-MM-DD-HHMM
+    const now = new Date();
+    const dateStr = now.toISOString().slice(0, 10) + '-' +
+      now.getHours().toString().padStart(2, '0') +
+      now.getMinutes().toString().padStart(2, '0');
 
     // sanitize helper: remove problematic chars and collapse whitespace
     const sanitize = (s?: string) =>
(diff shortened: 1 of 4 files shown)


sasha-dresden (approved):

front-end/src/app/shared/services/dot-fec.service.ts
```diff
@@ -45,7 +45,35 @@ export class DotFecService {
       },
     );
 
-    const download = { taskId: response.task_id, report, name: response.file_name, isComplete: false };
+    // current datetime as YYYY-MM-DD-HHMM
+    const now = new Date();
+    const dateStr =
+      now.toISOString().slice(0, 10) +
```
dheitzer: Just a preference, but I find `now.toISOString().split('T')[0]` to be a clearer as to what it does.  Up to you if you want to change it though.

front-end/src/app/shared/services/dot-fec.service.ts
```diff
@@ -45,7 +45,35 @@ export class DotFecService {
...
+      '-' +
+      now.getHours().toString().padStart(2, '0') +
+      now.getMinutes().toString().padStart(2, '0');
+
+    // sanitize helper: remove problematic chars and collapse whitespace
+    const sanitize = (s?: string) =>
```
dheitzer, line 57: This is tricky to understand without doing some research.  Can we use String.prototype.replaceAll() on each character we want to remove instead? like;

s = s.replaceAll(' ', '-');
s = s.replaceAll(...);

dheitzer: alternatively (and maybe my preference), we could create a simple mapping from report code to .fec file name prefix similar to [here](https://github.com/fecgov/fecfile-web-api/blob/ae685e85ba1d52c1574c24626d3f973336287d57/django-backend/fecfiler/reports/report_code_label.py#L4)

dheitzer (dismissed):
just some thoughts

front-end/src/app/shared/services/dot-fec.service.ts
Elaine-Krauss-TCG: I second this

Elaine-Krauss-TCG (commented):
I agree with David's suggestions and otherwise approve.

front-end/src/app/shared/services/dot-fec.service.ts
toddlees: I think since we need to ultimately sanitize the user generated name, it makes sense to just use the same rules for the form name rather than making a mapping, which i would be in favor of otherwise

toddlees (approved):

front-end/src/app/shared/services/dot-fec.service.ts
danguyf: > Some people, when confronted with a problem, think “I know, I'll use regular expressions.” Now they have two problems. - jwz

front-end/src/app/shared/services/dot-fec.service.ts
danguyf: Good call.

PR #3228, opened 2025-11-06
Feature/2636 - Merge Sidebars into single sidebar

sasha-dresden:
Issue [FECFILE-2636](https://fecgov.atlassian.net/browse/FECFILE-2636)


danguyf (approved):
Huge improvement.  In my local testing I didn't find any (new) issues (but did find a thing or two to ask about as potential future fixes).

front-end/package-lock.json
```diff
@@ -78,7 +78,6 @@
             "integrity": "sha512-sEyWjw28a/9iluA37KLGu8vjxEIlb60uxznfTUmXImy7H5NvbpSO6yYgmgH5KiD7j+zTUUihiST0jEP12IoXow==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
```
toddlees, line 81: I don't think there were any changes to package.json, so we don't need package-lock.json in this diff

toddlees (changes requested):

front-end/package-lock.json
sasha-dresden: Ah good point. I think I had just been working on the angular upgrade ticket so my packages were all vastly different so I needed to do a fresh install to get back on track. Just did a copy from develop to remove it from the files changed list.

dheitzer (approved):

toddlees (approved):

PR #3237, opened 2025-11-07
feature/2640 - Setup Glossary

sasha-dresden:
Issue [FECFILE-2640](https://fecgov.atlassian.net/browse/FECFILE-2640)


danguyf (approved):

front-end/src/app/shared/components/glossary/glossary.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Individual Receipt": ["Anything of value (money, goods, services or property) received by a political committee."],
+    "Individual Expenditure": ["An expenditure for a communication","That expressly advocates the election or defeat of a clearly identified candidate and","That is not made in cooperation, consultation or concert with, or at the request or suggestion of, any candidate, or his or her authorized committees or agents, or a political party committee or its agents. 11 CFR 100.16."],
```
toddlees: Should be "Independent Expenditure"

toddlees (changes requested):
Mitchell has asked for this to be behind a feature flag. Can you add one to cover the glossary component and the book button?

front-end/src/app/shared/components/glossary/glossary.json
```diff
@@ -0,0 +1,6 @@
+{
```
toddlees, line 1: Really love this solution!  My gut reaction was that there should be a display name and it should be different from the key, but since it's used as the search term as well, this is just really simple and to-the-point until we have more requirements on this

front-end/src/app/shared/components/glossary/glossary.json
```diff
@@ -0,0 +1,6 @@
...
+    "Individual Receipt": ["Anything of value (money, goods, services or property) received by a political committee."],
+    "Individual Expenditure": ["An expenditure for a communication","That expressly advocates the election or defeat of a clearly identified candidate and","That is not made in cooperation, consultation or concert with, or at the request or suggestion of, any candidate, or his or her authorized committees or agents, or a political party committee or its agents. 11 CFR 100.16."],
+    "In-kind Receipt": ["A contribution of goods, services or property offered free or at less than the usual and normal charge. The term also includes payments made on behalf of, but not directly to, candidates and political committees (except for independent expenditures or non-coordinated communications). 11 CFR 100.52(d)."],
```
toddlees, line 4: I don't think this definition matches the transaction type.  In-Kind Receipt is a receipt and not a contribution.  Will reach out to Mitchell in ticket

toddlees (changes requested):

front-end/src/app/shared/components/glossary/glossary.component.html
```diff
@@ -0,0 +1,29 @@
...
+  <h1>Glossary</h1>
+  <div class="field mt-4">
+    <label for="searchTerm">SEARCH TERMS</label>
+    <input id="searchTerm" type="text" [(ngModel)]="searchTerm" />
+    <span class="font-italic mt-1">Example: Conduit Earmark Receipt</span>
+    <div class="mt-4">
+      <div class="my-4 flex justify-content-between">
+        @if (found()) {
+          <span class="font-bold">{{ found() }}</span>
+          <div class="minus-icon-mask"></div>
```
toddlees: In Mitchell's video, he has the minus icon toggle hiding and showing the definition. Asking him now if that needs to be a blocker

toddlees: confirmed that mitchell is hoping for the plus/minus to open/hide the definition

sasha-dresden: Ah, for some reason I didn't realize that was a video haha. Clearly says video over it, but I must've missed that. Was able to implement a collapsible content feature. I also had it animate the collapsing as well as when the size of the content changes, for example going from Independent Expenditure to Individual Receipt.

front-end/src/app/shared/components/glossary/glossary.json
sasha-dresden: Ah good catch. Thank you. Fixed.

front-end/src/app/shared/components/glossary/glossary.json
sasha-dresden: Yea I had no idea where we'd be pulling this data from so I just threw it in a basic JSON format for now so it's a bit more "real" than if we just stored it in the component class itself. Figured we might need to be doing an API pull of some kind to get that data, so this'd be a simple somewhat realistic solution for now.

sasha-dresden:
> Mitchell has asked for this to be behind a feature flag. Can you add one to cover the glossary component and the book button?

Oh I didn't see this until now. Created a feature flag for it. Set to true for local, dev, and stage. False for test and prod.

toddlees (approved):

PR #3241, opened 2025-11-13
[Do not merge] Innovation/admin panel

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2642)

Related PRs: [App](https://github.com/fecgov/fecfile-web-app/pull/3241), [API](https://github.com/fecgov/fecfile-web-api/pull/1767)

### DIFF
--- a/front-end/src/app/admin/admin/admin.component.scss
+++ b/front-end/src/app/admin/admin/admin.component.scss
@@ -1,9 +1,5 @@
-.management-tool {
-  width: 800px;
-}
-
-.justify-content-right {
-  justify-content: right !important;
+.admin-panel {
+	width: 800px;
 }
 
 .justify-content-between {
--- a/front-end/src/app/admin/admin/admin.component.scss
+++ b/front-end/src/app/admin/admin/admin.component.scss
@@ -3,7 +3,7 @@
 }
 
 .justify-content-between {
-	justify-content: space-between !important;
+	justify-content: space-between;
 }
 
 .flex {
--- a/front-end/src/app/admin/admin/admin.component.html
+++ b/front-end/src/app/admin/admin/admin.component.html
@@ -1,0 +1,8 @@
+<h1>Admin Panel</h1>
+<hr type="solid" />
+
+<br /><a href="/admin/dashboard/committees-overview">Registered Committees Overview</a>
+
+<br /><a href="/admin/tools/enable-user">Enable User</a>
+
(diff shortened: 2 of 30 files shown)


front-end/src/app/admin/admin/admin.component.scss
```diff
@@ -0,0 +1,11 @@
+.admin-panel {
+	width: 800px;
+}
+
+.justify-content-between {
+	justify-content: space-between !important;
```
dheitzer: I'm wondering if there's a way around the `!important` which we generally like to avoid.  I'm wondering if you don't need both grid and flex classes on the containing div? (one of those probably is opinionated about justification). Same for the other place we're doing this in the scss.

dheitzer (commented):
Nice clean code! Well done @Elaine-Krauss-TCG ! I like this a lot.  Just a minor thoughts.  Also, I wasn't able to get the routing to work work to the admin page so I wasn't able to test much. 

Like the API PR, I would approve it but I'm not sure we want to integrate this into the code base yet.

danguyf (commented):
I was likewise unable to get the /admin routing working.

front-end/src/app/admin/admin/admin.component.scss
Elaine-Krauss-TCG: You need to do a clean build of the API with the linked branch so that the default test user will have its is_staff value set appropriately.  If you've done that, and the links to the admin panel aren't working, then that'd be very surprising to me.

Elaine-Krauss-TCG: The !important modifiers were a product of a gradual process of trying to wrangle the scss, and it turns out that they're not even necessary in the final state of things.  I was able to remove them with no side effects.

danguyf (commented):
This is really cool. I love the idea of an admin dashboard for us.

I ran into a few issues but none that detract from this as a POC:
* Register Committees is showing me all committees regardless of the "Results per page" setting.  So, for example, by default it's 10 but it's displaying 13 so when I click the "Next" button it just spins.
* I think it would be ideal to replace "Enable user" and "Disable user" with a User Dashboard that shows all users and can be searched and/or filtered by enabled/disabled and then put the enable/disable in an action.  You could tie it into the "Add user to committee" as well, as you have with the Committee Dashboard.
* On the "Add user to committee" page the vertical position and height of "Role" does not match the others.

toddlees:
Closing pr

PR #3247, opened 2025-11-17
feature/2590 - Make Report Names Clickable

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2590

Related PRs:


sasha-dresden (approved):
Works great. I kinda wish it could be solved without the output just so we don't repeat that `(reportNameClick)="this.editItem($event)"` on every form list since it's going to be the exact same thing for all of them, but honestly that's just me being nitpicky haha

dheitzer:
> Works great. I kinda wish it could be solved without the output just so we don't repeat that `(reportNameClick)="this.editItem($event)"` on every form list since it's going to be the exact same thing for all of them, but honestly that's just me being nitpicky haha

haha, I thought the exact same thing and spent about 20m playing around trying to get it to work in report-list.component.html without a huge refactor of the way things are designed.  No luck :(

max-zaremba-tcg (approved):
Tested and works great!

PR #3274, opened 2025-11-25
Feature/2597 - 508: Fix Earmark Transactions

max-zaremba-tcg:
Ticket link: [FECFILE-2597](https://fecgov.atlassian.net/browse/FECFILE-2597)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
+++ b/front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
@@ -8,7 +8,7 @@
   <div class="grid">
     <div class="col-12">
       <div class="field">
-        <label for="searchBox">LOOKUP</label>
+        <span class="span-label" id="contactSearchBoxLabel">LOOKUP</span>
         <app-transaction-contact-lookup
           id="contact_1_lookup"
           contactProperty="contact_1"
--- a/front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
+++ b/front-end/src/app/reports/transactions/transaction-input/transaction-input.component.html
@@ -8,7 +8,7 @@
   <div class="grid">
     <div class="col-12">
       <div class="field">
-        <span class="span-label" id="contactSearchBoxLabel">LOOKUP</span>
+        <span class="span-label" id="contact_searchbox_label">LOOKUP</span>
         <app-transaction-contact-lookup
           id="contact_1_lookup"
           contactProperty="contact_1"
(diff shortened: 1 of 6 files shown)


dheitzer (changes requested):
I think we would prefer to continue using label if possible for semantic reasons within the document.  Is there an alternative to replacing these with span? [To Dan's point](https://fecgov.atlassian.net/browse/FECFILE-2597?focusedCommentId=43186), can we get more info on what the goal of this ticket is? (maybe the specific error info would be helpful).

sasha-dresden (approved):
This seems good. I was a bit worried it would complain about multiple elements with the same ID but it seems ok. That might be something for a future improvement when we finish the Contact Management rework.

Elaine-Krauss-TCG (approved):

max-zaremba-tcg:
> I think we would prefer to continue using label if possible for semantic reasons within the document. Is there an alternative to replacing these with span? [To Dan's point](https://fecgov.atlassian.net/browse/FECFILE-2597?focusedCommentId=43186), can we get more info on what the goal of this ticket is? (maybe the specific error info would be helpful).

This ticket's goal is to fix 508 accessibility errors flagged by WAVE so our app is 508 compliant. WAVE is a Chrome plugin that you can run on each page of the app to check for 508 issues. The "multiple form label" error just means that there are multiple labels assigned to a specific form element. Something like this can confuse screen readers, so that's why it's flagged. 

Spans are used instead of labels because using labels for primeNG elements leads to "orphaned form label" errors. Normally, labels are specifically attached to form elements like select, input, textarea, etc., but something like a p-select doesn't use any of those other elements. From the perspective of software that checks 508 compliance, this appears as if a label is being attached to a non-form element. Hence the error.

dheitzer (approved):

dheitzer:
> > I think we would prefer to continue using label if possible for semantic reasons within the document. Is there an alternative to replacing these with span? [To Dan's point](https://fecgov.atlassian.net/browse/FECFILE-2597?focusedCommentId=43186), can we get more info on what the goal of this ticket is? (maybe the specific error info would be helpful).
> 
> This ticket's goal is to fix 508 accessibility errors flagged by WAVE so our app is 508 compliant. WAVE is a Chrome plugin that you can run on each page of the app to check for 508 issues. The "multiple form label" error just means that there are multiple labels assigned to a specific form element. Something like this can confuse screen readers, so that's why it's flagged.
> 
> Spans are used instead of labels because using labels for primeNG elements leads to "orphaned form label" errors. Normally, labels are specifically attached to form elements like select, input, textarea, etc., but something like a p-select doesn't use any of those other elements. From the perspective of software that checks 508 compliance, this appears as if a label is being attached to a non-form element. Hence the error.

Makes sense @max-zaremba-tcg thanks for the great explanation!

toddlees (approved):

PR #3285, opened 2025-11-28
Circleci parameterized jobs

sVmsepi0l:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2689

When running the parameterized jobs e2e-nightly, e2e-test, and test, they also ran the primary workflow alongside which was unintended. The fix makes it so only those parameterized jobs are run when selected in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,7 +265,10 @@ parameters:
 workflows:
   primary:
     when:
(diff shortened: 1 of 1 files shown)


danguyf (approved):
Makes sense!

toddlees (approved):

PR #3297, opened 2025-12-04
FEFILE-2687: additional test coverage

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2687

Related PRs:
N/A

I readily admit that several of these tests are likely pointless but I think it was a good exercise for me.


sasha-dresden (approved):
This looks pretty good. Definitely adds some more coverage. Some of them are literally just creating the missing spec file, but like those files didn't really have much to test anyways. Others added a whole host of useful tests, like all those ones for the Validators Utils. Looks great! Definitely a great first pass of adding more unit tests. I feel like we might want, at some point in the future as like a second round to this ticket, to go through and look for any specs that just have a basic "I exist" test to see if they should have some further tests added.

dheitzer (approved):

PR #3298, opened 2025-12-05
feature/2663 - Create streamlined transaction serializer for list endpoint

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2663

Related PRs: https://github.com/fecgov/fecfile-web-api/pull/1794

These changes are to create and use a minimal transaction object to be used when querying for a list of transactions (specifically in transaction list page and the contact page) so that we do not have to query/retrieve an entire transaction object for these scenarios.


sasha-dresden:
Ok, made a few changes and converted the items array to be a signal in order to convert the transactions to transaction list dynamically as it's updated.

PR #3307, opened 2025-12-08
Feature/2648 - UI: Update Modal - Delete Transaction - Are you sure

max-zaremba-tcg:
Ticket link: [FECFILE-2648](https://fecgov.atlassian.net/browse/FECFILE-2648)

Related PRs: 
[FECFILE-2649](https://fecgov.atlassian.net/browse/FECFILE-2649)
[FECFILE-2650](https://fecgov.atlassian.net/browse/FECFILE-2650)
[FECFILE-2652](https://fecgov.atlassian.net/browse/FECFILE-2652)

### DIFF
--- a/front-end/src/app/app.component.ts
+++ b/front-end/src/app/app.component.ts
@@ -1,6 +1,5 @@
 import { Component, ElementRef, inject } from '@angular/core';
 import { PollerComponent } from './shared/components/poller/poller.component';
-import { ConfirmDialog } from 'primeng/confirmdialog';
 import { Toast } from 'primeng/toast';
 import { DownloadTrayComponent } from './shared/components/download-tray/download-tray.component';
 import { RouterOutlet } from '@angular/router';
@@ -16,7 +15,6 @@ import { ConfirmDialogComponent } from './shared/components/confirm-dialog/confi
   styleUrls: ['./app.component.scss'],
   imports: [
     PollerComponent,
-    ConfirmDialog,
     Toast,
     DownloadTrayComponent,
     RouterOutlet,
--- a/front-end/src/app/app.component.ts
+++ b/front-end/src/app/app.component.ts
@@ -8,6 +8,7 @@ import { SecondCommitteeAdminDialogComponent } from './shared/components/second-
 import { ButtonModule } from 'primeng/button';
 import { GlossaryComponent } from './shared/components/glossary/glossary.component';
 import { environment } from 'environments/environment';
+import { ConfirmDialogComponent } from './shared/components/confirm-dialog/confirm-dialog.component';
 
 @Component({
   selector: 'app-root',
@@ -22,19 +23,20 @@ import { environment } from 'environments/environment';
     SecondCommitteeAdminDialogComponent,
     ButtonModule,
     GlossaryComponent,
+    ConfirmDialogComponent,
   ],
 })
-export class AppComponent implements AfterContentChecked {
+export class AppComponent {
   protected readonly elementRef = inject(ElementRef);
-  readonly confirmDialog = viewChild.required(ConfirmDialog);
+  // readonly confirmDialog = viewChild.required(ConfirmDialog);
   readonly showGlossary = environment.showGlossary;
 
-  ngAfterContentChecked(): void {
-    const visible = this.confirmDialog().visible;
-    if (!visible) return;
-    const closeButton = (<HTMLElement>this.elementRef.nativeElement).querySelector('.p-dialog-close-button');
-    if (!closeButton) return;
-    if (closeButton.ariaLabel) return;
-    closeButton.ariaLabel = 'Close';
-  }
+  // ngAfterContentChecked(): void {
+  //   const visible = this.confirmDialog().visible;
+  //   if (!visible) return;
+  //   const closeButton = (<HTMLElement>this.elementRef.nativeElement).querySelector('.p-dialog-close-button');
+  //   if (!closeButton) return;
+  //   if (closeButton.ariaLabel) return;
+  //   closeButton.ariaLabel = 'Close';
+  // }
 }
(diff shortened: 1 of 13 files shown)


dheitzer (approved):
looks nice!

front-end/src/app/reports/transactions/transaction-list/transaction-list.component.ts
```diff
@@ -31,6 +32,7 @@ import { TableAction } from 'app/shared/components/table-actions-button/table-ac
     ButtonDirective,
     SelectReportDialogComponent,
     SecondaryReportSelectionDialogComponent,
+    ConfirmDialogComponent,
```
sasha-dresden: You don't need this import any longer since you moved the ConfirmDialogComponent into the AppComponent

front-end/src/app/app.component.ts
```diff
@@ -1,4 +1,4 @@
-import { AfterContentChecked, Component, ElementRef, inject, viewChild } from '@angular/core';
+import { Component, ElementRef, inject } from '@angular/core';
 import { PollerComponent } from './shared/components/poller/poller.component';
 import { ConfirmDialog } from 'primeng/confirmdialog';
```
sasha-dresden: Because we swapped from ConfirmDialog to ConfirmDialogComponent you can get rid of the ConfrimDialog import from here and the @Component section

front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  visible = signal(false);
+  message = '';
+  confirmation?: Confirmation;
+
+  private subscription?: Subscription;
+
+  private confirmationService = inject(ConfirmationService);
+
+  ngOnInit(): void {
+    this.subscription = this.confirmationService.requireConfirmation$.subscribe((conf) => {
```
sasha-dresden: What you have here is perfectly ok. I'm trying to slowly wean us off of life cycle events so we could potentially be zoneless in the future. So a way to write this zoneless you get rid of all the lifecycle events (onInit, onDestroy) and rewrite it like this:

```
 private readonly _confirmation = toSignal(this.confirmationService.requireConfirmation$);
  readonly confirmation = computed(() => (this._confirmation()?.key === this.key() ? this._confirmation() : undefined));
  readonly message = computed(() => this.confirmation()?.message);

  constructor() {
    effect(() => {
      if (this.confirmation()) this.visible.set(true);
    });
  }
```

You'd also need to make a couple updates to the hmtl to change confirmation.header and message into signals. You could even rewrite the header like this `readonly header = computed(() => this.confirmation()?.header ?? 'Are you sure?');` so you don't need to do that or logic in the html.

front-end/src/app/shared/components/dialog/dialog.component.ts
```diff
@@ -23,5 +26,9 @@ export class DialogComponent {
         this.dialog().nativeElement.close();
       }
     });
+
+    effect(() => {
+      this.hasCustomFooter.set(!!this.projectedFooter());
```
sasha-dresden: Since projectedFooter is a signal, you can actually make hasCustomFooter  a computed signal. This way it can't be modified outside of the projectedFooter changing and you don't need to write this effect, it's built right into the computed like this:
`readonly hasCustomFooter = computed(() => !!this.projectedFooter())`

sasha-dresden: If you make the changes add accept/reject or confirm/cancel outputs (that I mentioned in an earlier comment), you can completely get rid of the custom footer logic here

front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
```diff
@@ -0,0 +1,11 @@
+<app-dialog
+  [(visible)]="visible"
+  [title]="confirmation?.header || 'Are you sure?'"
+  submitLabel="Confirm"
+  (submitForm)="confirmOption()"
+>
+  <p>{{ message }}</p>
+  <ng-container dialog-footer #dialogFooterRef>
+    <button pButton class="p-button-secondary" label="Cancel" (click)="cancelOption()"></button>
```
sasha-dresden: Actually looking at these components, all you're really changing with your custom cancel button is what it does. Hmm, I know I said we didn't really need a close dialog emitter, but that would really solve this issue without needing to make a custom component. You might want to rename things a bit like this:

```
readonly confirm = output<void>();
readonly cancel = output<void>();
```
so either confirm/cancel or maybe accept/reject. Either one should work.  Then you could just create a close method method which does:
```
close() {
    this.visible.set(false);
    this.cancel.emit();
  }
```
And then in all three places in the html swap it to `(close)="close()"` or `(click)="close()"`

Then in your confirmDialog html you'd just have:
```
(confirm)="confirmOption()"
(cancel)="cancelOption()"
```

Don't forget to update the committee-member to `(confirm)="submitForm()"`.
I think confirm/accept would be a better output name than submitForm sends we won't always be submitting a form on clicking confirm.

sasha-dresden (commented):
This looks great! And it does work perfectly as intended. I included a couple of code cleanups and a suggestion for an alternative approach instead of using a custom footer template.

front-end/src/app/app.component.ts
```diff
@@ -1,40 +1,30 @@
...
     GlossaryComponent,
+    ConfirmDialogComponent,
   ],
 })
-export class AppComponent implements AfterContentChecked {
+export class AppComponent {
   protected readonly elementRef = inject(ElementRef);
-  readonly confirmDialog = viewChild.required(ConfirmDialog);
   readonly showGlossary = environment.showGlossary;
-
```
toddlees, line 31: while tracking down what this might have been added for, i found the ticket it was implemented in.  Not necessarily this snippet, but we did lose the aria label for the x close button
https://fecgov.atlassian.net/browse/FECFILE-2357

toddlees:
Force itemize was still using the old modal. removed the key and template to use the new one.

Delete report did not have whitespace between lines.  updated

toddlees:
flakey unit test caused me some grief.  doesn't seem related.  created ticket https://fecgov.atlassian.net/browse/FECFILE-2712

PR #3311, opened 2025-12-09
feature/2654 - Change of address on F3X submit report page being overwritten by current committee address

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2654

Related PRs:


front-end/src/app/reports/submission-workflow/submit-report.component.ts
```diff
@@ -193,15 +225,17 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
...
 
-      if (this.form.controls['change_of_address']) {
-        payload.change_of_address = this.form.value.change_of_address;
-        payload.street_1 = this.committeeAccount().street_1;
-        payload.street_2 = this.committeeAccount().street_2;
-        payload.city = this.committeeAccount().city;
-        payload.state = this.committeeAccount().state;
-        payload.zip = this.committeeAccount().zip;
-      }
+      payload.confirmation_email_1 = this.form.value.confirmation_email_1;
```
toddlees: emails and committee namer are a duplicate of line 222

front-end/src/app/reports/submission-workflow/submit-report.component.ts
```diff
@@ -129,6 +130,27 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
           this.form.removeControl('backdoor_code');
         }
       });
+    this.form
+      .get('change_of_address')
+      ?.valueChanges.pipe(takeUntil(this.destroy$))
+      .subscribe((value) => {
```
toddlees: can you update value to be a little more descriptive?  either `change_of_address`, `addressIsChanging`, `enteringNewAddress`, or something similar. reading `value` at line 138 is not as clear as it could be

toddlees (changes requested):

front-end/src/app/reports/submission-workflow/submit-report.component.ts
dheitzer: nice catch - removed them here

front-end/src/app/reports/submission-workflow/submit-report.component.ts
dheitzer: changed it to `changeOfAddress`

toddlees (approved):

toddlees:
Had to update FecUpdate so that it doesn't overwrite changes made to the report.  Example, you change the address, go to download .fec or print preview.  It was undoing the change.  Now it leaves the change.  Question we should raise to BA, this causes a weird scenario if user downloads their .fec or does a webprint, THEN changes their f1 data before submitting.  If they log in again and do a download or webprint, it will be old data

PR #3348, opened 2025-12-12
feature/2651 - UI: Update Modal - Error Outside Coverage Dates - Just checking…

max-zaremba-tcg:
Ticket link: [FECFILE-2651](https://fecgov.atlassian.net/browse/FECFILE-2651)

### DIFF
--- a/front-end/src/app/shared/components/dialog/dialog.component.html
+++ b/front-end/src/app/shared/components/dialog/dialog.component.html
@@ -9,21 +9,21 @@
   <ng-content></ng-content>
   <div class="dialog-footer">
     <hr type="solid" />
-    <div class="flex justify-content-between">
-      <ng-content select="[dialog-footer]"></ng-content>
-      @if (!hasCustomFooter()) {
+    <ng-content select="[dialog-footer]"></ng-content>
+    @if (!hasCustomFooter()) {
+      <div class="flex justify-content-between">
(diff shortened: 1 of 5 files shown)


danguyf (changes requested):
A few AC quibbles:
* The title is 26px not 24px.
* The x icon is 24x24px not 20x20px.

max-zaremba-tcg:
> A few AC quibbles:
> 
> * The title is 26px not 24px.
> * The x icon is 24x24px not 20x20px.

I agree with the title font-size change, but the x icon looks too small when it's set to 20x20px. This was an issue that came-up during the initial development of the dialog wrapper component (2545). I discussed adjusting the size of the X with Mitchell at the time, but I can bring it up with him again as a reminder. I still think the 24x24 size is closer to the desired state.

danguyf (approved):

toddlees (changes requested):
Unfortunately it seems like we've run into a similar problem as we did in the past while working on this modal.  When closing the modal, the date picker is focused and opens up.  You can see on develop that it seems like focus goes to something after the memo item checkbox.  I think it's the error message actually.

@sasha-dresden do you remember how we solved this problem in the past?

toddlees (approved):

toddlees:
Mitchell landed on the new behavior as desired.

PR #3349, opened 2025-12-12
feature/2601 - Convert Deleted Contact to own page and update to use Column Definitions

sasha-dresden:
Issue [FECFILE-2601](https://fecgov.atlassian.net/browse/FECFILE-2601)


danguyf (dismissed):
Everything looks great, with two minor quibbles: 
* The second Task bullet says that the title should be an h1.
* The table cell top/bottom padding should be 10px rather than 14px.

sasha-dresden:
> Everything looks great, with two minor quibbles:
> 
> * The second Task bullet says that the title should be an h1.
> * The table cell top/bottom padding should be 10px rather than 14px.

Oh good point. Fixed.

max-zaremba-tcg (approved):
Good work! Tested locally and seems to match AC.

toddlees (approved):
updated css to only affect the deleted contact table.  reached out to @MitchellTCG to clarify if it should be the same as the other tables or if other tables should be 10px top/bottom padding

PR #3350, opened 2025-12-12
feature/2669 - UI: Update Modal - Create an account - What am I searching

max-zaremba-tcg:
Ticket link: [FECFILE-2669](https://fecgov.atlassian.net/browse/FECFILE-2669)

### DIFF
--- a/front-end/src/app/committee/create-committee/create-committee.component.html
+++ b/front-end/src/app/committee/create-committee/create-committee.component.html
@@ -12,16 +12,10 @@
     <div class="right-side-box">
       <h2 class="m-0">Find your committee</h2>
       <div class="search-label">
-        <label for="search-input" class="font-karla-bold mb-0 mr-1">COMMITTEE ID</label>
-        <i
-          tabindex="0"
-          (keydown.enter)="explanationVisible.set(true)"
-          (keydown.space)="explanationVisible.set(true)"
-          (click)="explanationVisible.set(true)"
-          aria-label="explanation"
-        >
+        <label for="search-input" class="font-karla-bold mr-1 line-height-1">COMMITTEE ID</label>
+        <button class="rounded-image-button" (click)="explanationVisible.set(true)" aria-label="explanation">
           <img alt="question mark" src="assets/img/i-question-circle.svg" />
-        </i>
+        </button>
       </div>
       <p-inputGroup class="search-group">
         <input id="search-input" type="text" class="p-inputtext search-input" [formControl]="searchBoxFormControl" />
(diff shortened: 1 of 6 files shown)


front-end/src/app/committee/create-committee/create-committee.component.html
```diff
@@ -70,30 +70,25 @@ <h3>{{ selectedCommittee.name }}</h3>
...
-    <span class="explanation-header">What am I searching?</span>
-  </ng-template>
-  <p>
+<app-dialog tabindex="-1" aria-labelledby="dialog-title" title="What am I searching?" [(visible)]="explanationVisible">
+  <p id="explanation-text">
     The committee ID lookup searches FEC's global lookup of all committees registered with the Federal Election
     Commission.
-  </p>
-  <p>
+    <br /><br />
```
danguyf: I don't love changing starting a new paragraph (semantic) to two breaks but related ticket FECFILE-2651 did seem to be explicitly calling for it so I'm living with it here.

danguyf (approved):

front-end/src/app/committee/create-committee/create-committee.component.html
sasha-dresden: What you have definitely works, but yea, it would probably be better to do it as two paragraphs like this:

```
<p>
    The committee ID lookup searches FEC's global lookup of all committees registered with the Federal Election
    Commission.
  </p>
  <p class="mb-5">
    If the system cannot find the committee ID entered, please doublecheck that the ID was entered correctly. Also, if
    the committee was just created and the Form 1 was recently submitted, please wait 24 hours for the database to
    update. For all other questions, don't hesitate to get in touch with FECfile+ technical support for assistance.
  </p>
```

Not sure why the default margin-bottom for paragraphs is set as 2.4rem in our typography!? That seems excessive. The browser default is 1em, which works really well here since it's based on the font size we said for the paragraph. In our case 14px so the spacing is 14px.
```
app-dialog p {
  font-family: var(--karla);
  font-size: 14px;
  line-height: 1.3;
  color: var(--p-content-color);
  margin-bottom: 1em;
}
```

sasha-dresden (commented):
So there was one comment I had on the text formatting. It doesn't _need_ to be fixed. One other thing I noticed and I feel like it _should_ be a part of this ticket but it isn't mentioned. You can't tab to the question mark icon. I was trying to make sure keyboard actions still worked with the custom footer (they do!) but I couldn't open the dialog without using the mouse

sasha-dresden (approved):

PR #3355, opened 2025-12-15
feature/2712 - Flakey unit test - transaction type picker

max-zaremba-tcg:
Ticket link: [FECFILE-2712](https://fecgov.atlassian.net/browse/FECFILE-2712)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.spec.ts
+++ b/front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.spec.ts
@@ -30,9 +30,9 @@ import { selectUserLoginData } from 'app/store/user-login-data.selectors';
 describe('TransactionTypePickerComponent', () => {
   let component: TransactionTypePickerComponent;
   let fixture: ComponentFixture<TransactionTypePickerComponent>;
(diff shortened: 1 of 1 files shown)


front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.spec.ts
```diff
@@ -153,22 +156,24 @@ describe('TransactionTypePickerComponent', () => {
     beforeEach(() => {
       fixture = TestBed.createComponent(TransactionTypePickerComponent);
       component = fixture.componentInstance;
+      spyOn(component, 'showTransaction').and.returnValue(true);
```
toddlees, line 159: Is this a correct assessment of the issue? 'debt' was sometimes in the routeparams from the f3x tests causing a bunch of types to be filtered out in the `if (this.debtId()) {` block. 

We might not ned the showTransaction change since that only impacts schedule Fs

toddlees (approved):

front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.spec.ts
toddlees: actually looks like the two are schedule Fs!

I guess we need both!

PR #3366, opened 2025-12-22
FECFILE-2715 video artifacts for failing e2e tests in CircleCI

sVmsepi0l:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2715

testing length of runtime for both test suites:
- control group
- include the same intentionally failing tests in both suites, but with no video
- all passing tests, video setting on (shouldn't keep any video artifacts)
- include the same intentionally failing tests in both suites, video setting on (should keep videos of failing tests and only those)

all of the below have screenshots of failures (even first attempt failures before retry)
when video is on, a video will be recorded of a failed first attempt even if the retry succeeds

e2e-smoke (solo): no video by default, can be ran with e2e-video parameter for video)
e2e-extended (solo): no video by default, can be ran with e2e-video parameter for video)
full-nightly (test, e2e-smoke, e2e-extended): has video by default
nightly: video by default


.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ commands:
     description: "Setup and start fecfile-web-api for E2E tests"
     steps:
       - setup_remote_docker:
-          docker_layer_caching: false
+          docker_layer_caching: true
```
toddlees: I believe this was false for a reason.  I think we want the api to rebuild in case there were changes to it.  Those changes wouldn't trigger a change to the docker file, which is what validates/invalidate the docker_layer_cache

.circleci/config.yml
```diff
@@ -146,58 +146,105 @@ jobs:
   e2e-smoke:
     docker:
       - image: cimg/node:lts-browsers
+    resource_class: large
```
toddlees: Let's not blow our credit budget.  Can you remove this large

front-end/cypress/e2e-extended/contacts/contacts.list.cy.ts
```diff
@@ -78,7 +78,7 @@ describe('Contacts List (/contacts)', () => {
     cy.contains('Save').should('not.exist');
     ContactListPage.goToPage();
     ContactsHelpers.assertColumnHeaders(ContactsHelpers.CONTACTS_HEADERS);
-    cy.contains('button,a', 'Restore deleted contacts').should('exist');
+    cy.contains('button,a', 'Restore deleted contacts').should('not.exist');
```
toddlees, line 81: What prompted changes to this test?  This one in particular seems to reverse the expected behavior?

front-end/cypress.config.ts
```diff
@@ -28,5 +32,37 @@ export default defineConfig({
   e2e: {
     baseUrl: 'http://localhost:4200',
     specPattern: ['cypress/e2e-smoke/**/*.cy.ts', 'cypress/e2e-extended/**/*.cy.ts'],
+    setupNodeEvents(on) {
```
toddlees, line 35: Can you pull this out of the object like this:

```
    setupNodeEvents(on) {
      deleteVideoOnSuccess(on);
    },
  },
});

function deleteVideoOnSuccess(on: Cypress.PluginEvents) {
  on('after:spec', (_spec, results) => {
    if (!results?.video) {
      return;
    }
    const hasFailedAttempt = results.tests?.some((test) => test.attempts?.some((attempt) => attempt.state === 'failed')
    );
    if (hasFailedAttempt) {
      return;
    }
    try {
      fs.rmSync(results.video, { force: true });
    } catch {
      // Ignore cleanup failures
    }
  });
}
```

.circleci/config.yml
```diff
@@ -146,58 +146,105 @@ jobs:
...
+      - store_artifacts:
+          path: /tmp/cypress/screenshots
+          destination: cypress/screenshots
 
-  # An extended e2e test suite running cypress/extended/**/*.cy.ts
+  # An extended e2e test suite running cypress/e2e-extended/**/*.cy.ts
   e2e-extended:
     docker:
       - image: cimg/node:lts-browsers
+    resource_class: large
```
toddlees: same as above comment

front-end/cypress.config.ts
```diff
@@ -28,5 +32,37 @@ export default defineConfig({
...
+        } catch {
+          // Ignore cleanup failures
+        }
+      });
+    },
   },
 });
+
+function resolveCypressVideo(value: string | undefined) {
+  if (!value) {
```
toddlees: I don't think we want most cases to resolve to `true`.   Let's keep it simple with 

```
function getBooleanFromString(string?: string) {
  return string?.toLowerCase() === 'true';
}
```

toddlees (dismissed):

front-end/cypress/e2e-extended/contacts/contacts.list.cy.ts
sVmsepi0l: the restore deleted contacts button had been there because of all the soft deleted contacts, but it should be missing since there are none now

danguyf (approved):
There is some FECFILE-2483 in here as well, muddying things a little, but the FECFILE-2715 bits look sound.

max-zaremba-tcg (approved):
Tentative approval. Please refer to senior dev's opinions for any potential issues.

PR #3397, opened 2026-01-06
Feature/2704 - Create streamlined transaction serializer for list endpoint

sasha-dresden:
Issue [FECFILE-2704](https://fecgov.atlassian.net/browse/FECFILE-2704)
API PR: [PR1837](https://github.com/fecgov/fecfile-web-api/pull/1837)

These changes are to create and use a minimal transaction object to be used when querying for a list of transactions (specifically in transaction list page and the contact page) so that we do not have to query/retrieve an entire transaction object for these scenarios.


danguyf (approved):
All changes look like improvements to me; nothing further to add or suggest. Testing succeeds.

front-end/src/app/shared/models/transaction-list-record.model.ts
toddlees, line 1: what a great idea!  I think this is gonna be pivotal when we tackle "All Transactions"

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
```diff
@@ -459,7 +432,12 @@ export class ContactDialogComponent extends FormComponent implements OnInit {
     this.resetForm();
   }
 
-  async openTransaction(transaction: TransactionData) {
-    await this.router.navigate([`reports/transactions/report/${transaction.report_ids[0]}/list/${transaction.id}`]);
+  async openTransaction(transactionListRecord: TransactionListRecord) {
+    if (transactionListRecord.id) {
+      const transaction = await this.transactionService.get(transactionListRecord.id);
```
toddlees: @sasha-dresden what do you think about updating the streamlined serializer to include the report ids (not the report objects)?  It shouldn't require any joins or queries. It's kind of a bummer to add that extra 100ms to open transactions now and I think it could be a safe win.

toddlees (changes requested):

front-end/src/app/shared/components/contact-dialog/contact-dialog.component.ts
sasha-dresden: Oh that's a great point! I needed to update the backend to pass that along, but it made me realize we can avoid the extra api call in the action button as well! We were already passing the transaction type identifier, so we just needed the converter, and the only other data we needed from the API was the parent_transaction_id. Passing that and we no longer need to make a backend call when the action button is clicked!

toddlees (approved):

PR #3399, opened 2026-01-06
feature/2695 - Update reattribute/redesignation actions to not be available for transaction types with negativeAmountValueOnly

sasha-dresden:
Issue [FECFILE-2695](https://fecgov.atlassian.net/browse/FECFILE-2695)


danguyf:
Am I understanding correctly that with this fix the "Reattribute" option in the Actions should no longer appear for a transaction with a negative amount?  I just added a transaction to a F3X with a negative amount and Reattribute is still in the Actions menu.

toddlees (changes requested):
hold on merging until we can answer requirements question:

Dan's question raises another question which is, should we instead just disable reatt/redes if the amount is negative regardless of type

toddlees (approved):

toddlees:
> hold on merging until we can answer requirements question:
> 
> Dan's question raises another question which is, should we instead just disable reatt/redes if the amount is negative regardless of type

Resolution in ticket comments.  Not concerned with the scenario where a user tries to reatt/redes a transaction with a negative amount that is not in this list

PR #3410, opened 2026-01-08
feature/1962 - Update column sizing breakpoints for report tables

sasha-dresden:
Issue [FECFILE-1962](https://fecgov.atlassian.net/browse/FECFILE-1962)

### DIFF
--- a/front-end/src/app/reports/report-list/report-list.component.scss
+++ b/front-end/src/app/reports/report-list/report-list.component.scss
@@ -94,4 +94,8 @@
       @include fixed-width(6.8%);
     }
   }
+
+  .p-datatable-header {
+    padding-top: 48px;
+  }
 }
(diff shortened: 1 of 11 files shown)


danguyf (dismissed):
I love how efficient this refactor is.

ACs to correct:
* Top and bottom padding in the table cells are 14px instead of 10px.
* The vertical spacing between forms has disappeared entirely for me?

Image

front-end/src/app/reports/report-list/report-list.component.scss
```diff
@@ -1,5 +1,97 @@
-:host ::ng-deep {
-  .p-datatable-header {
-    padding-top: 48px;
```
Elaine-Krauss-TCG, line 3: This padding needs to get re-added

Elaine-Krauss-TCG (dismissed):
The code is beautiful, but I'm going to echo Dan.  Padding in `report-list.component.scss` got removed and I think it needs to be re-added.

front-end/src/app/reports/report-list/report-list.component.scss
toddlees: fixed

toddlees (approved):

PR #3411, opened 2026-01-08
feature/1995 - UI: Update Data Tables on Transactions in this report page

max-zaremba-tcg:
Ticket link: [FECFILE-1995](https://fecgov.atlassian.net/browse/FECFILE-1995)


front-end/src/app/shared/pipes/dynamic.pipe.ts
Elaine-Krauss-TCG, line 1: Beautiful, but I'd like to see this file renamed for clarity

Elaine-Krauss-TCG (approved):

toddlees (changes requested):
Sasha's pr seems like some good changes.  The line_label fix is important to make sorting work on the line_label

One outstanding thing is that it doesn't appear to be setting the line as the default sort order any longer.  Let's sync with @MitchellTCG and @sasha-dresden tomorrow to see if we want to try to patch that or make a followup

toddlees (approved):

PR #3452, opened 2026-01-21
feature/2770 - [UI] Update "Line 6a override" page to fix the all breakpoints

max-zaremba-tcg:
Ticket link: [FECFILE-2770](https://fecgov.atlassian.net/browse/FECFILE-2770)

NOTE TO REVIEWERS:
There are missing requirements from the AC, so the end result does not completely match the desired state image on the PR. Please see my comment on the ticket for details.

### DIFF
--- a/front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
+++ b/front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
@@ -1,6 +1,5 @@
 <h1 class="page-title">Line 6a override</h1>
 <hr type="solid" class="top-hr" />
-
 <form id="form" [formGroup]="form" class="form-layout">
   <div class="form-content">
     <section class="left-content">
@@ -19,7 +18,6 @@
         campaign finance analyst for guidance.
       </p>
     </section>
-
     <section class="right-content">
       <div class="right-grid">
         <div class="field year">
@@ -30,7 +28,6 @@
             [control]="yearFormControl"
           />
         </div>
-
         <div class="field current">
           <label for="currentAmount"
             >CURRENT LINE 6A, <br />
@@ -43,12 +40,10 @@
             inputStyleClass="readonly"
           />
         </div>
-
         <div class="field new">
           <label for="newAmount">NEW AMOUNT</label>
           <app-input-number inputId="newAmount" formControlName="newAmount" />
         </div>
-
         <div class="field action">
           <button
             type="button"
@@ -62,8 +57,6 @@
         </div>
       </div>
     </section>
-
-    <div class="space"></div>
   </div>
 </form>
 <hr type="solid" class="bottom-hr" />
--- a/front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
+++ b/front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
@@ -1,9 +1,10 @@
-<h1>Line 6a override</h1>
+<h1 class="page-title">Line 6a override</h1>
 <hr type="solid" class="top-hr" />
+
 <form id="form" [formGroup]="form" class="form-layout">
   <div class="form-content">
-    <div class="left-content">
-      <strong>WHAT IS A LINE 6A OVERRIDE?</strong>
+    <section class="left-content">
+      <strong class="section-header">WHAT IS A LINE 6A OVERRIDE?</strong>
       <p>
         When there is a discrepancy between the reported cash on hand and the correct cash on hand (bank account
         balance), committees filing a Form 3X can override the reported cash on hand by manually entering an amount for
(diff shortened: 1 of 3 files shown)


front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.html
```diff
@@ -1,64 +1,69 @@
...
-          [disabled]="!newAmountFormControl.dirty"
-          label="Update LINE 6a"
-          class="p-button-primary"
-          (click)="updateLine6a()"
-        ></button>
-      </div>
-    </div>
+    </section>
+
+    <div class="space"></div>
```
sasha-dresden: I'm not quite sure what this "space" div is doing?

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+  .section-header {
+    font-family: var(--karla-bold);
+    margin-bottom: 8px;
+  }
+
+  .top-hr {
+    margin-bottom: 32px;
+  }
+
+  .bottom-hr {
```
sasha-dresden: Was just reading through the ticket comments and realized this isn't really necessary for this ticket since it sounds like a new ticket will address the HRs.

This bottom-hr is actually causing 64px of spacing except when you are in 'sm' due to the p margin-bottom.
My recommendation is to leave this alone (though I prefer using the built in "mt-5" instead of making a custom class). And then update left-content to be a grid with `gap:32px`. You'll need to move the `section-header` into the `<p>` elements (or you could wrap them so there's only 2 elements in the left-content. That will make sure there's never any margin after the last `<p>` element. Or you could also add a rule so the last has 0 bottom margin, but I really like using gap and dislike messing with first/last properties.

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+    height: 36px;
+  }
+
+  .form-layout {
+    display: flex;
+    justify-content: center;
+  }
+
+  .form-content {
+    @include fixed-width(512px);
```
sasha-dresden: you can just use @include standard-container();

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+    gap: 20px;
+    display: grid;
+  }
+
+  .left-content,
+  .right-content {
```
sasha-dresden: Pretty sure this is unnecessary with the way you have everything else set up.

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+    .form-content {
+      @include fixed-width(704px);
+      grid-template-columns: 8fr 4fr;
+    }
+
+    .left-content {
+      grid-column: 1 / 2;
+    }
+
+    .right-content {
```
sasha-dresden: With the way you have everything else set up, you don't need these grid-columns

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+  @include breakpoint('xl') {
+    .form-content {
+      @include fixed-width(852px);
+    }
+  }
+
+  @include breakpoint('xxl') {
+    .form-content {
+      @include fixed-width(1052px);
+      grid-template-columns: 6fr 3fr 3fr;
```
sasha-dresden: You've already declared 
grid-template-columns: 6fr 3fr 3fr; in 'lg' so you don't need to redeclare here.

front-end/src/app/tools/cash-on-hand-override/cash-on-hand-override.component.scss
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  .left-content,
+  .right-content {
+    width: 100%;
+  }
+
+  .right-grid {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
```
sasha-dresden, line 41: Really love you use of fr! This solution to the format of the order of the form elements of the right/bottom elements is amazing!

sasha-dresden (changes requested):

sasha-dresden (approved):
Ok, I think this looks good. I feel like the right grid should probably be 
```
    row-gap: 32px;
    column-gap: 20px;
```
instead of `gap: 20px` but it's a bit unclear from the comments so maybe that's one of the things the future ticket would address (or not address and confirm what you have is perfect)

sVmsepi0l (approved):

PR #3459, opened 2026-01-21
feature/2804 - Make reports with coverages default sort descending coverage

sasha-dresden:
Issue [FECFILE-2804](https://fecgov.atlassian.net/browse/FECFILE-2804)

API [PR1864](https://github.com/fecgov/fecfile-web-api/pull/1864)


danguyf (approved):
This seems like a big improvement overall. Works in my testing.

toddlees (approved):
I have a suggestion in the API pr, but maybe we can leave it to a later ticket if it's really needed

front-end/src/app/shared/models/reports/report.model.ts
```diff
@@ -29,7 +29,8 @@ export abstract class Report extends BaseModel {
   @Type(() => UploadSubmission)
   @Transform(UploadSubmission.transform)
   upload_submission: UploadSubmission | undefined;
-  report_status: string | undefined;
+  @Transform(({ value }) => statusMap[value] || ReportStatus.IN_PROGRESS)
```
toddlees: landed on a best-of-both-worlds approach that handles the sorting with the same field AND outputs the human readable string in the payload.  No need to transform here since the value will be the string

toddlees (changes requested):

toddlees (approved):

toddlees (changes requested):
seeing issue with sort of report code.  looking into it
@sasha-dresden would you be able to see what's causing the first column (report type)

sasha-dresden:
Fixed report type sort order issue by getting report of formSubLabel and just using report_code_label everywhere per our discussion.

toddlees (approved):

PR #3461, opened 2026-01-22
FECFILE-2792: Balance at Close calculation fix

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2792

Related PRs:
N/A

Fixed calculation and added unit and e2e tests.

Additionally, I removed the hardcoded `--browser chrome` from the e2e scripts at the top of `packages.json` so that the configured default browser will be used.


front-end/package.json
```diff
@@ -16,10 +16,10 @@
...
         "test": "npx --node-options='--max-old-space-size=4000' -y -p @angular/cli ng test",
         "e2e": "npx -y -p @angular/cli ng e2e",
-        "e2e:headless": "npm run e2e -- --headless --watch=false --browser chrome",
-        "e2e:extended:headless": "npm run e2e -- --spec 'cypress/e2e-extended/**/*.cy.ts' --headless --watch=false --browser chrome",
-        "e2e:smoke:headless": "npm run e2e -- --spec 'cypress/e2e-smoke/**/*.cy.ts' --headless --watch=false --browser chrome",
-        "e2e:spec": "npm run e2e -- --spec=$npm_config_spec --headless --watch=false --browser=chrome",
+        "e2e:headless": "npm run e2e -- --headless --watch=false",
+        "e2e:extended:headless": "npm run e2e -- --spec 'cypress/e2e-extended/**/*.cy.ts' --headless --watch=false",
+        "e2e:smoke:headless": "npm run e2e -- --spec 'cypress/e2e-smoke/**/*.cy.ts' --headless --watch=false",
+        "e2e:spec": "npm run e2e -- --spec=$npm_config_spec --headless --watch=false",
```
sVmsepi0l, line 22: i set them to chrome so it would be as close as possible to the CI (which specifies chrome like this)

sVmsepi0l: everything else looks great
sonarqube is complaining about a couple of things, though

sVmsepi0l (approved):

front-end/src/app/shared/components/inputs/debt-input/debt-input.component.ts
```diff
@@ -25,12 +25,13 @@ export class DebtInputComponent extends BaseInputComponent implements OnInit {
...
-        amount = isNaN(parseFloat(amount)) ? 0 : parseFloat(amount);
+        const amountNumber = isNaN(parseFloat(amount)) ? 0 : parseFloat(amount);
+        const tx = this.transaction();
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const beginning_balance = parseFloat((<any>this.transaction)[this.templateMap.balance] ?? '0');
+        const beginning_balance = parseFloat(String((tx as any)?.[this.templateMap.balance] ?? 0));
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const payment_amount = parseFloat((<any>this.transaction)['payment_amount'] ?? '0');
-        const balance_at_close = beginning_balance + amount - payment_amount;
+        const payment_amount = parseFloat(String((tx as any)?.['payment_amount'] ?? 0));
```
sasha-dresden: This works but it might be better to do something like this to avoid the unnecessary conversion to any

const tx = this.transaction() as SchDTransaction | undefined;
const beginning_balance = parseFloat(String(tx?.[this.templateMap.balance as keyof SchDTransaction] ?? 0));
const payment_amount = parseFloat(String(tx?.['payment_amount'] ?? 0));
const balance_at_close = beginning_balance + amountNumber - payment_amount;

sasha-dresden (approved):

toddlees (commented):
works great!

PR #3469, opened 2026-01-26
feature/2625 - [UI] Update "Add a receipt", "Add a disbursement", and "Add loans and debts" accordions and pages

max-zaremba-tcg:
Ticket link: [FECFILE-2625](https://fecgov.atlassian.net/browse/FECFILE-2625)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.scss
+++ b/front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.scss
@@ -1,8 +1,81 @@
 @use 'mixins' as *;
 
 :host::ng-deep {
+  .p-accordion {
+    position: relative;
+  }
+
+  .p-accordion::before {
+    content: '';
+    display: block;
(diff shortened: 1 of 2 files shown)


front-end/src/app/reports/transactions/transaction-type-picker/transaction-type-picker.component.scss
```diff
@@ -1,4 +1,106 @@
...
+    width: 100%;
+  }
+
+  .accordion-icon {
+    width: 20px;
+    height: 20px;
+    margin-left: auto;
+  }
+
+  .content-area {
```
sasha-dresden, line 70: you can replace this with
@include standard-container();

Since we use the same breakpoints in nearly all cases, I set this up to make things more streamlined.

sasha-dresden (changes requested):

sasha-dresden (approved):
Looks good!

sVmsepi0l (approved):

PR #3476, opened 2026-01-26
FECFILE-2826: 5xx tripwire update to fail only the 'it' block with the 5xx

sVmsepi0l:
update to run next 'it' block(s) in spec if one of them fails due to a 5xx, not just terminate the whole spec

Ticket link: 
[FECFILE-2826](https://fecgov.atlassian.net/browse/FECFILE-2826)

### DIFF
--- a/front-end/cypress/e2e-extended/users/users.validation.cy.ts
+++ b/front-end/cypress/e2e-extended/users/users.validation.cy.ts
@@ -80,7 +80,7 @@ describe("Users: Validation and API failure states", () => {
     PageUtils.closeToast();
   });
 
-  it('should stub 500 on invite, keep submit enabled, then succeed on retry @allow-5xx', () => {
+  it('should stub 500 on invite, keep submit enabled, then succeed on retry', () => {
     const adminUser = { ...userFormData, role: Roles.COMMITTEE_ADMINISTRATOR };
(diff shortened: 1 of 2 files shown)


sVmsepi0l:
sanity test worked on feature/2802-update (removing one of the allow5xx tags from users.validation.cy.ts to make sure a 5xx only fails that specific ‘it’ block and continues onto the next)

image

5xx artifacts captured during failure

image

for the last two ‘it’ blocks, the allow5xx tag was removed from the first to ensure it captured the failure and moved onto the next 'it block (still containing the allow5xx tag)

sVmsepi0l:
it will keep this from happening (where it fails the entire spec, as seen here [(Artifacts)](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/10316/workflows/17ea4fd0-bb48-4d9a-838e-922f6e833315/jobs/34173/artifacts))

image

danguyf (approved):
Sometimes it's shocking the scaffolding it takes to make a small change.

Would some of the new methods in `front-end/cypress/support/fail-on-5xx.ts` be better placed in a shared utility file for future use by other tests?

sVmsepi0l:
> Sometimes it's shocking the scaffolding it takes to make a small change.
> 
> Would some of the new methods in `front-end/cypress/support/fail-on-5xx.ts` be better placed in a shared utility file for future use by other tests?

I'd be willing to look into placing the new methods elsewhere in a new ticket for sure

PR #3496, opened 2026-01-29
Feature/2547 - Update Previous endpoints to return the amount rather than the entire model

sasha-dresden:
Issue [FECFILE-2547](https://fecgov.atlassian.net/browse/FECFILE-2547)

API [PR1874](https://github.com/fecgov/fecfile-web-api/pull/1874)

List of changes

- Update /previous/ endpoints to return the three aggregate amounts, so APP endpoints update to work with those values directly
- Updated tests appropriately
- Also noticed a deprecation warning. Simple css ordering fix.

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
@@ -162,7 +162,7 @@ export class TransactionFormUtils {
     templateMap: TransactionTemplateMapType,
   ) {
     if (!transaction.transactionType.inheritCalendarYTD) {
-      const previous_election$: Observable<Transaction | undefined> =
+      const previous_election$: Observable<number | null> =
         merge(
           (form.get(templateMap.date) as SubscriptionFormControl).valueChanges,
           (form.get(templateMap.date2) as SubscriptionFormControl).valueChanges,
@@ -206,7 +206,7 @@ export class TransactionFormUtils {
       const parent = await component.transactionService.get(transaction.parent_transaction?.id ?? '');
       const inheritedElectionAggregate = (parent as SchETransaction).calendar_ytd_per_election_office;
       if (inheritedElectionAggregate) {
-        this.updateAggregate(form, 'calendar_ytd', templateMap, transaction, undefined, inheritedElectionAggregate);
+        this.updateAggregate(form, 'calendar_ytd', templateMap, transaction, null, inheritedElectionAggregate);
       }
     }
   }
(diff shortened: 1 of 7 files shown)


danguyf (approved):

front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
```diff
@@ -267,11 +267,10 @@ export class TransactionFormUtils {
     field: TemplateMapKeyType,
     templateMap: TransactionTemplateMapType,
     transaction: Transaction,
-    previousTransaction: Transaction | undefined,
+    previousTransaction: number | null,
```
toddlees: let's call this previousAggregate

front-end/src/app/shared/services/transaction.service.ts
```diff
@@ -40,12 +46,10 @@ export class TransactionService {
...
       );
-      if (response.status === HttpStatusCode.NotFound) {
-        return undefined;
-      }
-      return getFromJSON(response.body);
+
+      return response.body?.aggregate ?? 0;
     }
-    return undefined;
+    return 0;
```
toddlees: my thought on undefined vs 0 is that we lose information by returning 0s on misses.  if getPreviousTransactionForAggregate returns 0 for both a miss and for when the previous transaction's aggregate is 0, we introduce ambiguity.  This second part is not a hard request just a little push back on that move.  I think the current way we're using this method behaves the same way if we use undefined or if we use 0 for all.  I'm just usually more thankful as a debugging developer when i have a better idea of what my return value tells me about what happened.

front-end/src/app/shared/services/transaction.service.ts
toddlees: Love that the result is now the aggregate.  let's change the name of these functions to getPreviousAggregate.  much simpler

toddlees (changes requested):

front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
sasha-dresden: Yea makes sense to update these names.

front-end/src/app/shared/services/transaction.service.ts
sasha-dresden: Yea, I updated the names. I also update the logic to be much more unified so they all flow through the same process.

front-end/src/app/shared/services/transaction.service.ts
sasha-dresden: I went with null since I think there was a spot in the TransactionForm where we had it coming in as Transaction or null previously, so now I just made everything number or null.
The one thing I really dislike about this is we use 404 Not Found to represent that there are no previous transactions, so when you look at the network tab/console it looks like all these failures happen, but they are expected failures. Feel like it'd be better to return like 200 but null results or something to indicate no previous aggregate data from the API. I don't know, just something that has always bothered me with the previous aggregate logic. (Always hate seeing red network calls)

toddlees (approved):
It definitely didn't happen here, but i'm wondering when we started calculating and showing an aggregate before all the dependent fields were filled in.  It doesn't make much sense to show an aggregate before the user has specified all the parts

PR #3509, opened 2026-02-02
feature/2779 - [UI] Update multi-transaction accordions

max-zaremba-tcg:
Ticket link:[ FECFILE-2779](https://fecgov.atlassian.net/browse/FECFILE-2779?search_id=c06ac43d-4abc-4e5b-9d2c-88eb2d37edff)


front-end/src/app/reports/transactions/double-transaction-detail/double-transaction-detail.component.scss
```diff
@@ -1,17 +1,92 @@
...
+    line-height: 20px;
+  }
+
+  .accordion-icon {
+    width: 20px;
+    height: 20px;
+    margin-left: auto;
+  }
+
+  .content-area {
```
sasha-dresden: This should probably be moved to the layout component and applied to the `.main-content` since this won't effect the width of everything correctly unless you apply it to the main-content. Actually this might be a bit complicated because of how it applies to various other pages. I'd recommend this

```
:host::ng-deep {
  .sidebar-content-container > .main-content {
    @include standard-container(true);
  }
 .container {
    @include standard-container();
  }
}
```
That should set the width appropriately when it's a sidebar page and a non-sidebar page, but leave the screens like the landing page alone.

front-end/src/app/reports/transactions/double-transaction-detail/double-transaction-detail.component.scss
```diff
@@ -1,17 +1,92 @@
+@use 'mixins' as *;
+
+:host::ng-deep {
```
sasha-dresden: We might want to just move this and the rest to the transactions.scss just to prevent some duplication of styling.

sasha-dresden (changes requested):
This looks really good! There's a few things that need to be fixed. The two main things are 1. the standard-container isn't applying correctly with how you have it set up. You applied it to the content of the accordion which works great for that section but the Save/Cancel section is outside of that and stretches when it should snap with everything. See my comment on the .scss file with some suggestions.
2. Reattributions/Redesignations aren't updated. The html file there needs the same structure you've done with double and triple.

The only other thing I noticed isn't mentioned in the ticket at all, just the wireframes, there's an extra divider below the guarantor. I don't think this should be handled in this ticket but maybe something we'd want to address eventually. It's a bit annoying cause it's two self-contained components that just happen to have a bad interaction.

sasha-dresden (approved):

PR #3510, opened 2026-02-03
FECFILE-2821: Updated lodash packages to latest.

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2821

Related PRs:
None.


sVmsepi0l (approved):

sasha-dresden (approved):
Looks good. Though this makes me think about the upcoming bundle size optimization ticket. I'm fairly confident we could replace lodash. The only thing we are using it for it cloning transactions and since we already have class-transform we could probably just use that to do the cloning.

PR #3514, opened 2026-02-04
[E2E] removing old cron nightly workflow from old stage branches (Sprint 76)

sVmsepi0l:
[FECFILE-2757](https://fecgov.atlassian.net/browse/FECFILE-2757)


sVmsepi0l:
deleting stale branch instead

PR #3543, opened 2026-02-09
feature/2666 - Add 300ms more delay before searching contact lookup

sasha-dresden:
Issue [FECFILE-2666](https://fecgov.atlassian.net/browse/FECFILE-2666)

There is a built in delay field for the PrimeNG Autocomplete. The default is 300ms so we already had that 300ms delay. So I simply increased it by another 300ms for a total of 600ms.


sVmsepi0l (approved):

toddlees (approved):

toddlees:
works great

PR #3550, opened 2026-02-10
feature/2596 - Update Detailed Summary Page and its Data Tables

max-zaremba-tcg:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2596


sasha-dresden (dismissed):
This looks absolutely great! I love how simplified this became and there was only a single styling issue I noticed. In the Line column there are times when it should be bold and that isn't happening. Love the work you did for the description column. I think you'll just need to do a slightly less complicated version for the line column. 
Also there's some imports you no longer need in the ReportDetailedSummaryComponent like the Card and the pipes.

toddlees (approved):

toddlees (approved):
@max-zaremba-tcg this is really sweet!  nice job!

PR #3558, opened 2026-02-11
fixes debt and report e2e tests

toddlees:


lbeaufort (approved):
Per Slack, [FECFile-2842](https://fecgov.atlassian.net/browse/FECFILE-2842) is supposed to make this less painful.

PR #3576, opened 2026-02-18
feature/2572 -[UI] Move Loan Agreement under the Horizontal Rule

max-zaremba-tcg:
Ticket link: [FECFILE-2572](https://fecgov.atlassian.net/browse/FECFILE-2572?search_id=c69e0875-163c-4960-8afc-c18675c42c11&referrer=quick-find)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-detail/transaction-detail.component.html
+++ b/front-end/src/app/reports/transactions/transaction-detail/transaction-detail.component.html
@@ -5,9 +5,6 @@
     </h1>
   }
 
-  <h1 [class.dark-gray-text]="transactionType?.showParentTransactionTitle">
-    {{ transactionType?.title }}
-  </h1>
   @if (showGlossary) {
     <img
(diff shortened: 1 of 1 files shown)


danguyf (commented):
Does this rely on any other commits?

The code change looks right but when I test this locally it looks very different from the screenshots in the ticket.

max-zaremba-tcg:
> Does this rely on any other commits?
> 
> The code change looks right but when I test this locally it looks very different from the screenshots in the ticket.

Nope. What are you seeing locally? You can view this change in a form 3X when you create a "Loan Received from Bank" and click "New loan agreement" in the actions column within the table where the transaction is listed. It's also visible under "Review loan agreement", which can be found in the same area in the actions column. You have to create multiple form 3Xs that show the same loan to make those options visible though. It's a rare page for sure.

danguyf (approved):

sasha-dresden (approved):

PR #3602, opened 2026-02-24
feature/2827 - Remove reattribute option for In-kind, Partnership, Receipt from Unregistered Entity contributions

max-zaremba-tcg:
Ticket link: [FECFILE-2827](https://fecgov.atlassian.net/browse/FECFILE-2827)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
@@ -169,18 +169,8 @@ export abstract class TransactionListTableBaseComponent
       'Reattribute',
       this.createReattribution.bind(this),
       (transaction: TransactionListRecord) => {
-        const excludedTypes = [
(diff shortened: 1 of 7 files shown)


sasha-dresden (commented):
So what you have works perfectly fine, however I feel like we should set this up as a property of the transaction type on whether or not it can potentially be reattributed.
So what I would recommend is at the base TransactionType model:
`get isReattributable() { return false; }`
at the SchATransactionType:
`override get isReattributable() { return !this.negativeAmountValueOnly; }` 
and then in these three set:
`override get isReattributable()  { return false; }`

And then in the component you just need:
```
transaction.transactionType.isReattributable &&
!transaction.parent_transaction_id &&
!ReattRedesUtils.isReattRedes(transaction, [
  ReattRedesTypes.REATTRIBUTION_FROM,
  ReattRedesTypes.REATTRIBUTION_TO,
]) &&
!ReattRedesUtils.isAtAmountLimit(transaction),
```

sasha-dresden (approved):

PR #3603, opened 2026-02-24
[SPRINT-81]Feature/2878 - Improve lazy loading

sasha-dresden:
Issue [FECFILE-2878](https://fecgov.atlassian.net/browse/FECFILE-2878)

So, kinda a lot of changes...but it mostly boils down to repeating the same thing across all the models.
This does a couple things related to lazy loading. First, it updates the routes so they won't load their component until that specific route is hit. This prevents loading some components before they are used. But the big change is to the TransactionTypeUtils. The way we had this set up before, if anything uses this, it needs to load every single TransactionType model. This wouldn't be a big deal, but every Transaction Type model has an associated validation. This validation file also gets imported. This is what was contributing by far the largest amount to our initial bundle because we were essentially providing the entire FECfile validation suite on the loading of the homepage. So what I did was set this up the dynamically import the file as it is requested. This made all the utility functions into promises which had massive trickle down effects to all the models. I also split that utility file up into TransactionTypeUtils and TransactionUtils because some of it we were using was for TransactionTypes and some for Transactions.  Another issue was that the TransactionTypes referenced the associated Transaction in their model for the fromJson constructor. This meant that if you just loaded the type, the transaction gets bundled with it. So I moved that constructor pattern into the utils s.
Another big change was to get rid of the models index.ts. While this is useful for importing it means all the files in there get bundled together, so if you use one model, you have to load the bundle with all the models.


danguyf:
> every Transaction Type model has an associated validation. This validation file also gets imported. This is what was contributing by far the largest amount to our initial bundle because we were essentially providing the entire FECfile validation suite on the loading of the homepage

Great catch!

danguyf (approved):
This is, as you say, A Lot. I got about 60 files in before I started scrolling looking for new patterns and unique changes, so to speak. I've checked off 123 as (re)Viewed and think I've looked at all the unique changes and a representative sample of the common changes; between that and the passing unit tests I am satisfied.

PR #3613, opened 2026-02-25
feature/2768 - Make Form24  list only fetch when the Form 24 option is selected

sasha-dresden:
Issue[FECFILE-2768](https://fecgov.atlassian.net/browse/FECFILE-2768)


toddlees (approved):
great!

PR #3630, opened 2026-03-02
feature/2881 - [UI] Update Modal - Changed dates in Edit Report Details - Can't save report

max-zaremba-tcg:
Ticket link: [FECFILE-2881](https://fecgov.atlassian.net/browse/FECFILE-2881?search_id=8b6da204-f720-4741-9330-2025bfe9765d&referrer=quick-find)

### DIFF
--- a/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
+++ b/front-end/src/app/reports/f3x/create-workflow/create-f3x-step1.component.html
@@ -86,11 +86,13 @@
 <app-dialog
   [(visible)]="coverageDatesDialogVisible"
   title="Can’t save report"
+  (reject)="onHide()"
   tabindex="-1"
   aria-labelledby="dialog-title"
 >
-  <p>Some transactions would now fall outside your new coverage dates.</p>
-  <p>To fix this, delete those transactions.</p>
+  <p class="mb-0">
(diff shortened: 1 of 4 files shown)


danguyf (changes requested):
The ticket says that the "close" X button should be 20x20 and be positioned -20x-20 from the upper right.  It is 24x24 and positioned -20x-32 from the upper right.

Based on the screenshot and ticket language about z-order, I think that the X should exist independently of the div that contains it and should not affect the width of the hr.

Looking at the code, I think these issues may be with the new dialog.component modal  and not with your PR.

toddlees (changes requested):
Also likely a result of the shared dialog, but the font for the body of the dialog is serif'ed in implementation, but sans-serif in the ticket

danguyf (approved):
Looks right to me now!

danguyf (changes requested):
Based on the errant display I noted occurring on FECFILE-2879 as well, I again think this fix needs to go in the modal component rather than in the instance.

danguyf (approved):
n/m, the fix is already to the dialog component!

sasha-dresden (approved):
Looks great.

toddlees (approved):
nice work!

PR #3641, opened 2026-03-04
feature/2879 - [UI] Update Modal - Edit Form 24 Name - Edit

max-zaremba-tcg:
Ticket link: [FECFILE-2879](https://fecgov.atlassian.net/browse/FECFILE-2879)


danguyf (changes requested):
I have the same quibbles here as I did on FECFILE-2881.  I think that the changes requested there need to be made to the underlying modal template, not per-use.

danguyf (approved):
n/m, the fix is already to the dialog component!

sasha-dresden (approved):

PR #3644, opened 2026-03-05
[WAIT for PR #3668] FECFILE-2923: E2E `GetLoans` intercept / reportListPage flakiness

sVmsepi0l:
Ticket link: [FECFILE-2923](https://fecgov.atlassian.net/browse/FECFILE-2923)

# don't merge yet

[PR #3668](https://github.com/fecgov/fecfile-web-app/pull/3668) needs to be patched into  `release/sprint-81` first,
after which I'll merge `release/sprint-81` into this branch to make sure the changes in this PR
don't get overwritten in a bad merge.

refactored old goToReportList to address the failures in reports-f24-independent-expenditures, debts-balance-calculation, & disbursements


front-end/cypress/e2e-smoke/pages/reportListPage.ts
```diff
@@ -56,10 +56,8 @@ export class ReportListPage {
   }
 
   static editReport(reportName: string, fieldName = 'Edit') {
-    cy.intercept('GET', 'http://localhost:8080/api/v1/reports/**').as('GetReports');
     ReportListPage.goToPage();
-    cy.wait(`@GetReports`);
-    cy.wait(`@GetReports`);
+    cy.contains('td', reportName, { timeout: 10000 }).should('be.visible');
```
danguyf: That we previously had two `.wait`s in a row suggests that the answer to this is "no", but is there no way to wait on the td state rather than a timeout?

sVmsepi0l: i can take out the extended timeout for sure

sVmsepi0l: I'm working on another branch right now but as soon as i get to a stopping point I'll check out here an make the change

toddlees: @sVmsepi0l I don't actually see where in the code base we are calling this.  Since there's no way to verify that this is working as intended, let's not make any changes to it right now, or we can get rid of it

front-end/cypress/e2e-smoke/pages/reportListPage.ts
```diff
@@ -75,53 +73,66 @@ export class ReportListPage {
     ReportListPage.goToPage();
   }
 
+  // backward-compatible alias used by existing smoke specs before big refactors for e2e-smoke
```
toddlees: I see that in my skeleton commit, i renamed goToReporList to goToReportListPage.  Wrapping the new name in the old function doesn't really get us anything. You could change the name back to goToReportList or update the method name in all the places it was called before

front-end/cypress/e2e-smoke/pages/reportListPage.ts
```diff
@@ -75,53 +73,66 @@ export class ReportListPage {
...
-          report_id: reportId,
-          schedules: 'C,D',
-          page: '1',
-          ordering: 'line_label,created',
-          page_size: '5',
-        },
-      }).as('GetLoans');
-    }
+  static goToReportListPage(reportId: string, includeReceipts = true, includeDisbursements = true, includeLoans = true) {
+    ReportListPage.registerReportListInterceptions(reportId, includeReceipts, includeDisbursements, includeLoans);
```
toddlees: so the reason checkReportListPageLoaded has a visit argument is so that it can take on the responsibility of registering the intercepts.  In my commit earlier you can see that goToReportListPage only defines the visit and passes it into the checkReportListPageLoaded.

We want the register in the check method like it was in 5f74f430740b53da16d382202f59b621a3bfd17c

toddlees (dismissed):

front-end/cypress/e2e-smoke/pages/reportListPage.ts
sVmsepi0l: I just removed the timeout, so in the refactor of reportListPage i'll make sure to address this w/ you

front-end/cypress/e2e-smoke/pages/pageUtils.ts
```diff
@@ -51,8 +53,7 @@ export class PageUtils {
   }
 
   static pickDay(day: string) {
-    cy.get('@calendarElement').find('td').find('span').not('.p-disabled').parent().contains(day).click();
```
sVmsepi0l, line 54: it was a duplicate

front-end/cypress/e2e-smoke/pages/reportTransactionListPage.ts
```diff
@@ -0,0 +1,121 @@
+import { F3xCreateReportPage } from './f3xCreateReportPage';
+import { defaultForm24Data, defaultForm3XData as defaultReportFormData } from '../models/ReportFormModel';
+import { PageUtils } from './pageUtils';
```
sVmsepi0l, line 3: renamed reportListPage.ts to this file as well as updated the class and methods so it's showing up

front-end/package-lock.json
```diff
@@ -9732,7 +9732,6 @@
             "version": "0.1.8",
             "resolved": "https://registry.npmjs.org/errno/-/errno-0.1.8.tgz",
             "integrity": "sha512-dJ6oBr5SQ1VSd9qkk7ByRgb/1SH4JZjCHSW/mr63/QcXO9zLVxvJ6Oy13nio03rxpSnVDDjFor75SjVeZWPW/A==",
-            "dev": true,
```
toddlees: I've noticed your package-lock.json changes are often removing `"dev": true` from dev dependencies.  is it possible that you are running npm install with `--legacy-peer-deps` or is `legacy-peer-deps=true` in your .npmrc?  If so, can you make sure you're not running `npm install` with this flag so we don't introduce any dependency bugs?

front-end/cypress/e2e-smoke/F1M/f1m-affiliation.cy.ts
```diff
@@ -68,14 +80,14 @@ describe('Manage reports', () => {
     cy.wrap(Promise.all(apiCalls)).then(() => {
       expect(candidates).to.have.lengthOf(candidateList.length);
 
-      ReportListPage.createF1M();
+      ReportTransactionListPage.createF1M();
       PageUtils.valueCheck('[data-cy="committee-id-input"]', 'C99999999');
-      cy.get('[data-cy="state-party-radio"]').click();
```
toddlees, line 73: we talked about getting rid of this rewrite

.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ jobs:
               -e CYPRESS_EMAIL=${CYPRESS_EMAIL} \
               -e CYPRESS_COMMITTEE_ID=${CYPRESS_COMMITTEE_ID} \
               -e CYPRESS_PASSWORD=${CYPRESS_PASSWORD} \
-              -e CYPRESS_FILING_PASSWORD=${CYPRESS_FILING_PASSWORD} \
+              -e CYPRESS_FILING_PASSWORD="filing!Passw0rd" \
```
toddlees: since we're not using the environment variable any more, we don't need to set one here

front-end/cypress/e2e-extended/contacts/contacts.delete.cy.ts
```diff
@@ -34,7 +34,7 @@ type ContactsDeletedResponse = {
 
 type LinkedContactSearch = Pick<ContactListItem, 'first_name' | 'last_name' | 'name'>;
 
-type ReportListItem = {
+type ReportTransactionListItem = {
```
toddlees: I steered you wrong with the rename.  it should have only been that one goToReportListPage function name that should have changed.  I've gone ahead and fixed that issue on this branch.  The confusion comes from the file being about the actual report list page, but goToReportListPage was actually going to the transaction list in of a report, which happens to have the url path `reports/{id}/list` which is unfortunate in this situation

toddlees (changes requested):
I ran e2e-smoke locally and the `:visible` tags are causing the tests to fail.  I don't want these other changes keeping the goToReportTransactionList fix from getting into this sprint.  We need to just limit this pr to that change and address other things as they come up.

To get this in, this pr needs to be just the report transaction list change and the places where we use it like loans-bank.cy.ts

front-end/package-lock.json
sVmsepi0l: i'll check my rc

front-end/cypress/e2e-smoke/F1M/f1m-affiliation.cy.ts
sVmsepi0l: sorry i don't remember that, getting rid of all the extraneous changes now

sVmsepi0l (commented):
requested changes made

sVmsepi0l:
# don't merge yet

[PR #3668](https://github.com/fecgov/fecfile-web-app/pull/3668) needs to be patched in first
then I'll merge the changes into this branch to make sure the changes don't get overwritten in a bad merge

danguyf (approved):

toddlees (approved):

PR #3647, opened 2026-03-05
Feature/2052 - Setup Split button Add another transaction

sasha-dresden:
Issue [FECFILE-2052](https://fecgov.atlassian.net/browse/FECFILE-2052)


toddlees (changes requested):
Looking REALLY good!  I think you're going to need to use a recipe other than `SAVE_LIST_CONTROL` to match the transaction types described in the ticket.  (like Partnership Receipt should not have this option)

`STANDARD_CONTROLS` seems like a good candidate, but you might still need to add some exceptions if it's not a total match to the requirements

sasha-dresden:
> Looking REALLY good! I think you're going to need to use a recipe other than `SAVE_LIST_CONTROL` to match the transaction types described in the ticket. (like Partnership Receipt should not have this option)
> 
> `STANDARD_CONTROLS` seems like a good candidate, but you might still need to add some exceptions if it's not a total match to the requirements

STANDARD_CONTROLS worked perfectly, it was used in every situation where we were not using Parent or child controls. 
I also updated the styling of the button and dropdown after receiving some feedback from Mitchel to more closely match the requirements.

toddlees (approved):
works great!

PR #3652, opened 2026-03-06
[WIP] feature/2930 - Remove models index to prevent mass model loading

sasha-dresden:
Issue [FECFILE-2930](https://fecgov.atlassian.net/browse/FECFILE-2930)

I just remembered removing the bucket exposed a whole bunch of circular dependencies in our code. Need to fix that before this is ready to go


toddlees:
closing with the deprioritizing of bundle size

PR #3662, opened 2026-03-09
Feature/2883 - [UI] Update Modal - Welcome to FECfile+

max-zaremba-tcg:
Ticket link: [FECFILE-2883](https://fecgov.atlassian.net/browse/FECFILE-2883)

### DIFF
--- a/front-end/src/app/shared/components/dialog/dialog.component.html
+++ b/front-end/src/app/shared/components/dialog/dialog.component.html
@@ -1,4 +1,4 @@
-<dialog #dialog tabindex="-1" aria-labelledby="dialog-title" (close)="close()">
+<dialog #dialog tabindex="-1" aria-labelledby="dialog-title" (close)="close()" (keydown.escape)="handleEscape($event)">
   <div class="dialog-header">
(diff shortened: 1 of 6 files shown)


front-end/src/app/shared/components/dialog/dialog.component.html
```diff
@@ -1,4 +1,4 @@
-<dialog #dialog tabindex="-1" aria-labelledby="dialog-title" (close)="close()">
+<dialog #dialog tabindex="-1" aria-labelledby="dialog-title" (close)="close()" (keydown.escape)="handleEscape($event)">
```
toddlees, line 1: You can actually prevent escape close by just adding `closedby="none"`

toddlees (changes requested):

front-end/src/app/shared/components/dialog/dialog.component.html
toddlees: ignore!  @max-zaremba-tcg had done the research and saw that this would not work with safari

toddlees (approved):
works great

PR #3664, opened 2026-03-10
feature/2856 - Fix spacing under "Is there a line of credit"

max-zaremba-tcg:
Ticket link: [FECFILE-2856](https://fecgov.atlassian.net/browse/FECFILE-2856)


sasha-dresden (changes requested):
So, there's a bit of an issue when the checkbox is clicked. It's causing the spacing between the checkmark and the section that appears to be wrong.

sasha-dresden (approved):
Looks great now

PR #3667, opened 2026-03-11
Feature/2052 - Some styling fixes, and fix to the contact section not hiding on new transaction

sasha-dresden:
Issue [FECFILE-2052](https://fecgov.atlassian.net/browse/FECFILE-2052)

### DIFF
--- a/front-end/src/app/layout/footer/footer.component.scss
+++ b/front-end/src/app/layout/footer/footer.component.scss
@@ -28,7 +28,7 @@ footer label {
 
 #footer-lower {
   height: fit-content;
-  background-color: #164f85;
+  background-color: var(--primary-button-hover-color);
   padding: 5px;
   padding-left: 28px;
 }
--- a/front-end/src/app/login/login/login.component.scss
+++ b/front-end/src/app/login/login/login.component.scss
@@ -19,7 +19,7 @@
 }
 
 .header-productivity {
-  color: #164f85;
+  color: var(--primary-button-hover-color);
 }
 
 .description-text {
--- a/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
+++ b/front-end/src/app/shared/components/navigation-control/navigation-control.component.html
@@ -69,17 +69,19 @@
     </p-popover>
   }
   @case (controlType.SPLITBUTTON) {
-    <div class="split-button-container">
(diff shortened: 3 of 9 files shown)


max-zaremba-tcg (dismissed):
A few things I noticed. The top and bottom margins of the "Save" button are 3px, but the DEV Notes mention making these margins 7px. 

The pipe between the "Save" and arrow buttons is still 27px tall, but the dev notes mention making it 19px.

When the arrow .svg is clicked, the space on the left side of "Add another" looks larger than the right.  The requirements don't seem to mention anything about this, but the wireframe shows the text centered. Since, the requirements don't mention it, this change can be optional. I assume it was probably overlooked when the ticket was created.

sasha-dresden:
> A few things I noticed. The top and bottom margins of the "Save" button are 3px, but the DEV Notes mention making these margins 7px.
> 
> The pipe between the "Save" and arrow buttons is still 27px tall, but the dev notes mention making it 19px.
> 
> When the arrow .svg is clicked, the space on the left side of "Add another" looks larger than the right. The requirements don't seem to mention anything about this, but the wireframe shows the text centered. Since, the requirements don't mention it, this change can be optional. I assume it was probably overlooked when the ticket was created.

Hmm, I'm a bit confused, I definitely changed the height, so that should be good.  I also made sure to add 20px margin to the left and right, so the spacing should be exactly the same. Maybe it looks like more because it measures from the bottom of the 'A' and the top of the 'r' so I don't know, it does kind of make it seem like there's more space with that A.

toddlees:
> A few things I noticed. The top and bottom margins of the "Save" button are 3px, but the DEV Notes mention making these margins 7px.
> 
> The pipe between the "Save" and arrow buttons is still 27px tall, but the dev notes mention making it 19px.
> 
> When the arrow .svg is clicked, the space on the left side of "Add another" looks larger than the right. The requirements don't seem to mention anything about this, but the wireframe shows the text centered. Since, the requirements don't mention it, this change can be optional. I assume it was probably overlooked when the ticket was created.

I double checked and the margins and heights look correct
image

The text also appears to be centered.
image

I'm going to merge, but @max-zaremba-tcg please message me if we're missing something and we can address it

toddlees (approved):
great work!

PR #3675, opened 2026-03-12
feature/2509 - Update Electronic Filing Password

sasha-dresden:
Issue [FECFILE-2509](https://fecgov.atlassian.net/browse/FECFILE-2509)

### DIFF
--- a/front-end/src/environments/environment.cloud.gov.dev.ts
+++ b/front-end/src/environments/environment.cloud.gov.dev.ts
@@ -22,7 +22,7 @@ export const environment = {
   disableLogin: false,
   fecSpec: 8.5,
   showGlossary: false,
-  webForms: 'https://webforms.stage.efo.fec.gov',
+  webForms: 'https://webforms.fec.gov',
 };
 
 /*
--- a/front-end/src/environments/environment.cloud.gov.dev.ts
+++ b/front-end/src/environments/environment.cloud.gov.dev.ts
@@ -22,6 +22,7 @@ export const environment = {
   disableLogin: false,
   fecSpec: 8.5,
   showGlossary: false,
+  webForms: 'https://webforms.stage.efo.fec.gov',
 };
 
 /*
--- a/front-end/src/app/reports/submission-workflow/submit-report.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report.component.html
@@ -115,13 +115,16 @@
     </div>
     <hr type="solid" />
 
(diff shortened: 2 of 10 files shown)


danguyf (approved):

dheitzer (approved):

front-end/src/environments/environment.cloud.gov.dev.ts
```diff
@@ -22,6 +22,7 @@ export const environment = {
   disableLogin: false,
   fecSpec: 8.5,
   showGlossary: false,
+  webForms: 'https://webforms.stage.efo.fec.gov',
```
lbeaufort: @sasha-dresden Sorry if I'm reading the ticket wrong, but should this go to production webforms? 

>DEV, STAGE, PROD:

> Electronic Filing Password Assignment System: https://webforms.fec.gov/psa/help.htm
> 
> Forgot your password?: https://webforms.fec.gov/psa/index.htm

Elaine-Krauss-TCG (approved):

front-end/src/environments/environment.cloud.gov.dev.ts
Elaine-Krauss-TCG: image

The requirements of the ticket are that only TEST uses the test EFO links, but is that right?  I would normally expect that only PROD would use the production EFO links?  Maybe TEST too?  I'm not sure, I guess...

toddlees: Thanks for critically looking at this, everyone.  I can confirm that the AC's are correct.  For the purposes of the links, we want dev and stage to reflect what will be on production, in order to verify production behavior on those environments.  This is different from how we split up the committee info data flow which IS:  
Prod efo/fec -> Prod
Stage efo/fec -> Dev, Stage, Test

I see now that we've actually diverged from this in some place like the f1 link.  So i created a ticket to address that https://fecgov.atlassian.net/browse/FECFILE-2972.

@sasha-dresden can you update this pr so that dev and stage point to prod webforms?

toddlees (changes requested):
Request for change to meet laura's comment above

sVmsepi0l (commented):
if you cherry-pick [this commit](https://github.com/fecgov/fecfile-web-app/commit/d24b3a327ce280fec65623a18c78c458c5521c7e), the knip dependency warning will go away

sVmsepi0l (approved):

toddlees (approved):

PR #3682, opened 2026-03-13
feature/1179 - Update text and styling for security notice

sasha-dresden:
Issue [FECFILE-1179](https://fecgov.atlassian.net/browse/FECFILE-1179)


danguyf (approved):
I'm not clear how we're hiding this in TEST but the changes look correct!

sasha-dresden:
> I'm not clear how we're hiding this in TEST but the changes look correct!

We have two different notices, and which one is used is determined based on environment.

Elaine-Krauss-TCG (approved):

sVmsepi0l (approved):

toddlees (changes requested):
We'll need to make the same sort of change that we did for the password links so that this is following the pattern: dev, stage, prod are the same, test is different

toddlees (approved):
excellent.  i tried an e2e test and it was still able to get through

PR #3686, opened 2026-03-16
feature/2927 - IN_KIND_OUT should not inherit from SCHEDULE_B_MEMO, which requires memo_code

sasha-dresden:
Issue [FECFILE-2927](https://fecgov.atlassian.net/browse/FECFILE-2927)


danguyf (approved):
Reproduced issue, tested fix successfully. I love how small and elegant the fix was.

Elaine-Krauss-TCG (approved):

toddlees (approved):
tested, works great.  Nice solution with the existing system

PR #3705, opened 2026-03-19
Feature/2928 - Upgrade to angular 21 

sasha-dresden:
Issue [FECFILE-2928](https://fecgov.atlassian.net/browse/FECFILE-2928)

Updated to angular 21 went very smoothly. I didn't notice any issues with PrimeNG components, but that's always a concern. I also updated our testing from the deprecated karma to the new angular standard of vitest. This did require updates to most of the unit tests. Most of this process was automated by angular upgrade, however some of the tests exposed issues in our code. For example in the Transaction Guarantor we were passing in a Transaction but later using it as a TransactionListRecord. It also exposed one (of the many) circular dependency issues our code has. Thankfully this one was a relatively straightforward fix of moving the Form24 Unique Name validator from the Validator Utils to the Form24Service as that validator must use the service call. But those were the only two non-spec changes that were necessary.


sVmsepi0l (changes requested):
sonar is complaining ( [link](https://sonarcloud.io/component_measures?id=fecgov_fecfile-web-app&branch=feature%2F2928&metric=new_coverage&view=list) ) about 0% code coverage on the new code (picture below)

Image

sasha-dresden:
> sonar is complaining ( [link](https://sonarcloud.io/component_measures?id=fecgov_fecfile-web-app&branch=feature%2F2928&metric=new_coverage&view=list) ) about 0% code coverage on the new code (picture below)
> 
> Image

I thought that was just cause I had moved things but hadn't actually changed anything but it was actually a good catch because switching from jasmine to vitest, the code coverage changed a little bit, so none of the code coverage was getting processed. Got that resolved.

sVmsepi0l (approved):

front-end/knip.json
```diff
@@ -11,13 +11,14 @@
     "@angular-eslint/schematics",
     "@angular-eslint/template-parser",
     "@typescript-eslint/eslint-plugin",
-    "@typescript-eslint/parser"
+    "@typescript-eslint/parser",
+    "@vitest/coverage-v8"
   ],
+  "ignore": ["src/test.ts"],
```
danguyf: Rather than add this to an ignore list, what if we explicitly add it to the vitest config?
````
  "vitest": {
    "entry": [
      "**/*.{bench,test,test-d,spec,spec-d}.?(c|m)[jt]s?(x)",
      "**/__mocks__/**/*.[jt]s?(x)",
      "src/test.{ts,tsx}"
    ]
  },
````

File: front-end/knip.json [hunk 2]
```diff
@@ -39,11 +40,6 @@
       "front-end/cypress/support/**/*.{js,ts}"
     ]
   },
```
danguyf: This wasn't anything in your commit but this is the default config so can be safely removed.

danguyf (changes requested):
knip updated again and now I'm getting a new warning in my local:
````
$ npx knip --version
6.0.4
$ npx knip
Unused devDependencies (1)
jsdom  package.json:76:6
````
I haven't found a quick way to clue knip in that we're using it so I think we'll want to add it to `ignoreDependencies`.

**Comment by sasha-dresden** [comment]
> knip updated again and now I'm getting a new warning in my local:
> 
> ```
> $ npx knip --version
> 6.0.4
> $ npx knip
> Unused devDependencies (1)
> jsdom  package.json:76:6
> ```
> 
> I haven't found a quick way to clue knip in that we're using it so I think we'll want to add it to `ignoreDependencies`.

hmm, oddly, when I merged develop in this issue resolved. I went ahead and updated the knip.json based on your suggestion though. Definitely makes more sense to have it as an entry rather than an ignore.

File: front-end/knip.json (hunk 2 above)
Comment (danguyf) [reply front-end/knip.json]: And by "the default config" I mean this entire block. I realize that may have been unclear since one of the nodes is named "config".

**Comment by danguyf** [comment]
> hmm, oddly, when I merged develop in this issue resolved. I went ahead and updated the knip.json based on your suggestion though. Definitely makes more sense to have it as an entry rather than an ignore.

I'm still getting the complaint about jsdom, though now knip is on 6.0.5!

**Review by danguyf — APPROVED** [review]
It's passing in circleci even though my local complains about jsdom so maybe it's a fluke in my local.

File: front-end/src/app/reports/shared/print-preview/print-preview.component.spec.ts [hunk 3]
```diff
@@ -97,45 +97,45 @@ describe('PrintPreviewComponent', () => {
...
-    const submit = spyOn(webPrintService, 'submitPrintJob').and.resolveTo({});
-    const poll = spyOn(component, 'pollPrintStatus');
-    const update = spyOn(reportService, 'fecUpdate').and.resolveTo(component.report as Form3X);
-    component.submitPrintJob();
+    const submit = vi.spyOn(webPrintService, 'submitPrintJob').mockResolvedValue({});
+    const poll = vi.spyOn(component, 'pollPrintStatus');
+    const update = vi.spyOn(reportService, 'fecUpdate').mockResolvedValue(component.report as Form3X);
+    await component.submitPrintJob();
 
-    tick(100);
```
dheitzer, line 108: so nice that we can get away from tick for these!

PR #3709, opened 2026-03-20
Backfill main into develop

danguyf:
Ticket link:
N/A

Related PRs:
https://github.com/fecgov/fecfile-web-app/pull/3708

Following the above hotfix, backfilling it into develop (and discovering a previously unbackfilled hotfix).


toddlees (approved):
We noticed that a patch into sprint-76 never got backfilled and up until now only main has had the print preview change.  will look into a way to way to improve process to identify this sooner

PR #3719, opened 2026-03-24
feature/2937 - [UI] Remove 2px border around ALL buttons to match FEC Pattern Library

max-zaremba-tcg:
Ticket link: [FECFILE-2937](https://fecgov.atlassian.net/browse/FECFILE-2937?search_id=47e89ef4-b85a-4e76-8375-708822c00865&referrer=quick-find)

### DIFF
--- a/front-end/src/app/shared/components/navigation-control/navigation-control.component.scss
+++ b/front-end/src/app/shared/components/navigation-control/navigation-control.component.scss
@@ -30,6 +30,7 @@
   color: var(--p-content-color);
   font-weight: bold;
   background-color: var(--orange);
+  border: none;
   padding: 0 16px;
   height: 3rem;
 }
--- a/front-end/src/app/shared/components/save-cancel/save-cancel.component.html
+++ b/front-end/src/app/shared/components/save-cancel/save-cancel.component.html
@@ -16,7 +16,7 @@
       pButton
       pRipple
       appSingleClick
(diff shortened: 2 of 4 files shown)


danguyf (approved):
This appears to do everything requested but I have not checked every button.

sasha-dresden (changes requested):
This looks pretty good, but I noticed there were still a few buttons with borders.
- Secondary buttons still have `border: 1px solid #5b616b;` (You can see this on the "Cancel" button when adding a transaction.
- Disabled buttons still have `border: 2px solid gray;` (you can see this on the "Save both transactions" button of adding a loan, before the button activates.
- "Warning" buttons still have `border: 1px solid #fdb838;` (this can be seen on the Print Preview page

Ok, I think that is it. I combed through the theme.css and I think those are the only instances of borders on buttons that we actually use. Unfortunately there is a ton of dead styles in there that make it a bit challenging to easily browse through. For example there is still a border on the "success" variant of the p-button, but we don't use that anywhere. Also there are a few places in the theme where we change the border-color when you hover that are still, but since the border is now gone, doesn't really matter.
For example this section
```
.p-button:enabled:active {
  background: var(--primary-button-hover-color);
  color: #ffffff;
  border-color: var(--primary-button-hover-color);;
}
```

sasha-dresden (approved):

PR #3736, opened 2026-03-27
feature/2991 - Disable field when system generated

sasha-dresden:
Issue [FECFILE-2991](https://fecgov.atlassian.net/browse/FECFILE-2991)


toddlees (approved):
Perfect!  This is especially great because it leaves the error in place if the user tries to save.  In the off chance that we somehow generate a description that's too long and they get all the way to being able to save and it fails, we'll want that i think

PR #3741, opened 2026-03-27
feature/2571 - Fix dissemination date and disbursement date validation syncing

sasha-dresden:
Issue [FECFILE-2571](https://fecgov.atlassian.net/browse/FECFILE-2571)

### DIFF
--- a/front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
+++ b/front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
@@ -7,7 +7,6 @@ import { InputNumber } from 'primeng/inputnumber';
 import { BaseInputComponent } from '../base-input.component';
 import { MemoCodeInputComponent } from '../memo-code/memo-code.component';
 import { Form3X } from 'app/shared/models/reports/form-3x.model';
-import { SubscriptionFormControl } from 'app/shared/utils/subscription-form-control';
 import { CalendarComponent } from '../../calendar/calendar.component';
 import {
   LinkedReportInputComponent,
@@ -52,7 +51,6 @@ export class AmountInputComponent extends BaseInputComponent implements OnInit {
   readonly memoItemHelpText = input<string | undefined>();
 
   readonly amountInput = viewChild.required<InputNumber>('amountInput');
-  readonly memoCode = viewChild(MemoCodeInputComponent);
   readonly contributionAmountInputStyleClass = computed(() => (this.contributionAmountReadOnly() ? 'readonly' : ''));
 
   readonly isF24 = computed(() => this.report().report_type === ReportTypes.F24);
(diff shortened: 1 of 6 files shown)


danguyf (changes requested):
I was able to reproduce the bug locally and validated that this fixes it.

In my testing, though, the warning modal now does not repeat the date, i.e. it displays "Just checking... / falls outside the report coverages dates..." rather than "Just checking... / 09/09/2026 falls outside the report coverages dates..."

sasha-dresden:
> I was able to reproduce the bug locally and validated that this fixes it.
> 
> In my testing, though, the warning modal now does not repeat the date, i.e. it displays "Just checking... / falls outside the report coverages dates..." rather than "Just checking... / 09/09/2026 falls outside the report coverages dates..."

Oh good catch! I was trying to join everything into a single process instead of a couple separate processes but looks like I forgot to move the date set. Got that in and it should behave properly now

danguyf (approved):
Tested. Complete success.

toddlees:
I only noticed this because i was testing the earmark receipts before. 
If I open an existing earmark receipt where the memo transaction's date is outsite the coverage dates i see this image

toddlees (changes requested):
can i have you look into the issue in the comment above?

sasha-dresden:
Ah hmm, the issue with it loading up with the memo control was already ticked was a strange issue. I think the problem was that we were essentially handling the date value changes and the memo code value changes through two separate processes that might reference the other, but were separate. So when I updated the logic for the date controls it was getting the dates before the memo control was set, so it was seen as null. To fix this I merged the date change logic and the memo control change logic into one pathway and things all one through one channel now which actually simplifies things a bit and fixes the bug where it had the dates before it had the memo control value.

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
```diff
@@ -89,22 +89,16 @@ export class AmountInputComponent extends BaseInputComponent implements OnInit {
...
       if (dateControl && date2Control) {
-        dateControl.addSubscription(() => {
+        dateControl.addSubscription((val) => {
           date2Control.updateValueAndValidity({ emitEvent: false });
-          this.memoCode()?.coverageDateQuestion.set(
-            'Did you mean to enter a date outside of the report coverage period?',
-          );
-          // Opening of 'Just checking...' pop-up is handled in app-memo-code component directly.
+          if (val) {
+            this.memoCode()?.coverageDateQuestion.set('Did you mean to enter a date...');
```
toddlees, line 95: Need to correct these messages

toddlees (changes requested):

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
sasha-dresden: I fixed this yesterday?

toddlees (approved):

front-end/src/app/shared/components/inputs/amount-input/amount-input.component.ts
toddlees: yes you did 😀

PR #3744, opened 2026-03-27
feature/2965 - Change password autocomplete to off

sasha-dresden:
Issue [FECFILE-2965](https://fecgov.atlassian.net/browse/FECFILE-2965)

Changed the password autocomplete to "off". This allowed me to select the password from a list of saved passwords and when doing so _only_ populated the password field and nothing else. This worked for me in chrome and firefox. Interestingly, if I saved the password the first time I entered it, it keyed off of the EMAIL #1 field which makes a bit more sense. And would autopopulate the password based off that field, only running into issues when my saved password had a different associated email.

### DIFF
--- a/front-end/src/app/reports/submission-workflow/submit-report.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report.component.html
@@ -135,7 +135,7 @@
         formControlName="filingPassword"
         [feedback]="false"
         [toggleMask]="true"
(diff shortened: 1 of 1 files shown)


danguyf (approved):
I was able to reproduce the issue on the current release/sprint-82 branch and then confirmed that it is resolved as described on this branch.

PR #3758, opened 2026-03-31
Feature/2924 - Allow for editing reports in failed submission state

sasha-dresden:
Issue [FECFILE-2924](https://fecgov.atlassian.net/issues?filter=10787&selectedIssue=FECFILE-2924)

API [PR1969](https://github.com/fecgov/fecfile-web-api/pull/1969)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/transaction-list.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-list/transaction-list.component.ts
@@ -105,7 +105,6 @@ export class TransactionListComponent {
   readonly loans = viewChild.required(TransactionLoansAndDebtsComponent);
 
   readonly isForm24 = computed(() => this.report().report_type === ReportTypes.F24);
-  readonly isNotSubmitted = computed(() => this.report().report_status !== ReportStatus.SUBMIT_SUCCESS);
 
   async createTransactions(transactionCategory: string, report: Report): Promise<void> {
     await this.router.navigateByUrl(`/reports/transactions/report/${report?.id}/select/${transactionCategory}`);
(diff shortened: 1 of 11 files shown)


front-end/src/app/reports/transactions/transaction-list/transaction-list.component.ts
```diff
@@ -105,7 +105,7 @@ export class TransactionListComponent {
   readonly loans = viewChild.required(TransactionLoansAndDebtsComponent);
 
   readonly isForm24 = computed(() => this.report().report_type === ReportTypes.F24);
-  readonly isInProgress = computed(() => this.report().report_status === ReportStatus.IN_PROGRESS);
+  readonly isNotSubmitted = computed(() => this.report().report_status !== ReportStatus.SUBMIT_SUCCESS);
```
toddlees: This PR is looking great!  

I think this is gonna include SUBMIT_PENDING, which isn't a state where we want to allow users to add a transaction.  Is there a nice way to use the same logic as we do in the row actions?  Should we add some utils to report.model?

button appears while pending
image

toddlees (changes requested):

front-end/src/app/reports/transactions/transaction-list/transaction-list.component.ts
sasha-dresden: Yea, good idea. I added a canEdit property to reports. This made it easy to reuse in both places. It also worked as a good toggle for Edit/Review since I noticed in Failed status you were able to click Edit and Review, which didn't make much sense since they did the same thing essentially.

toddlees (approved):

PR #3766, opened 2026-04-02
feature/3019 - Fix css issue with accordion content space style being used when hidden

sasha-dresden:
Issue [FECFILE-3019](https://fecgov.atlassian.net/browse/FECFILE-3019)


toddlees (approved):
Works great!

PR #3768, opened 2026-04-03
feature/3009 - Update addGuarantor to wait for the appropriate PUT/POST to finish before proceeding. Created a framework for intercepting/waiting on requests

sasha-dresden:
Issue {FECFILE-3009](https://fecgov.atlassian.net/browse/FECFILE-3009)


danguyf (approved):
Great optimization and tests passing.

front-end/cypress/e2e-smoke/pages/transactionDetailPage.ts
```diff
@@ -297,37 +296,51 @@ export class TransactionDetailPage {
...
+      cy.wait('@saveGuarantor');
+      PageUtils.closeToast();
+      PageUtils.urlCheck('create-sub-transaction' + '/C2_LOAN_GUARANTOR');
+      this.clickCancel();
+    };
+
+    TransactionDetailPage.interceptTransactions(reportId, fx);
+  }
+
+  static interceptTransactions(reportId: string, fx: Function) {
```
toddlees: This seems to be tackling the same problem as 
```
gotToReportTransactionListPage
checkReportTransactionListPageLoaded
registerReporTransactiontListInterceptions
waitForReportTransactionListRequests
```
Any simple way of deduplicating that?

toddlees (changes requested):

front-end/cypress/e2e-smoke/pages/transactionDetailPage.ts
sasha-dresden: Ah I hadn't noticed that, yea, these are essentially doing the same things. Each has their pros and cons, so I merged it to try and get the pros of both of them.

toddlees (approved):
Works great when i add the manual slowdown to the api
```
if request.method == "PUT" and "transaction" in request.path:
            time.sleep(4)
```

PR #3781, opened 2026-04-08
(wait for Todd) feature/3011 - Update /previous/ calls to limit number of calls made so e2e tests won't crash

sasha-dresden:
Issue [FECFILE-3011](https://fecgov.atlassian.net/browse/FECFILE-3011)

So the underlying issue was the way /previous/ calls worked. It was essentially making 3 calls to the API every time one of the pertinent values changed. It was also triggering with every change, so when you type in the year it was doing a search for '2', '20', '202', and '2026'. Each of those was 3 calls to the API, so just to select which type it was and the year was 15 API calls and only the last 3 were even potentially valid. This was a massive problem when I set the backend to a 4 second delay on each of those, it was taking well over 30 seconds to complete all of them. So I updated the way we handled this so now it will only send 1 per change and for the election year, it won't attempt until it has a valid value. This solved 90% of the problem.

The other problem is we can't simply wait on this endpoint to complete. Well we can, but the problem is we don't know exactly how many /previous/ calls will be made. I've done what I can to limit the potential number, but it's possible there could be one or could be multiple depending upon how the form is interacted with.

The other issue is that we treat No previous transactions as 404 NOT FOUND, but I don't feel like this is a good http code to use. We should be returning 200 OK with an empty response and handle that on the front end. 404 seems to indicate the endpoint was mistyped or something, not that there just weren't any results. It would be like opening up the list of transactions and when there are no transactions for a report return 404 there. Part of the issue with this is that it's hard to generalize a test script when we're unsure if the result will be 404 or 200. Or just how many there will be. I tried to resolve this by using `cy.waitForNetworkIdle(1000)` which should allow us to wait for any number of responses and response types to finish, but it's not ideal since it's essentially a slightly more advanced cy.wait(1000) However all of this together does resolve the problem even with a 4 second delay on /previous/ calls.


front-end/cypress/e2e-extended/reports/f3x/f3x-aggregation.helpers.ts
```diff
@@ -474,6 +478,7 @@ export class F3XAggregationHelpers {
     TransactionDetailPage.enterSheduleFormDataForVoidExpenditure(formData, args.candidate, false, '', 'date_signed');
     this.clearAndType('#electionYear', `${electionYearFromCode}`);
     cy.blurActiveField();
+    cy.waitForNetworkIdle(1000);
```
danguyf, line 481: I too don't love this but, as you've noted, it allows us to bypass a number of possible issues. Great band-aid, and maybe we can create a follow-up ticket to address the underlying issues that you've raised in the OP.

danguyf (approved):

lbeaufort (changes requested):
Noting merge conflicts

sasha-dresden:
Got the merge conflicts resolved. Also noticed a wierd code smell notification so I got that resolved as well.

dheitzer:
Regarding's @sasha-dresden's great writeup on the issues, I also agree that we should not be returning 404 when no previous transactions are found as 404 is a [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#client_error_responses) according to the http spec.  @toddlees shall we create a ticket for this perhaps?

toddlees (approved):

toddlees (approved):

PR #3786, opened 2026-04-09
feature/3023 - Allow split button to work no matter how many have been done before it

sasha-dresden:
Issue [FECFILE-3023](https://fecgov.atlassian.net/browse/FECFILE-3023)

Remove second model from splitButton


toddlees (approved):
perfect!

PR #3789, opened 2026-04-10
feature/2747 - Update Contact Dialog from Transaction Page to allow any contact type.

sasha-dresden:
Issue [FECFILE-2747](https://fecgov.atlassian.net/browse/FECFILE-2747)


danguyf (changes requested):
In my testing I am not getting the ability to change the contact type in the modal.  I have tried a few different types of F3X transactions.  (This is the second frontend ticket this morning where I am not able to see the expected fix so it's possible that there is something amiss with my local frontent.)

danguyf (approved):
I take it all back. I misunderstood the change and how to test it.

I am able to reproduce the original issue and confirm that this addresses the request. When the Lookup allows a type selection then the type in the Create New Contact modal allows the type to be changed/selected.  (And when the Lookup does not allow a selection, just as an Individual Receipt or a Party Receipt, nor does the modal -- exactly as we wish.)

PR #3790, opened 2026-04-10
feature/2974 - Update knip to latest version and pin it

sasha-dresden:
Issue [FECFILE-2974](https://fecgov.atlassian.net/browse/FECFILE-2974)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ orbs:
 # Reusable Commands see https://circleci.com/docs/reference/reusing-config/
 commands:
   spin-up-api-containers:
-    description: 'Setup and start fecfile-web-api for E2E tests'
+    description: "Setup and start fecfile-web-api for E2E tests"
     steps:
       - setup_remote_docker:
           docker_layer_caching: false
@@ -73,7 +73,7 @@ jobs:
           working_directory: ~/project/front-end/
 
       - run:
-          name: 'Run Prettier with project configuration'
+          name: "Run Prettier with project configuration"
           command: npm run prettier:check:ci
           working_directory: ~/project/front-end/
   test:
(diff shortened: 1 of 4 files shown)


toddlees (changes requested):
The second part of this ticket is to use the version installed by npm rather than `npx knip` in config.yml

sasha-dresden:
Oh wierd, we essentially had mixed and matched the dependency version and the non-dependency version (by installing a dependency but then using npx) I went through the knip instructions on https://knip.dev/overview/getting-started to get us set up using the dependency version.

danguyf:
> Oh wierd, we essentially had mixed and matched the dependency version and the non-dependency version (by installing a dependency but then using npx) I went through the knip instructions on https://knip.dev/overview/getting-started to get us set up using the dependency version.

I think we need to update `.circleci/config.yml:302` to not use npx.  Or will these changes to devDependencies pin it for that npx call?

danguyf (approved):
Looks good!

sasha-dresden:
> > Oh wierd, we essentially had mixed and matched the dependency version and the non-dependency version (by installing a dependency but then using npx) I went through the knip instructions on https://knip.dev/overview/getting-started to get us set up using the dependency version.
> 
> I think we need to update `.circleci/config.yml:302` to not use npx. Or will these changes to devDependencies pin it for that npx call?



> > Oh wierd, we essentially had mixed and matched the dependency version and the non-dependency version (by installing a dependency but then using npx) I went through the knip instructions on https://knip.dev/overview/getting-started to get us set up using the dependency version.
> 
> I think we need to update `.circleci/config.yml:302` to not use npx. Or will these changes to devDependencies pin it for that npx call?

Ugh I hate the directory structure. I'm always in /front-end/ so you can run everything so when you run "npm add ." it doesn't grab the ".circle/config.yml"

lbeaufort (dismissed):
Thanks for your work on this, Sasha! Just noting that it looks like we've got merge conflicts

sasha-dresden:
> Thanks for your work on this, Sasha! Just noting that it looks like we've got merge conflicts

Conflicts resolved.

toddlees (approved):

PR #3792, opened 2026-04-10
remove loading icon for contact lookup

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3032

### DIFF
--- a/front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
+++ b/front-end/src/app/shared/components/contact-lookup/contact-lookup.component.html
@@ -53,6 +53,7 @@
         <ng-template pTemplate="dropdownicon">
           <img alt="Magnifying lens" src="assets/img/contact_lookup_magnifying_glass.svg" height="20px" width="20px" />
         </ng-template>
+        <ng-template pTemplate="loadingicon"><!--no loading icon--></ng-template>
       </p-autoComplete>
     </div>
   }
--- a/front-end/src/app/shared/components/contact-search/contact-search.component.html
+++ b/front-end/src/app/shared/components/contact-search/contact-search.component.html
@@ -48,6 +48,7 @@
         <ng-template pTemplate="dropdownicon">
           <img alt="Magnifying lens" src="assets/img/contact_lookup_magnifying_glass.svg" height="20px" width="20px" />
         </ng-template>
+        <ng-template pTemplate="loadingicon"><!--no loading icon--></ng-template>
       </p-autoComplete>
     </div>
   </div>


danguyf (changes requested):
This change makes logical sense but in my testing the loading spinner is still appearing behind the magnifying glass.

toddlees:
> This change makes logical sense but in my testing the loading spinner is still appearing behind the magnifying glass.

Thanks for testing Dan.  Do you know which one you're not seeing the change on?  I was surprised that I had to change two separate p-autocompletes

danguyf:
> Thanks for testing Dan. Do you know which one you're not seeing the change on? I was surprised that I had to change two separate p-autocompletes

I was adding an Individual Receipt to a F3X.  I'm now testing Sasha's FECFILE-2747 and also not seeing that change so I'm wondering if my node isn't rebuilding like it claims it is.

sasha-dresden:
> > This change makes logical sense but in my testing the loading spinner is still appearing behind the magnifying glass.
> 
> Thanks for testing Dan. Do you know which one you're not seeing the change on? I was surprised that I had to change two separate p-autocompletes

There's two because of the interrupted process of updating the Contact Management Service. We started a process of moving that to a new process, but we never finished and now have two different versions of that component structure.

lbeaufort:
> > > This change makes logical sense but in my testing the loading spinner is still appearing behind the magnifying glass.
> > 
> > 
> > Thanks for testing Dan. Do you know which one you're not seeing the change on? I was surprised that I had to change two separate p-autocompletes
> 
> There's two because of the interrupted process of updating the Contact Management Service. We started a process of moving that to a new process, but we never finished and now have two different versions of that component structure.

Thank you @sasha-dresden! @toddlees should we put in a cleanup ticket for combining these if there isn't one already?

lbeaufort (approved):
Looks good, minus the comment about consolidating the components

danguyf (approved):
I came back to this after testing another web-app ticket and now it's working. /shrug emoji

sasha-dresden:
> > > > This change makes logical sense but in my testing the loading spinner is still appearing behind the magnifying glass.
> > > 
> > > 
> > > Thanks for testing Dan. Do you know which one you're not seeing the change on? I was surprised that I had to change two separate p-autocompletes
> > 
> > 
> > There's two because of the interrupted process of updating the Contact Management Service. We started a process of moving that to a new process, but we never finished and now have two different versions of that component structure.
> 
> Thank you @sasha-dresden! @toddlees should we put in a cleanup ticket for combining these if there isn't one already?

@lbeaufort @toddlees I created [FECFILE-3038](https://fecgov.atlassian.net/browse/FECFILE-3038) to address finishing up the contact management upgrade process.

PR #3804, opened 2026-04-15
feature/3048 - Fix console log error

sasha-dresden:
Issue [FECFILE-3048](https://fecgov.atlassian.net/browse/FECFILE-3048)

I was able to track down the issue to the PrintPreview component. The issue was that that the polling for the print status was happening as the component was being destroyed and there was a bit of randomness that if the component got destroyed right as it was attempting to do something it would fail and try and to log the error but everything was being torn down around it so the RPC was closed before it had a chance to log. I updated our destroyer component to keep track of component on destroy status to stop the log from happening (which would prevent the specific test failure) and also updated the polling logic so it would stop polling when the component is destroyed rather than erroring out.

### DIFF
--- a/front-end/src/app/reports/shared/print-preview/print-preview.component.ts
+++ b/front-end/src/app/reports/shared/print-preview/print-preview.component.ts
@@ -138,7 +138,7 @@ export class PrintPreviewComponent extends DestroyerComponent implements OnInit 
         this.pollPrintStatus();
       }
     } catch (error) {
-      console.error('Error fetching report:', error);
+      if (this.componentAlive) console.error('Error fetching report:', error);
     }
   }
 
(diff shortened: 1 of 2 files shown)


danguyf (commented):
Can you share how to force/reproduce the issue and a summary of the cause and how this addresses it?  I think I get it, working backwards, but I'm hesitant to assume.

dheitzer (approved):
looks nice

danguyf (approved):

toddlees (approved):

sasha-dresden:
> Can you share how to force/reproduce the issue and a summary of the cause and how this addresses it? I think I get it, working backwards, but I'm hesitant to assume.

Oh! Sure! So, @danguyf  it didn't help with tracking down the issue, but to force this problem to occur I just went into a component and in the ngOnInit function added this:
```
setInterval(() => {
  console.log('TEST');
}, 1);
```

That forced a log to just keep happening indefinitely with no stop and happen often. Went from maybe 1 failure out 10 with the PrintPreview problem to like 50 problems every run.

PR #3805, opened 2026-04-15
feature/2950 - Update select committee for disabled committees

sasha-dresden:
Issue [FECFILE-2950](https://fecgov.atlassian.net/browse/FECFILE-2950)


front-end/src/app/committee/select-committee/select-committee.component.html
```diff
@@ -36,6 +36,46 @@ <h3>
       <span class="h5">LOOKING FOR AN EXISTING FECFILE+ COMMITTEE ACCOUNT?</span>
       Contact the administrator for that committee.
     </div>
+
+    <div class="disabled-committee-header" [class]="disabledShown() ? 'header-expanded' : ''">
```
toddlees: it's not in the ac, but i believe the intent is that if the user is not part of any disabled accounts, this component will not be visible at all.  will confirm with UX/BA

toddlees (changes requested):

front-end/src/app/committee/select-committee/select-committee.component.html
danguyf: I assumed so as well.

danguyf (dismissed):
This is really cool and works in my testing.

I have another AC UI question, though, that maybe @toddlees can ask about when he's checking on his. I think that when there is one or more disabled committees the disabled committees box should default to open.

front-end/src/app/committee/select-committee/select-committee.component.html
toddlees: UX confirmed that the disabled section should not appear if there are 0 disabled committees.

We also talked about @danguyf 's question below.  UX confirmed that the accordion should start out closed

danguyf:
> I have another AC UI question, though, that maybe @toddlees can ask about when he's checking on his. I think that when there is one or more disabled committees the disabled committees box should default to open.

Todd says they want it to default to closed so that members of disabled committees have to go hunting to figure out why they're not showing up.

front-end/src/app/committee/select-committee/select-committee.component.html
sasha-dresden: Ok, got that updated. Yea, I had been wondering about that, but there hadn't been anything in the ticket about it hiding so I figured they just wanted it always visible.

toddlees (approved):
This is great work @sasha-dresden

PR #3814, opened 2026-04-16
feature/3030 - Cleanup memo code styling

sasha-dresden:
Issue [FECFILE-3030](https://fecgov.atlassian.net/browse/FECFILE-3030)


front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
```diff
@@ -1,7 +1,7 @@
 <div [formGroup]="form">
-  <div class="grid memo-container">
+  <div class="flex flex-column gap-3">
```
danguyf: .gap-3 creates a 1rem/16px gap that looks larger than the Ideal State comps.  I think maybe .gap-2?

sasha-dresden: Yea, looking at it. I definitely agree gap-2 looks much closer to the ACs. This also made me review the ACs and I hadn't noticed the call to update the icon. So I fixed the icon and cleaned up the spacing there a little as well. (Need to check both regular transaction and conduit transaction to see both types)

danguyf (approved):

dheitzer (approved):

toddlees (approved):
thanks for the update on the padding

PR #3833, opened 2026-04-20
Feature/3012 Warn users on login page when API is unavailable

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3012

Related PRs: Followed by [#2014](https://github.com/fecgov/fecfile-web-app/pull/3856)


front-end/src/app/layout/service-unavailable-banner/service-unavailable-banner.component.html
```diff
@@ -0,0 +1,16 @@
+<section
+  id="service-unavailable-banner"
+  class="service-unavailable-banner"
+  aria-label="FECfile+ is temporarily unavailable: Please try logging in again shortly."
+>
+  <header class="service-unavailable-banner__header">
+    <div class="service-unavailable-banner__inner">
+      <div class="service-unavailable-banner__header-text">
+        <p aria-label="FECfile+ is temporarily unavailable: Please try logging in again shortly.">
+          <i class="pi pi-exclamation-circle" style="padding-top: 8px"></i>
```
dheitzer: I think we want to avoid inline styles due to csp policy

danguyf:
Is there a good way to test this locally?  I downed my API container but didn't get the warning -- though I also didn't get the error page when I clicked on the login button, which might count as an improvement.

front-end/src/app/layout/service-unavailable-banner/service-unavailable-banner.component.scss
```diff
@@ -0,0 +1,16 @@
+.service-unavailable-banner {
```
dheitzer: a nit, but I would propose changing these two class names to simply `banner` and `banner-header-text` since we're already in the service unavailable banner component

front-end/src/app/shared/services/api.service.ts
```diff
@@ -76,6 +76,17 @@ export class ApiService {
     );
   }
 
+  public get_from_devops<T>(endpoint: string, params?: QueryParams): Promise<T> {
+    const headers = this.getHeaders();
+    return firstValueFrom(
+      this.http.get<T>(`${environment.apiUrl.split('/api')[0]}${endpoint}`, {
```
dheitzer: I might recommend that we update our environment files like so:
```
const baseUrl = 'https://dev-api.fecfile.fec.gov';
const apiUrl = `${baseUrl}/api/v1';

```
then simply refer to `baseUrl` here to avoid any brittle breakage/string parsing if we changes our apiUrl going forward.

I also think having a baseUrl may be useful going forward if we have other non-api resources.

front-end/src/app/shared/components/dialog/dialog.component.html
```diff
@@ -1,6 +1,11 @@
 <dialog #dialog tabindex="-1" aria-labelledby="dialog-title" (close)="close()" (keydown.escape)="handleEscape($event)">
   <div class="dialog-header">
-    <h1 class="dialog-title" id="dialog-title">{{ title() }}</h1>
+    <h1 class="dialog-title" id="dialog-title">
+      @if (!!showTitleWarning()) {
+        <i class="pi pi-exclamation-circle" style="padding-top: 8px"></i>
```
dheitzer: also maybe avoid inline styles here too

front-end/src/app/shared/services/api.service.ts
```diff
@@ -76,6 +76,17 @@ export class ApiService {
     );
   }
 
+  public get_from_devops<T>(endpoint: string, params?: QueryParams): Promise<T> {
```
dheitzer: Maybe we can rename this to `get_from_base_uri` or similar (since it need not always be devops).

front-end/src/app/layout/header/header-links/header-links.component.ts
```diff
@@ -19,18 +23,27 @@ import { HeaderStyles } from '../header-styles';
...
+  async navigateToLoginDotGov() {
+    const available = this.serviceAvailable();
+    if (available === false) {
+      this.store.dispatch(setServiceAvailableAction({ payload: undefined }));
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const response: HttpResponse<any> = await this.apiService.get_from_devops('/devops/status/');
+    if (response.status) {
+      this.store.dispatch(setServiceAvailableAction({ payload: true }));
+      globalThis.location.href = this.loginDotGovAuthUrl ?? '';
```
dheitzer, line 46: nice catch using `globalThis`

dheitzer (changes requested):
looks nice! just a few thoughts

front-end/src/app/layout/service-unavailable-banner/service-unavailable-banner.component.html
Elaine-Krauss-TCG: Good call

dheitzer:
> Is there a good way to test this locally? I downed my API container but didn't get the warning -- though I also didn't get the error page when I clicked on the login button, which might count as an improvement.

@danguyf yes I ran into the same issue.  I noticed that OPTIONS is getting a CORS error when api is downed (at least locally) and am not clear if this is the issue.

I was able to test this by changing `get_current` to return 502 status in the API manually.

front-end/src/app/layout/layout.component.ts
```diff
@@ -82,6 +89,8 @@ export class LayoutComponent implements AfterViewChecked {
       mobileQuery.addEventListener('change', listener);
       this.destroyRef.onDestroy(() => mobileQuery.removeEventListener('change', listener));
     }
+
+    this.store.dispatch(setServiceAvailableAction({ payload: true }));
```
dheitzer: @Elaine-Krauss-TCG when I was testing locally, I was unable to get the banner to appear due to this line here.  It seems to be overriding whatever is set in the interceptor on error.

I'm wondering if we even need this since the state is defaulted to `true` [here](https://github.com/fecgov/fecfile-web-app/blob/a4a523c94b1dd453b1f1d55980331b2082b174ce/front-end/src/app/store/service-available.reducer.ts#L4)?

dheitzer (changes requested):
changes look good! only one final question

front-end/src/environments/environment.cloud.gov.prod.ts
```diff
@@ -2,10 +2,12 @@
 // `ng build --configuration=cloud.gov.prod` replaces `environment.ts` with `environment.cloud.gov.prod.ts`.
 // The list of file replacements can be found in `angular.json`.
 
+const baseUri = 'https://api.fecfile.fec.gov';
 const apiUrl = 'https://api.fecfile.fec.gov/api/v1';
```
dheitzer: one final small request - can we update these `apiUrl` in the environment files to reference `baseUri` (so we aren't repeating the base portion)?

so like `const apiUrl = '${baseUri}/api/v1';` but with the fancy ticks that markdown won't let me use :)

Elaine-Krauss-TCG:
> Is there a good way to test this locally? I downed my API container but didn't get the warning -- though I also didn't get the error page when I clicked on the login button, which might count as an improvement.

Hi, @danguyf sorry I missed this yesterday!  I tested this locally by having the API return responses with 502 errors.  I did this on the /devops/status endpoint in order to test the login page, and I did this on the /reports list endpoint to test that the banner and modal did not appear on non-login pages.

You want to `return Response(status=502)` specifically

front-end/src/app/layout/layout.component.ts
Elaine-Krauss-TCG: I have no issue being able to get the banner or modal to appear in my testing, but this line is vestigial and I'm ripping it out.

Elaine-Krauss-TCG: Turns out, this was _not_ vestigial.  I forgot why I had put this here, but further testing reminded me of the origin.  The store doesn't get emptied on page reloads, so if serviceAvailable gets set to false, this modal will pop up every time you load a page.

Elaine-Krauss-TCG: Aaaaaaaaaaand I was wrong about this.  I was working on FECFILE-2014, and the modal popping up on reload was caused by work there and not by this line having been removed.  The line has been removed again...

danguyf:
> Hi, @danguyf sorry I missed this yesterday! I tested this locally by having the API return responses with 502 errors. I did this on the /devops/status endpoint in order to test the login page, and I did this on the /reports list endpoint to test that the banner and modal did not appear on non-login pages.
> 
> You want to `return Response(status=502)` specifically

At the risk of being pedantic, are we limiting ourselves overly much by only looking for HTTP502s?  If I down the API entirely and it's not responsive shouldn't we display this warning banner??

Elaine-Krauss-TCG:
> > Hi, @danguyf sorry I missed this yesterday! I tested this locally by having the API return responses with 502 errors. I did this on the /devops/status endpoint in order to test the login page, and I did this on the /reports list endpoint to test that the banner and modal did not appear on non-login pages.
> > You want to `return Response(status=502)` specifically
> 
> At the risk of being pedantic, are we limiting ourselves overly much by only looking for HTTP502s? If I down the API entirely and it's not responsive shouldn't we display this warning banner??

It's a really, really fair concern, but that's the letter of the ticket
image
Or at least my interpretation

danguyf:
> > At the risk of being pedantic, are we limiting ourselves overly much by only looking for HTTP502s? If I down the API entirely and it's not responsive shouldn't we display this warning banner??
> 
> It's a really, really fair concern, but that's the letter of the ticket image Or at least my interpretation

I agree that it's what the ticket specifies but "API is still down" **does not** necessarily equate to an HTTP502. I think that bit of the ticket is unnecessarily restrictive. Paging @toddlees and @lbeaufort ?

front-end/src/app/shared/interceptors/http-error.interceptor.ts
```diff
@@ -31,6 +32,9 @@ export class HttpErrorInterceptor implements HttpInterceptor {
           return of(new HttpResponse({ status: error.status }));
         }
         let errorMessage = '';
+        if (error.status === 502) {
+          this.store.dispatch(setServiceAvailableAction({ payload: false }));
+        }
```
danguyf, line 37: As outlined in my comments, I think checking only for HTTP502 is a problem. I think **any** 5xx error or a status of 0 should trigger this warning.

I propose instead:
```
const isServerError = Math.trunc(error.status / 100) === 5;
// status 0 means no HTTP response was received (offline, refused, DNS, CORS, etc.).
const isTransportFailure =
    error.status === 0 && !(error.error instanceof ProgressEvent && error.error.type === 'abort');

if (isServerError || isTransportFailure) {
    this.store.dispatch(setServiceAvailableAction({ payload: false }));
}
```

danguyf (changes requested):
I love everything about this. My ideological argument is with the ticket ACs.  I propose it should instead read:

> **Given** a user navigates to the login page on FECfile+,
> 
> **When** the API is offline or otherwise unresponsive (when the app receives *no response or a 5xx*),
> 
> **Then** they will see a banner stating “FECfile+ is temporarily unavailable: Please try logging in again shortly.”
> 
> **And** when they attempt to login and the API is still down (when the app receives *no response or a 5xx*)
> 
> **Then** they will see a modal stating “ FECfile+ is experiencing an issue. Please wait a moment and try to log in again. We apologize for the inconvenience.” 
> 
> **And** when they close out of the modal they will return to the login page
> 
> **And** when they try to log in again but the API is up
> 
> **Then** they will login as normal

I have provided a code update to cover this.

front-end/src/app/shared/interceptors/http-error.interceptor.ts
Elaine-Krauss-TCG: I definitely like the idea of a 0 triggering this.  I think a normal 500 error might not be something we want to trigger on, maybe.  The server throwing any uncaught exception is not necessarily something that we want to force a logout over.  But I'd definitely want to hear Laura and Todd's thoughts on it too!  I'll add it to the dev sync for Monday

danguyf (approved):
Does what is requested. We can add triggering on further error codes down the road.

dheitzer (approved):
changes looks great - very nice work on this @Elaine-Krauss-TCG !

front-end/src/app/login/login/login.component.ts
```diff
@@ -25,18 +36,41 @@ export class LoginComponent implements OnInit, AfterViewChecked {
     this.loginDotGovAuthUrl = environment.loginDotGovAuthUrl;
 
     window.addEventListener('resize', this.updateScrollbarWidth);
+
+    this.store
```
toddlees, line 40: We should be able to handle this with computed signals. something like 

```
constructor() {
    super();

    effect(() => {
      const available = this.serviceAvailable();
      if (available === false) {
        this.loginDialogVisible.set(true);
      }
    });
  }
```

toddlees (changes requested):
looking great! let's get some styling things sorted first:

Image

toddlees:
> looking great! let's get some styling things sorted first:
> 
> Image

OH we already have this svg.
can you do something like:
```
<svg alt="Exclamation mark">
          <use href="assets/img/exclamation.svg"></use>
        </svg>
      ```

front-end/src/app/shared/interceptors/http-error.interceptor.ts
toddlees: Good conversation all.  We've done a fair amount of deliberating with FEC on what we want to handle now and we are scoping it just to 502s which is what we get in the scenario where we shut down the api.  We'll revisit other cases later

toddlees:
We will now just merge in 2014 with all the changes together

front-end/src/app/shared/interceptors/http-error.interceptor.ts
danguyf: If the API server is down and not responding then the status is 0, not 502.  The server has to be up for a 502 response.

PR #3847, opened 2026-04-21
feature/735 - Make Conduit Earmark Out count as deposited/undeposited, but hide the selector.

sasha-dresden:
Issue [FECFILE-735](https://fecgov.atlassian.net/browse/FECFILE-735)

Updated so both Step 1 and Step 2 will show the UNDEPOSITED text in the popup. As a side effect this changed the memo item of Step 2 to be similar to Step 1. This is actually closer to the end goal.


toddlees (approved):
Works great

PR #3851, opened 2026-04-22
feature/2956 - Add closeOnly input to Dialog so we can easily reuse the same structure. Update Select Report dialog to alternate based on available reports

sasha-dresden:
Issue [FECFILE-2956](https://fecgov.atlassian.net/browse/FECFILE-2956)


toddlees (approved):
looks good

PR #3856, opened 2026-04-22
Feature/2014 - Warn logged-in users when API service is unavailable

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2014

Related PRs: Dependent on [#3012](https://github.com/fecgov/fecfile-web-app/pull/3833)


lbeaufort (commented):
@Elaine-Krauss-TCG can you add a descriptive title  to this PR?

danguyf (approved):
I tested this every which way and there were a few times when it let me go to another page but in the vast majority of cases it warned me straight off BUT I still think it should trigger for all 5xx and status==0.

front-end/src/app/login/login/login.component.ts
```diff
@@ -25,18 +36,41 @@ export class LoginComponent implements OnInit, AfterViewChecked {
     this.loginDotGovAuthUrl = environment.loginDotGovAuthUrl;
 
     window.addEventListener('resize', this.updateScrollbarWidth);
+
+    this.store
+      .select(selectServiceAvailable)
```
toddlees: re-commenting suggestion from 3012:
observable solution can be replaced with signal effect.  this also eliminates the need for DestroyerComponent
```
constructor() {
    super();

    effect(() => {
      const available = this.serviceAvailable();
      if (available === false) {
        this.loginDialogVisible.set(true);
      }
    });
  }
```

front-end/src/app/layout/layout.component.html
```diff
@@ -72,3 +75,22 @@
 @if (useDynamicSidebar) {
   <button (click)="toggleSidebar()">Toggle Sidebar</button>
 }
+
+@if (router.url !== '/login' && serviceAvailable() === false) {
```
toddlees: Talked to @Elaine-Krauss-TCG.  We're going to use the same terms from the condition on line 8.

toddlees (changes requested):

toddlees (approved):

PR #3868, opened 2026-04-24
[RELEASE/Sprint-84] feature/2966 - Fix a few validation issues related to strings being used for dates

sasha-dresden:
Issue [FECFILE-2966](https://fecgov.atlassian.net/browse/FECFILE-2966)

This fixes a few of the wonky validation issues we were having. All of our validations were built with the date being either a date object or null/undefined. However, now they can be strings which are then turned into dates if possible. There were a few things crashing when they received a string like "04/01/20YY" I added some checks to make sure the string being passed is a full date string.
I also fixed an issue where the invalid date format was briefly displaying when the format was correct.

### DIFF
--- a/front-end/src/app/shared/components/calendar/calendar.component.html
+++ b/front-end/src/app/shared/components/calendar/calendar.component.html
@@ -3,9 +3,8 @@
   <p-datepicker
     #datepicker
     appendTo="body"
-    (onFocus)="calendarOpened = true"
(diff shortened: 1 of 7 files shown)


danguyf (approved):
The code is even cleaner!

I was able to reproduce each of Shelly's validation issues and then successfully test their solution here.

max-zaremba-tcg (approved):

PR #3873, opened 2026-04-27
feature/2987 - Update layout for fixed sidebar.

sasha-dresden:
Issue [FECFILE-2987](https://fecgov.atlassian.net/browse/FECFILE-2897)

### DIFF
--- a/front-end/src/app/layout/header/header.component.html
+++ b/front-end/src/app/layout/header/header.component.html
@@ -1,6 +1,18 @@
 <nav class="w-full flex justify-content-between align-items-center py-0" aria-label="Navigation Bar">
   <div class="navbar-brand pb-0">
-    <img ngSrc="assets/img/seal-and-title.svg" class="seal-and-title mr-2" alt="FEC Seal" height="48" width="227" />
+    @if (useDynamicSidebar) {
+      <img
+        ngSrc="assets/img/seal-and-title.svg"
+        class="seal-and-title mr-2"
+        alt="FEC Seal"
+        height="48"
+        width="227"
+        (click)="toggleSidebar()"
+        (keydown.enter)="toggleSidebar()"
+      />
+    } @else {
+      <img ngSrc="assets/img/seal-and-title.svg" class="seal-and-title mr-2" alt="FEC Seal" height="48" width="227" />
+    }
   </div>
 
   <div class="collapse navbar-collapse justify-content-end" id="navbarTogglerDemo01">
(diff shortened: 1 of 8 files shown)


danguyf (changes requested):
The fixed sidebar looks great. I like using the seal to toggle the sidebar but wonder if Austin can come up with some sort of UX hinting.

I was not able to get it to display a scrollbar in the leftnav on Chrome.

sasha-dresden:
> The fixed sidebar looks great. I like using the seal to toggle the sidebar but wonder if Austin can come up with some sort of UX hinting.
> 
> I was not able to get it to display a scrollbar in the leftnav on Chrome.

I totally forgot about the vertical scrollbar needing to appear when the vertical viewing area shrunk. Got that fixed.
As for the seal to toggle the sidebar. That's just available on local to help make testing a bit easier. I believe eventually they want that ability to toggle the sidebar, but it hasn't been done yet. I'd been using a little button at the very bottom of the screen but it kind of threw off the sidebar alignment due to the extra height at the bottom that shouldn't be there. So I just incorporated it into the seal as a dev tool on your local. I considered just taking it out, but since we will eventually be doing a toggleable sidebar, figured I could just leave it there. It's set to local only.

danguyf (changes requested):
Love it, and love how clean and simple the solution is.

My only quibble is that, judging from the final desired state image, there should be some padding at the bottom of the lefthand nav so that the bottom white line isn't sitting right on the browser edge.

danguyf (approved):

front-end/src/app/layout/header/header.component.html
```diff
@@ -1,6 +1,18 @@
...
+        ngSrc="assets/img/seal-and-title.svg"
+        class="seal-and-title mr-2"
+        alt="FEC Seal"
+        height="48"
+        width="227"
+        (click)="toggleSidebar()"
+        (keydown.enter)="toggleSidebar()"
+      />
+    } @else {
+      <img ngSrc="assets/img/seal-and-title.svg" class="seal-and-title mr-2" alt="FEC Seal" height="48" width="227" />
```
dheitzer: rather than duplicating the ` `and attributes, can we just use conditional click/keydown handlers? like 
`(click)="useDynamicSidebar ? toggleSidebar()"`

dheitzer (changes requested):
looks good - just one optional question

front-end/src/app/layout/header/header.component.html
sasha-dresden: Good point. I didn't want it to even appear clickable in the non-local environment, but we aren't doing any styling and if (click) evaluates to false nothing happens when you click. Updated.

dheitzer (approved):
looks good!

PR #3874, opened 2026-04-28
feature/3036 - [UI] Update Modal - Confirm new contact

max-zaremba-tcg:
Ticket link: [FECFILE-3036](https://fecgov.atlassian.net/browse/FECFILE-3036)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
@@ -22,6 +22,7 @@
   />
 </div>
 
-<p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
+<app-confirmDialog />
+<!-- <p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
 <p-confirmDialog key="childDialog" [style]="{ width: '450px' }" />
-<p-confirmDialog key="childDialog_2" [style]="{ width: '450px' }" />
+<p-confirmDialog key="childDialog_2" [style]="{ width: '450px' }" /> -->
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
@@ -22,7 +22,7 @@
   />
 </div>
 
-<app-confirmDialog />
+<!-- <app-confirmDialog /> -->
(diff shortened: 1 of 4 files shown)


front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
```diff
@@ -22,6 +22,7 @@
   />
 </div>
 
-<p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
+<!-- <app-confirmDialog /> -->
+<!-- <p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
 <p-confirmDialog key="childDialog" [style]="{ width: '450px' }" />
-<p-confirmDialog key="childDialog_2" [style]="{ width: '450px' }" />
+<p-confirmDialog key="childDialog_2" [style]="{ width: '450px' }" /> -->
```
danguyf: Can we delete rather than comment out?

danguyf (changes requested):
This works, my only question/quibble is that the Current State uses _Last, First_ and the Desired State uses _First Last_ whereas the implementation remains the former.

front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
```diff
@@ -5,5 +5,5 @@
   (confirm)="confirmOption()"
   (close)="cancelOption()"
 >
-  <p>{{ message() }}</p>
+  <p [innerHTML]="message()"></p>
```
dheitzer, line 8: @max-zaremba-tcg in general, I think we want to avoid using innerHTML especially when displaying direct user input (as we are [here](https://github.com/fecgov/fecfile-web-app/blob/7d6446fcb06d25fc71c970e2e9f2644275580b0f/front-end/src/app/shared/services/confirmation-wrapper.service.ts#L112)) due to possible cross site scripting (although Angular is supposed to have some additional safeguards against this).  

All that to say - do you think there is any way we can use interpolation for this?

toddlees: Talked to max and the alternatives aren't great.  https://blog.angular-university.io/angular-innerhtml/
https://angular.dev/best-practices/security#:~:text=value%20during%20sanitization.-,Sanitization%20example,link,-The%20following%20template
these lead me to believe we're ok in this case.

toddlees (approved):

PR #3878, opened 2026-04-28
Fix issue with second date input not grabbing first input

sasha-dresden:
Fixes an issue with the calendar where it was grabbing the first transaction date when trying to convert string to date of the second transaction in a double transaction.

### DIFF
--- a/front-end/src/app/shared/components/calendar/calendar.component.ts
+++ b/front-end/src/app/shared/components/calendar/calendar.component.ts
@@ -49,8 +49,7 @@ export class CalendarComponent {
     const control = this.control();
 
     if (control) {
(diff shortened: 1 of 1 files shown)


sasha-dresden:
The issue was that we were using `document.getElementById(this.fieldName())` and both transactions had the same fieldName, so getElementByID was returning the very first one found, hence why as soon as you touched anything in the second transaction's date, it tried to validate and forced it to the value of the first date.
I originally wanted to use `this.datePicker().inputFieldValue` but that value was keyed to the last "good" date value, so if you had 06/02/2026 and then tried to go back and edit it again and started deleting to 06/02/YYYY, the inputFieldValue would unfortunately still be set as 06/02/2026.

toddlees (approved):
works great!

PR #3880, opened 2026-04-28
feature/2947 - [UI] Update Modal - Confirm contact updates

max-zaremba-tcg:
Ticket link: [FECFILE-2947](https://fecgov.atlassian.net/browse/FECFILE-2947)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
@@ -21,7 +21,3 @@
     [transaction]="transaction()"
   />
 </div>
-
-<p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
-<p-confirmDialog key="childDialog" [style]="{ width: '450px' }" />
-<p-confirmDialog key="childDialog_2" [style]="{ width: '450px' }" />
--- a/front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
+++ b/front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
@@ -5,5 +5,5 @@
   (confirm)="confirmOption()"
   (close)="cancelOption()"
 >
-  <p>{{ message() }}</p>
+  <p [innerHTML]="message()"></p>
 </app-dialog>
--- a/front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
@@ -74,14 +74,9 @@ export class TransactionContactUtils {
     );
   }
 
-  static displayConfirmationPopup(
-    message: string,
-    confirmationService: ConfirmationService,
-    targetDialog: 'dialog' | 'childDialog' | 'childDialog_2' = 'dialog',
-  ): Promise<boolean> {
+  static displayConfirmationPopup(message: string, confirmationService: ConfirmationService): Promise<boolean> {
     return new Promise((resolve) => {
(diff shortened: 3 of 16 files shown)


front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
```diff
@@ -5,5 +5,5 @@
   (confirm)="confirmOption()"
   (close)="cancelOption()"
 >
-  <p>{{ message() }}</p>
+  <p [innerHTML]="message()"></p>
```
lbeaufort, line 8: @dheitzer had a question about using `[innerHTML]` in another PR: https://github.com/fecgov/fecfile-web-app/pull/3874#pullrequestreview-4190068215

danguyf (changes requested):
Two issues apart from what Laura already raised:

1. The new modal appears when I go to Contacts and edit one. The old modal is still used when I go into a transaction and change contact details. Based on the wording of the ticket I believe it should apply to the latter.
2. The desired state image shows that the changed information should be bolded.

front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.html
toddlees: https://github.com/fecgov/fecfile-web-app/pull/3874#discussion_r3181825742
we've determined it to be a good use of innerHTML

front-end/src/app/shared/components/transaction-type-base/transaction-contact.utils.ts
```diff
@@ -65,7 +65,7 @@ export class TransactionContactUtils {
       if (!value) {
         return `<li>Removed ${ContactFields[property as keyof typeof ContactFields].toLowerCase()}</li>`;
       }
-      return `<li>Updated ${ContactFields[property as keyof typeof ContactFields].toLowerCase()} to ${value}</li>`;
+      return `<li>Updated ${ContactFields[property as keyof typeof ContactFields].toLowerCase()} to <b>${value}</b></li>`;
```
danguyf: I think this is a case where we want to use `<strong>` over `<b>` because it imparts semantic meaning and not just visual boldness.  Screen readers should be clued in to emphasize this, for example.

danguyf (changes requested):
Visually the new modal looks great.  It shows up when editing a contact within a transaction -- awesome.

I discovered quite by accident, though, that when you:
1. edit a contact within a transaction and then 
2. click the Save button it displays the modal correctly and as expected BUT
3. if you click Cancel or "X" the modal out

then the old modal appears behind it!  Really wild.  Maybe the existing contactDialogDialog needs to be removed or something?

danguyf (approved):

sasha-dresden (approved):

front-end/src/app/reports/transactions/transaction-container/transaction-container.component.html
```diff
@@ -21,7 +21,3 @@
     [transaction]="transaction()"
   />
 </div>
-
-<p-confirmDialog key="dialog" [style]="{ width: '450px' }" />
```
sasha-dresden, line 25: Is there any reason you moved these from outside the TransactionContainer to the Main app? Also, something that could probably be improved. Just using a single Confirmation Dialog for all 3. I think with our new framework we can use just a single one. This would mean we have to send less HTML over to the front end. Currently with this setup we're sending five Confirm Dialogs to every single page.

sasha-dresden (changes requested):

sasha-dresden (approved):
Just cleaned up the keys we're no longer using.

PR #3888, opened 2026-04-30
feature/3090 - Log user out of any committee when they get to the create committee and select committee pages

sasha-dresden:
Issue [FECFILE-3090](https://fecgov.atlassian.net/browse/FECFILE-3090)


danguyf (approved):
This makes sense, and works!

toddlees (approved):
works great

PR #3896, opened 2026-04-30
feature/3082 - Fix the datepicker, which pops open after hitting close on a modal for dates outside the coverage dates

max-zaremba-tcg:
Ticket link: [FECFILE-3082](https://fecgov.atlassian.net/browse/FECFILE-3082)

### DIFF
--- a/front-end/src/app/shared/components/calendar/calendar.component.html
+++ b/front-end/src/app/shared/components/calendar/calendar.component.html
@@ -3,7 +3,6 @@
   <p-datepicker
     #datepicker
     appendTo="body"
-    (onClose)="validateDate()"
     (onBlur)="validateDate()"
     [inputId]="fieldName()"
     [name]="fieldName()"
--- a/front-end/src/app/shared/components/calendar/calendar.component.html
+++ b/front-end/src/app/shared/components/calendar/calendar.component.html
@@ -3,6 +3,7 @@
   <p-datepicker
     #datepicker
     appendTo="body"
+    (onClose)="validateDate()"
     (onBlur)="validateDate()"
     [showOnFocus]="false"
     [inputId]="fieldName()"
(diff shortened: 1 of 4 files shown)


front-end/src/app/shared/components/calendar/calendar.component.html
```diff
@@ -3,8 +3,8 @@
   <p-datepicker
     #datepicker
     appendTo="body"
-    (onClose)="validateDate()"
```
sasha-dresden, line 6: I'm pretty sure we still need this. If we don't validate on close, it doesn't do the date check to validate the date, it drops you into the text entry part of the input and you then have to manually blur the field by clicking outside of it or tabbing in order to trigger validation.

front-end/src/app/shared/components/calendar/calendar.component.ts
sasha-dresden: I think we need some additional code here. If the control already has a date value and your using the datepicker dropdown, every single time you interact with the dropdown it will trigger a control.setValue(date) because the currentValue is a valid date. I think we'd need something like validating that the date from the currentValue does not already equal the control date. Something like 
`!(control.value instanceof Date) || control.value.getTime() !== date.getTime()`

sasha-dresden: I think these two lines should only run when the control.setValue() code is run. With it running even when that doesn't happen you get some wierd flashing when you have an invalid text date with the warning already present and then blur the input. It very briefly clears the validation error message and then adds it back in.

sasha-dresden (changes requested):

danguyf (approved):
This is working in my testing.

front-end/src/app/shared/components/calendar/calendar.component.ts
sasha-dresden, line 54: Y'know this if condition seems kind of redundant now. Now that we have better handling for dates as incomplete strings I'm not sure we need to be setting the control to null. Also, as far as I can tell, with the way the placeholder/input mask works this should always be a string or valid date object. We only really need to tell it to convert to a date and then trigger updateValueAndValidity.

Not really necessary for this ticket, just realizing this is extraneous now.

sasha-dresden (approved):

PR #3900, opened 2026-05-01
Feature/2945 - Add Navigation pane to terms of service page

sasha-dresden:
Issue [FECFILE-2945](https://fecgov.atlassian.net/browse/FECFILE-2945)

This includes [PR3873](https://github.com/fecgov/fecfile-web-app/pull/3873) to include the sticky sidebar updates.


danguyf (changes requested):
The desired state images all show the main text area left-aligned rather than centered.  It doesn't make it clear what the max-width should be, though, so I think we might need to check with the UI/UX team.

dheitzer (approved):

sasha-dresden:
> The desired state images all show the main text area left-aligned rather than centered. It doesn't make it clear what the max-width should be, though, so I think we might need to check with the UI/UX team.

Ah, yea, so the issue was I was letting this use the same behavior as the regular sidebar but they wanted this to behave differently. So I updated to more closely match what they requested. One of the key differences is that this one, the sidebar never goes away..

danguyf (approved):

PR #3911, opened 2026-05-05
Feature/2993 - Feeback Overlay Update

sasha-dresden:
Issue [FECFILE-2993](https://fecgov.atlassian.net/browse/FECFILE-2993)

I also took care of some SVG icons I noticed that weren't rendering correctly in firefox and just generally cleaned up our svg icon usage.


danguyf (changes requested):
Visually this looks great and addresses those requirements.

The "Contact EFO" button isn't working for me, though.  (Chrome)

There is also a merge conflict to address.

toddlees (changes requested):
checked with mitchell and jonella, we're going to split out the other svgs to a different ticket.  this should stay scoped to the feedback window

@danguyf  the contact efo link works on mine.  what sort of failure do you see?

PR #3931, opened 2026-05-07
[Sprint 86] Feature/2707 - Add link to form 1 and fix layout

sasha-dresden:
Issue [FECFILE-2707](https://fecgov.atlassian.net/browse/FECFILE-2707)

### DIFF
--- a/front-end/src/app/reports/submission-workflow/submit-report.component.ts
+++ b/front-end/src/app/reports/submission-workflow/submit-report.component.ts
@@ -86,7 +86,7 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
   readonly getContinueUrl = injectRouteData<(report?: Report) => string | undefined | null>('getContinueUrl');
   readonly continueUrl = computed(() => this.getContinueUrl()?.(this.activeReport()) || '/reports/');
   readonly backUrl = computed(() => this.getBackUrl()?.(this.activeReport()) || '');
-  readonly form1Link = environment.form1m_link;
+  readonly form1Link = environment.form1_link;
 
   constructor() {
     super();
--- a/front-end/src/app/reports/submission-workflow/submit-report.component.ts
+++ b/front-end/src/app/reports/submission-workflow/submit-report.component.ts
@@ -86,6 +86,7 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
   readonly getContinueUrl = injectRouteData<(report?: Report) => string | undefined | null>('getContinueUrl');
   readonly continueUrl = computed(() => this.getContinueUrl()?.(this.activeReport()) || '/reports/');
   readonly backUrl = computed(() => this.getBackUrl()?.(this.activeReport()) || '');
+  readonly form1Link = environment.form1m_link;
 
   constructor() {
     super();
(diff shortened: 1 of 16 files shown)


danguyf (approved):

front-end/src/app/reports/submission-workflow/submit-report.component.ts
```diff
@@ -86,6 +86,7 @@ export class SubmitReportComponent extends FormComponent implements OnInit {
   readonly getContinueUrl = injectRouteData<(report?: Report) => string | undefined | null>('getContinueUrl');
   readonly continueUrl = computed(() => this.getContinueUrl()?.(this.activeReport()) || '/reports/');
   readonly backUrl = computed(() => this.getBackUrl()?.(this.activeReport()) || '');
+  readonly form1Link = environment.form1m_link;
```
toddlees: Looks like our form 1 link is not matching the reqs.  the acs say dev/stage/prod should point to prod efo where test should point to stage efo.  I will confirm with aurelia about dev.  i think that's supposed to point to stage efo

toddlees (changes requested):

front-end/src/app/reports/submission-workflow/submit-report.component.ts
toddlees: got verification from Aurelia, we'll want to have it match the AC:
 dev/stage/prod=prod efo
 test = stage efo
 
Note: this will impact the link on the committee account page.  that's being noted in the ticket

Finally, let's update the environment variable to `form1_link` rather than `form1m_link`

sasha-dresden: Ok, I took this opportunity to clean up the way we are handling our environment so they are much more consistent, where you simply override the parts that differ as opposed to declaring everything every time even when they are things that never change between environment

front-end/src/environments/environment.base.ts
```diff
@@ -0,0 +1,48 @@
+export const baseEnvironment = {
+  production: false,
+  appTitle: 'FECfile+',
+  fecApiUrl: 'https://api.open.fec.gov/v1/',
```
toddlees: can probably drop this one

front-end/src/environments/environment.base.ts
```diff
@@ -0,0 +1,48 @@
...
+  ffapiTimeoutCookieName: 'ffapi_timeout',
+  userCanSetFilingFrequency: true,
+  sessionIdCookieName: 'sessionid',
```
toddlees: sessionIdCookieName can go.

front-end/src/environments/environment.cloud.gov.test.ts
```diff
@@ -1,53 +1,12 @@
...
-  committee_data_source: 'test',
-  form1m_link: 'https://webforms.stage.efo.fec.gov/webforms/form1/index.htm',
-  showForm3: false,
-  showSchedF: false,
-  disableLogin: false,
-  fecSpec: 8.5,
-  showGlossary: false,
+  name: 'local',
   webForms: 'https://webforms.stage.efo.fec.gov',
-  whoCanUseLink: 'https://www.fec.gov/efiling/getting-started-fecfile-pilot',
```
toddlees, line 28: whoCanUseLink needs to be unique for test environment.  similar to dev/stage/prod vs test pattern

toddlees (changes requested):

toddlees (approved):

PR #3935, opened 2026-05-08
feature/2947 - [UI] Update Modal - Confirm contact updates

max-zaremba-tcg:
Ticket link: [FECFILE-2947](https://fecgov.atlassian.net/browse/FECFILE-2947)

### DIFF
--- a/front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.ts
+++ b/front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.ts
@@ -21,12 +21,6 @@ export class ConfirmDialogComponent {
     effect(() => {
       if (this.confirmation()) this.visible.set(true);
     });
-    effect(() => {
-      const confirmation = this.confirmation();
-      if (!this.visible() && confirmation) {
-        confirmation.reject?.();
-      }
-    });
   }
 
   cancelOption() {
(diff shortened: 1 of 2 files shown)


front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.ts
```diff
@@ -21,6 +21,12 @@ export class ConfirmDialogComponent {
     effect(() => {
       if (this.confirmation()) this.visible.set(true);
     });
+    effect(() => {
+      const confirmation = this.confirmation();
```
toddlees: hey so i was looking at this and wondering, why isn't the cancelOption() getting called to reject in the first place.  it looks like out dialog.component doesn't configure the close output like the confirm output.  so in the confirm-daillog.component.html, where we say `(close)="cancelOption()"` that close output never actually does anything.  I think we can add that to the dialog.component rather than always rejecting when the modal is not visible.  that could potentially cause some bad side effects if the user selected confirm

toddlees (changes requested):

front-end/src/app/shared/components/confirm-dialog/confirm-dialog.component.ts
toddlees: I've updated the confirm dialog component to listen to `reject` rather than `close` from the app-dialog.  That eliminates the need to fire reject manually here

toddlees (approved):

PR #3938, opened 2026-05-08
feature/3115 - Eliminate error message blip on date field

sasha-dresden:
Issue [FECFILE-3115](https://fecgov.atlassian.net/browse/FECFILE-3115)

### DIFF
--- a/front-end/src/app/shared/utils/schema.utils.ts
+++ b/front-end/src/app/shared/utils/schema.utils.ts
@@ -213,7 +213,12 @@ export class SchemaUtils {
             result['exclusiveMax'] = { exclusiveMax: error.params['limit'] };
           }
           if (error.keyword === 'pattern') {
-            result['pattern'] = { requiredPattern: error.params['pattern'] };
+            if (
+              error.params['pattern'] === '^[0-9]{4}-[0-9]{2}-[0-9]{2}$' &&
+              form.get(property)?.value === 'MM/DD/YYYY'
+            )
+              result['required'] = true;
+            else result['pattern'] = { requiredPattern: error.params['pattern'] };
           }
           if (error.keyword === 'enum') {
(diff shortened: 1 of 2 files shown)


front-end/src/app/shared/utils/schema.utils.ts
```diff
@@ -213,7 +213,12 @@ export class SchemaUtils {
             result['exclusiveMax'] = { exclusiveMax: error.params['limit'] };
           }
           if (error.keyword === 'pattern') {
-            result['pattern'] = { requiredPattern: error.params['pattern'] };
+            if (
+              error.params['pattern'] === '^[0-9]{4}-[0-9]{2}-[0-9]{2}$' &&
+              form.get(property)?.value === 'MM/DD/YYYY'
+            )
+              result['required'] = true;
```
toddlees, line 220: I don't think this is an assumption we can make for all pattern fields.  Some might not be required.  Is there some way we can address this closer to the component that causes the issue?

toddlees (changes requested):

front-end/src/app/shared/utils/schema.utils.ts
sasha-dresden: It's something we can make the assumption for all Date pattern fields. That's why it has to use that specific Pattern and have that specific value. This is the only way to resolve the issue at the validation level. The problem about doing it at the component level is it's already too late. It's already validated it as a pattern and then we check it after to correct as being required. That's the reason for the blip. 
HOWEVER, after doing some research I found that if I use a directive I can override the registerOnChange to set the value _as_ it's changing. This means we get to do it before it heads to the validator.  This method was significantly different so I made a new [PR](https://github.com/fecgov/fecfile-web-app/pull/3945).

PR #3939, opened 2026-05-08
feature/3116 - Update svg usage to be more consistent and work with firefox

sasha-dresden:
Issue [FECFILE-3116](https://fecgov.atlassian.net/browse/FECFILE-3116)

There were actually several SVGs not displaying properly in Firefox. The underlying problem was that Firefox is incredibly picky about SVGs compared to chrome. If you don't specify both the file AND the id (within the SVG) it doesn't display properly.
Additionally I tried to clean up our SVG usage. We had several icons which were duplicates. Additionally, for these icons I unified them so they all used the same viewbox size which should make sizing the icons a bit more consistent.
List of pertinent changes:
- Exclamation icon.
- Committee icon
- Sort Icon: Got rid of duplicate. Just one icon that gets rotated.
- Arrow Icon: Got rid of three duplicates. Just one icon now that gets rotated appropriately. Also unified usage. We were using arrows in three different ways. Now it is consistent.
- Last arrow/First Arrow: Merged into one icon that gets rotated appropriately.
- Double Arrow - Fit down to the same viewbox as the rest.

Some other small changes. 
- The first arrow on the table paginator had a different color than the rest of the buttons when disabled.


danguyf (approved):
Works in my testing. Love the reduced complexity and increased consistency.

toddlees (approved):

PR #3946, opened 2026-05-11
Feature/3108 - Set universal input height to 36px and input font-size to 14px

sasha-dresden:
Issue [FECFILE-3108](https://fecgov.atlassian.net/browse/FECFILE-3108)


toddlees (approved):
looking good

PR #3947, opened 2026-05-11
feature/3118 - Make splitbutton work with singleclick disable directive

sasha-dresden:
Issue [FECFILE-3118](https://fecgov.atlassian.net/browse/FECFILE-3118)


danguyf (approved):

toddlees (changes requested):
can you also disable the 'add another' button?  i'm able to double click it and save two transactions

toddlees (approved):
works great!

PR #3955, opened 2026-05-13
feature/3130 - Add ToUpper Directive to committee id input

sasha-dresden:
Issue [FECFILE-3130](https://fecgov.atlassian.net/browse/FECFILE-3130)


danguyf (approved):
Love that it forces the text ToUpper and doesn't just upper it in flight to the API.

PR #3961, opened 2026-05-14
feature/3053 - add msx/msa to f99 text codes

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3053

Related PRs: [API](https://github.com/fecgov/fecfile-web-api/pull/2041)


front-end/src/app/reports/f99/main-form/main-form.component.ts
```diff
@@ -67,7 +68,7 @@ export class MainFormComponent extends MainFormBaseComponent<Form99> implements
   });
   readonly isLoanAgreement = computed(() => this.documentType() === 'MSW');
   readonly showFilingFrequency = computed(() => {
-    return this.documentType() in textCodesWithFilingFrequencies;
+    return environment.fecSpec === 8.5 && this.documentType() in textCodesWithFilingFrequencies;
```
toddlees: I think this might be unnecessary

toddlees (changes requested):

front-end/src/app/reports/f99/main-form/main-form.component.ts
dheitzer: @toddlees in the AC's I saw

> And if the FEC Format version is set to 8.5, and the user selects MSM or MSR, only then they will be required to select a filing frequency of either (otherwise these options will not show):

toddlees: Good eye!  i've removed that portion from the ACs.  that was leftover from the ticket this was cloned off of

toddlees (approved):

PR #3962, opened 2026-05-15
feature/3130 - Fix ToUpperDirective so it also updates the associated control to keep in sync

sasha-dresden:
Issue [FECFILE-3130](https://fecgov.atlassian.net/browse/FECFILE-3130)


toddlees (changes requested):
looking into issue where typing in a candidate or committee id blips the error message where it didn't before:

https://github.com/user-attachments/assets/eb5619db-176a-4414-ac43-9860abb1b44e

toddlees (approved):
works great!

PR #3989, opened 2026-05-22
feature/3128 - Add SingleClickDirective to Search for Committee and Create Committee buttons

sasha-dresden:
Issue [FECFILE-3128](https://fecgov.atlassian.net/browse/FECFILE-3128)


danguyf (approved):
I stepped through this with a mouse click breakpoint and it looks good to me. Awesome that we already have the appSingleClick and the overlay in the codebase so this can be only a matter of setting these buttons to use it.

dheitzer (approved):

toddlees (approved):

PR #3990, opened 2026-05-22
feature/2882 - Add placeholder for Form24 name

sasha-dresden:
Issue [FECFILE-2882](https://fecgov.atlassian.net/browse/FECFILE-2882)


toddlees (changes requested):
looking good.  one issue i'm seeing is that you can now skip the 24/48 option.  I think if we disable the submit button until that's selected, we won't need an error message for it.  We can check with mitchell on that

danguyf (approved):

toddlees (changes requested):
we'll want to add the 24/48hr field to enabling/disabling the submit button
Image

toddlees (commented):
wonderful!

toddlees (approved):

PR #3998, opened 2026-05-26
feature/2221 - Allow user to change coverage dates or report type for a Form 3 report in progress

sasha-dresden:
Issue [FECFILE-2221](https://fecgov.atlassian.net/browse/FECFILE-2221)

Update Form3 model to use same getMenuItems as Form3X.
Remove model index barrel causing circular import issue.


Elaine-Krauss-TCG (changes requested):
Changing a Form 3's Report Code (I.E, from Q1 to Q2) will result in a new report being created instead of the old one being updated.

PR #3999, opened 2026-05-27
Feature/2814 - [UI] Update Interest Rate and Due Date entry on Schedule C

max-zaremba-tcg:
Ticket link: [FECFILE-2814](https://fecgov.atlassian.net/browse/FECFILE-2814)


front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
```diff
@@ -306,10 +306,14 @@ export class TransactionFormUtils {
     }
 
     let formValues = SchemaUtils.getFormValues(form, transaction.transactionType?.schema, formProperties);
+    console.log('getFormValues', formValues);
     formValues = TransactionFormUtils.retrieveMemoText(transaction, form, formValues);
     formValues = TransactionFormUtils.addExtraFormFields(transaction, form, formValues);
+    console.log('addExtraFormValues', formValues);
     formValues = TransactionFormUtils.removeUnsavedFormFields(transaction, formValues);
 
+    console.log('Finished Formvalues', formValues);
```
Elaine-Krauss-TCG: Were you wanting to keep these in, or were these left-over debugging aids?  (I do that all the time, tbf)

Elaine-Krauss-TCG (approved):
This works great!  Everything looks good!

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.percent-icon {
+  position: absolute;
+  right: 0.75rem;
+  top: 50%;
+  transform: translateY(-50%);
+  pointer-events: none;
+  width: 1rem;
+  height: 1rem;
```
danguyf: The ACs say that the icon should be 14px x 14px whereas 1rem is 16px.  Even 14x14 looks larger than the Desired State images, though; by my eye it should be closer to 0.7rem -- all questions for the design team.

danguyf (changes requested):

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.ts
```diff
@@ -245,14 +234,14 @@ export class LoanTermsDatesInputComponent extends BaseInputComponent implements
   }
 
   get interestRateSettingField(): SubscriptionFormControl | null {
-    return this.form.get(this.templateMap['interest_rate_setting']) as SubscriptionFormControl;
+    return this.form.get(this.templateMap['loan_interest_rate_is_percent']) as SubscriptionFormControl;
   }
 
-  get interestRateSetting(): string {
+  get interestRateSetting(): boolean {
```
sasha-dresden, line 240: This seems to work just fine, however _technically_ this should be `boolean | null` since the initial state is null. Also it's wierd that the nullish coalescence goes to empty string when this is saying it should return a boolean (or null). So I'd think the return should be `this.interestRateSettingField?.value ?? null`
Same for get dueDateSetting(). Ugh I can't wait till Angular 22 gets here and we can just use native angular signals for the form control value so we don't have to do these sorts of hacky workarounds.

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.scss
```diff
@@ -0,0 +1,59 @@
+.percent-input-wrapper {
+  position: relative;
+  display: inline-block;
+  width: 11rem;
+  overflow: visible;
+
+  label {
```
sasha-dresden: You don't need this, there's no label inside your percent-input-wrapper

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.scss
```diff
@@ -0,0 +1,59 @@
+.percent-input-wrapper {
+  position: relative;
+  display: inline-block;
+  width: 11rem;
```
sasha-dresden: I'd probably use fit-content on the off chance we change that width

front-end/src/app/shared/components/inputs/loan-terms-dates-input/loan-terms-dates-input.component.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.label-wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+}
+
+.percent-input {
+  width: 11rem;
```
sasha-dresden: I feel like we should move this from being hard-coded to being a set value, so if we ever wanted to change the date field width it would update this as well. Like in variables.scss add something like: 
`--date-field-width: 11rem;`
and then use that variable here and in the calendar.component.scss

sasha-dresden (changes requested):

danguyf (approved):

sasha-dresden (approved):

toddlees (approved):
great!

PR #4002, opened 2026-05-28
Feature/416 + feature/2221 - Setup Create/Edit Form 3 to match style of Form 3X

sasha-dresden:
Issues
-  [FECFILE-416](https://fecgov.atlassian.net/browse/FECFILE-416)
- [FECFILE-2221](https://fecgov.atlassian.net/browse/FECFILE-2221)

APP [PR](https://github.com/fecgov/fecfile-web-api/pull/2065)


front-end/src/app/reports/f3-shared/create-shared-f3/create-shared-f3.component.spec.ts
```diff
@@ -58,17 +66,22 @@ describe('CreateF3XStep1Component: New', () => {
...
     expect(component).toBeTruthy();
+    expect(component.isF3X()).toBe(true);
+    expect(component.title()).toBe('Form 3X');
     expect(component.reportId()).toBeNull();
+
+    await fixture.whenStable();
+    fixture.detectChanges();
+
     expect(component.form.get('filing_frequency')?.value).toBe('Q');
-    // the next line is redundant given the one following it but ensures F3XA member is referenced in codebase for knip
```
toddlees, line 71: i think either we leave the comment in or if it's not necessary for knip we can take it out

front-end/src/app/reports/f3-shared/create-shared-f3/create-shared-f3.component.spec.ts
```diff
@@ -93,13 +106,12 @@ describe('CreateF3XStep1Component: New', () => {
...
   it('should not save and should dispatch singleClickEnableAction if form is invalid', async () => {
-    const createSpy = vi.spyOn(form3XService, 'create');
-    const updateSpy = vi.spyOn(form3XService, 'updateWithAllowedErrorCodes');
+    const createSpy = vi.spyOn(activeServiceInstance, 'create');
+    const updateSpy = vi.spyOn(activeServiceInstance, 'updateWithAllowedErrorCodes');
 
     await fixture.whenStable();
     fixture.detectChanges();
 
-    // now invalidate the form
```
toddlees, line 102: any reason for removing comments in this pr?

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -182,119 +194,120 @@ export class CreateF3XStep1Component extends FormComponent implements OnInit {
...
-    if (summary.report_code === ReportCodes.TER) {
-      summary.form_type = F3xFormTypes.F3XT;
+    if (this.form.invalid) {
+      printFormErrors(this.form);
+      this.store.dispatch(singleClickEnableAction());
+      return;
     }
 
+    const rawValues = SchemaUtils.getFormValues(this.form, this.activeSchema(), this.formProperties);
+    const summary = this.isF3X() ? Form3X.fromJSON(rawValues) : Form3.fromJSON(rawValues);
```
toddlees: I think we've got an opportunity to name this something like `report_data` or something.  We called it summary before because we were thinking about the summary page

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -326,39 +338,19 @@ export class CreateF3XStep1Component extends FormComponent implements OnInit {
...
 
   private checkDisableReportCode(reportCode: ReportCodes) {
     if (this.report()?.report_code === reportCode) return false;
     return this.usedReportCodes().includes(reportCode);
   }
 
-  /**
-   * FECFILE-2500
-   * ELECTION YEAR
-   *   current date is between Feb 1 – Dec 31 of an even-numbered year
```
toddlees, line 342: this seems like a pretty important comment

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
```diff
@@ -48,22 +57,32 @@ import { DialogComponent } from 'app/shared/components/dialog/dialog.component';
     SearchableSelectComponent,
     SaveCancelComponent,
     RouterLink,
-    Button,
+    ButtonModule,
     DialogComponent,
   ],
 })
-export class CreateF3XStep1Component extends FormComponent implements OnInit {
+export class CreateF3Step1Component extends FormComponent implements OnInit {
```
toddlees: these common `f3` and `form_3` bits are messing with my head because it's hard to tell what's a shared f3_ thing and what's a specific f3 thing.  but i'm struggling to come up with anything better than adding "base" to the names.

toddlees (changes requested):

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
sasha-dresden: Oh yea, that is an important one, Huh, the way it's evaluating the differences is a lot different than the way I did it. It seems to be comparing base-f3/create-workflow against the old f3x/create-workflow (which I suppose makes sense since it probably has more in common with that one than the f3 version) When I actually started from the f3/create-workflow and copied relevant stuff from the f3x version into it (and apparently forgot to copy function description comments... :shame:)

front-end/src/app/reports/f3-shared/create-shared-f3/create-shared-f3.component.spec.ts
sasha-dresden: Not really, I sometimes get comment removal happy when it's like, yea obviously that's what it's doing haha

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
sasha-dresden: yea, though TypeScript is camelCase so `reportData`. Also no reason to call  `summary`. Can just call it `report` since it is a report, or at least a report stub (maybe `reportStub` would be better?), at that point

front-end/src/app/reports/f3-shared/create-shared-f3/create-shared-f3.component.spec.ts
sasha-dresden: oh yea, good point, I did update this for knip, so I can remove this assert

front-end/src/app/reports/base-f3/create-workflow/create-f3-step1.component.ts
sasha-dresden: Yea I really don't know, maybe we should rename Form 3 to Form 3 Standard? But that doesn't seem like a great idea since it wouldn't match the Form name from FEC. 
Maybe components that work for all Form 3 variants could be something like one of these?
- UnifiedForm3CreateComponent
- UniversalForm3CreateComponent
- CreateForm3VariantComponent

toddlees (approved):

PR #4017, opened 2026-06-03
[Sprint 88] FECFILE-3157: Avoid E2E assertion timeouts in F3X aggregation

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3157

Related PRs:
N/A

* Updated `assertCalendarYtdFieldOnOpen` to use pattern already found in `waitForLoanBalanceRestoreByApi`
* Removed a flaky intermediary assertion -- we already assert the begin and end state
* Retrieve cypress configured `responseTimeout` and use it rather than default `defaultCommandTimeout` when waiting for data that arrives after a backend-driven refresh, not a straight DOM render.

### DIFF
--- a/front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
+++ b/front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
@@ -34,8 +34,7 @@ describe('Extended F3X Schedule E Aggregation', () => {
 
         F3XAggregationHelpers.openDisbursement(secondId);
         ContactLookup.getCandidate(result.candidateSenate, [], [], '#contact_2_lookup');
-        cy.blurActiveField();
-        F3XAggregationHelpers.assertCalendarYtdField('$50.00');
+        F3XAggregationHelpers.assertCalendarYtdAfterBlur('$50.00');
         F3XAggregationHelpers.clickSave();
 
         F3XAggregationHelpers.assertCalendarYtdFieldOnOpen(firstId, '$100.00');
@@ -72,9 +71,8 @@ describe('Extended F3X Schedule E Aggregation', () => {
         F3XAggregationHelpers.openDisbursement(secondId);
         PageUtils.selectDropdownSetValue('[inputid="electionType"]', 'G');
         F3XAggregationHelpers.clearAndType('#electionYear', `${currentYear}`);
-        cy.blurActiveField();
-        F3XAggregationHelpers.assertCalendarYtdField('$40.00');
         F3XAggregationHelpers.clickSave();
+        F3XAggregationHelpers.goToReport(result.report);
 
         F3XAggregationHelpers.assertCalendarYtdFieldOnOpen(firstId, '$80.00');
         F3XAggregationHelpers.clickSave();
--- a/front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
+++ b/front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
@@ -71,6 +71,7 @@ describe('Extended F3X Schedule E Aggregation', () => {
         F3XAggregationHelpers.openDisbursement(secondId);
         PageUtils.selectDropdownSetValue('[inputid="electionType"]', 'G');
         F3XAggregationHelpers.clearAndType('#electionYear', `${currentYear}`);
+        F3XAggregationHelpers.assertCalendarYtdAfterBlur('$40.00');
         F3XAggregationHelpers.clickSave();
         F3XAggregationHelpers.goToReport(result.report);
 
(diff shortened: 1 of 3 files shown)


front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
```diff
@@ -72,9 +71,8 @@ describe('Extended F3X Schedule E Aggregation', () => {
         F3XAggregationHelpers.openDisbursement(secondId);
         PageUtils.selectDropdownSetValue('[inputid="electionType"]', 'G');
         F3XAggregationHelpers.clearAndType('#electionYear', `${currentYear}`);
-        cy.blurActiveField();
-        F3XAggregationHelpers.assertCalendarYtdField('$40.00');
```
Elaine-Krauss-TCG, line 76: Is there a reason why we're removing this assertion instead of changing it to the new assertCalendarYtdAfterBlur?

front-end/cypress/e2e-extended/reports/f3x/f3x-aggregation.helpers.ts
```diff
@@ -882,9 +884,26 @@ export class F3XAggregationHelpers {
...
-    this.assertCalendarYtdField(expected);
+    return cy.wait(`@${alias}`).then(() => {
+      return cy.get('#calendar_ytd').should('have.value', expectedFormatted);
+    }) as unknown as Cypress.Chainable<void>;
+  }
+
+  static assertCalendarYtdAfterBlur(expectedFormatted: string): Cypress.Chainable<void> {
+    const responseTimeout = Number(Cypress.config('responseTimeout'));
+    cy.blurActiveField();
+    cy.waitForNetworkIdle(1000);
```
Elaine-Krauss-TCG, line 905: What a cool tool!  I didn't even know we had this plugin installed, tbh

front-end/cypress/e2e-smoke/pages/pageUtils.ts
```diff
@@ -394,10 +394,13 @@ export class PageUtils {
   }
 
   static switchCommittee(committeeId: string) {
+    cy.intercept('POST', `http://localhost:8080/api/v1/committees/${committeeId}/activate/`).as('ActivateCommittee');
     cy.intercept('GET', 'http://localhost:8080/api/v1/committee-members/').as('GetCommitteeMembers');
     cy.visit('/login/select-committee');
     cy.get('.committee-list .committee-info').get(`[id="${committeeId}"]`).click();
+    cy.wait('@ActivateCommittee');
     cy.wait('@GetCommitteeMembers'); // Wait for the guard request to resolve
+    cy.location('pathname', { timeout: 20000 }).should('not.include', '/login/select-committee');
```
Elaine-Krauss-TCG, line 403: Is retrieving the url taking a long time?

front-end/cypress/e2e-extended/reports/f3x/f3x-aggregation.helpers.ts
```diff
@@ -600,7 +600,9 @@ export class F3XAggregationHelpers {
 
   static rowLinkById(tableRoot: string, transactionId: string): Cypress.Chainable<JQuery<HTMLElement>> {
     this.assertTransactionId(transactionId, 'rowLinkById');
-    return cy.get(`${tableRoot} a[href*="/list/${transactionId}"]`).first().should('exist');
+    const timeout = Number(Cypress.config('responseTimeout'));
+    cy.get(tableRoot, { timeout }).should('exist');
+    return cy.get(`${tableRoot} a[href*="/list/${transactionId}"]`, { timeout }).first().should('exist');
```
Elaine-Krauss-TCG, line 605: Is the [defaultCommandTimeout](https://docs.cypress.io/app/references/configuration#Timeouts) not long enough?  We've got it set to 10 seconds in our config, and if it's taking long enough that setting it to 30s is required, then that seems like a problem in its own right.

danguyf: I am trying to balance not relying on timeouts on one hand with retaining a hard stop to prevent tests from endlessly hanging on the other.

In this case, the primary thing we're waiting on is not DOM rendering but network response so I think it's more correct to use responseTimeout than defaultCommandTimeout.

front-end/cypress/e2e-smoke/pages/pageUtils.ts
danguyf: Yes.  This ticket is a result of network slowness in PROD causing these two specs to fail inconsistently.

front-end/cypress/e2e-extended/reports/f3x/f3x-aggregation.helpers.ts
Elaine-Krauss-TCG: Ok, I see, that makes complete sense then!

Elaine-Krauss-TCG (approved):

front-end/cypress/e2e-extended/reports/f3x/aggregation-schedule-e.cy.ts
danguyf: I removed it because it was flaky and we were already asserting the beginning and end states, but I think that was before I added the `assertCalendarYtdAfterBlur()`.  Your suggestion is a good one so I've implemented it and so far it's passing in my testing.

toddlees (approved):

PR #4021, opened 2026-06-04
Feature/3080 - Fix issue with submit disable when switching between edit role and add user

sasha-dresden:
Issue [FECFILE-3080](https://fecgov.atlassian.net/browse/FECFILE-3080)

Fixes issue where switching from doing a successful edit user to an add user blocked submission.
Fixed by separating into their own  smaller dialogs. Originally edit/add user were very similar but now it's more like Add User / Edit Role. So we were doing a lot of form manipulation just to get them to validate the correct fields when there was literally only one or two fields to validate. Splitting into two components drastically simplified both the logic. As a byproduct it fixes the issue because now they are two completely separate forms. I also updated the submitDisabled to be directly tied to the form status.


front-end/src/app/committee/manage-committee/manage-committee.component.ts
```diff
@@ -34,14 +37,16 @@ export class ManageCommitteeComponent extends TableListBaseComponent<CommitteeMe
...
+
+  readonly editVisible = signal(false);
+  readonly editSelfVisible = signal(false);
 
   protected readonly rowActions: TableAction<CommitteeMember>[] = [
     new TableAction<CommitteeMember>('Edit Role', this.openEdit.bind(this), undefined),
     new TableAction<CommitteeMember>('Delete', this.confirmDelete.bind(this)),
   ];
   protected readonly editName: TableAction<CommitteeMember>[] = [
-    new TableAction<CommitteeMember>('Edit Name', this.openSelf.bind(this), undefined),
```
toddlees, line 44: nicely done! openSelf and editSelf was a little confusing

toddlees (approved):
Great!

PR #4024, opened 2026-06-05
Feature/3127 - System to show only relevant form types to committees [RAW]

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3127
  
Related PRs: [App](https://github.com/fecgov/fecfile-web-app/pull/4024), [API](https://github.com/fecgov/fecfile-web-api/pull/2084)

### DIFF
--- a/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.ts
+++ b/front-end/src/app/reports/form-type-dialog/form-type-dialog.component.ts
@@ -26,13 +26,7 @@ export class FormTypeDialogComponent {
   readonly formTypeOptions: ReportTypes[] = Array.from(getFormTypes(environment.showForm3), (mapping) => mapping[0]);
   readonly filteredOptions: Signal<ReportTypes[]> = computed(() => {
     return this.formTypeOptions.filter((type) => {
-      const eligible_report_types = this.eligibleReportTypes();
-      if (eligible_report_types) {
-        return eligible_report_types.includes(type);
-      }
-
-      console.error('No eligible report types on active committee');
-      return false;
+      return this.eligibleReportTypes().has(type);
     });
   });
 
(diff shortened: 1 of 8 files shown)


sasha-dresden (approved):
This looks good to me. The one suggestion I might make is to stub out the Processed to return all report types so that on the front end you don't have to have that piece of code commented out until the processed ticket is finished.

sasha-dresden (changes requested):
Oh one minor change, lint is failing because of unused import

Elaine-Krauss-TCG:
> When I tested this yesterday the only issue that I saw was that "Create report" was offering me both F3 and F3X as options, which Aurelia confirmed should not happen.
> 
> Today, the "Manage reports" page has a loading spinner that never goes away on initial load. If I refresh the page it is gone, but then when I click the "Create report" button the "Form type" picker says "No results found". This is not happening on your previous commit, [4eec4df](https://github.com/fecgov/fecfile-web-app/pull/4024/commits/4eec4dfdf5ca5ef98c20e1b640a1ff32c046bb7d).

As far as showing F3 and F3X goes, I haven't touched any of the mechanisms that hide F3 reports from users, and on DEV I see both F3 and F3X as options.  I don't _think_ there's a problem here, but I'm open to the possibility.

As to the spinner thing, I'll take a look

front-end/src/app/reports/form-type-dialog/form-type-dialog.component.ts
```diff
@@ -20,13 +22,27 @@ import { DialogComponent } from 'app/shared/components/dialog/dialog.component';
...
   readonly formTypeOptions: ReportTypes[] = Array.from(getFormTypes(environment.showForm3), (mapping) => mapping[0]);
+  readonly filteredOptions: Signal<ReportTypes[]> = computed(() => {
+    return this.formTypeOptions.filter((type) => {
+      const eligible_report_types = this.eligibleReportTypes();
+      if (eligible_report_types) {
+        return eligible_report_types.includes(type);
+      }
+
+      console.error('No eligible report types on active committee');
+      return false;
```
sasha-dresden: This section works perfectly fine, but I feel like the code is a bit more complicated than it needs to be. For one thing, committeeAccount.eligible_report_types should always be an empty array, but if we convert it to a Set here, even if it's undefined or null somehow, it would be the same as empty array, just a set with no entries, so we don't need that if statement to check it's existence.

`
readonly eligibleReportTypes = computed(() => new Set(this.committeeAccount().eligible_report_types));
  readonly filteredOptions: Signal<ReportTypes[]> = computed(() => {
    const eligibleReportTypes = this.eligibleReportTypes();
    return this.formTypeOptions.filter((type) => eligibleReportTypes.has(type));
  });
`

sasha-dresden (changes requested):
One teensy change, otherwise this looks great!

sasha-dresden (approved):

max-zaremba-tcg (approved):
Nice work!

danguyf (approved):
Confirmed that the endless spinner issue I was seeing before is no longer happening and that the picker fails gracefully when there are no eligible report types.

toddlees (approved):

PR #4033, opened 2026-06-09
Hotfix/turn on msx msa

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3063

Turns on MSX and MSA in F99 on production


dheitzer (approved):

sasha-dresden (changes requested):
Wouldn't this mean it's true in every environment, so it might make sense to just remove it entirely and delete everything that requires `IncludeMsaAndMsx` ? Like in `Form99.model.ts`

sasha-dresden (approved):
Actually, I don't think it'd be necessary, just a little bit of dead code sitting there since it would be true for ALL environments.

PR #4022, opened 2026-06-05
feature/2889 - new tooltip component

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2889

Related PRs:

Adding the following `app-tooltip` direction attributes will adjust the layout of the tooltip with respect to the element being targeted:
```
direction="topright"
direction="bottomright"
direction="topleft"
direction="bottomleft"
```

PR #4043, opened 2026-06-11
feature/2889 tooltip directive

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2889

Related PRs: [old component pr](https://github.com/fecgov/fecfile-web-app/pull/4022)

### DIFF
--- a/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
+++ b/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
@@ -19,7 +19,7 @@
             <span class="paren-label">{{ parenLabel() }}</span>
           }
         </span>
-        <span appTooltip="{{ memoItemHelpText() }}">
+        <span appTooltip="{{ memoItemHelpText() }}" vposition="bottom" hposition="right">
           <svg height="20" width="20">
             <use href="assets/img/i-info-circle.svg"></use>
           </svg>
--- a/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
+++ b/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
@@ -19,9 +19,11 @@
             <span class="paren-label">{{ parenLabel() }}</span>
           }
         </span>
-        <svg pTooltip="{{ memoItemHelpText() }}" height="20" width="20">
-          <use href="assets/img/i-info-circle.svg"></use>
-        </svg>
+        <div appTooltip="{{ memoItemHelpText() }}">
+          <svg height="20" width="20">
+            <use href="assets/img/i-info-circle.svg"></use>
+          </svg>
+        </div>
       </div>
       <app-error-messages
         [form]="form"
--- a/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
+++ b/front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
@@ -19,11 +19,11 @@
             <span class="paren-label">{{ parenLabel() }}</span>
           }
         </span>
-        <div appTooltip="{{ memoItemHelpText() }}">
+        <span appTooltip="{{ memoItemHelpText() }}">
           <svg height="20" width="20">
             <use href="assets/img/i-info-circle.svg"></use>
           </svg>
-        </div>
+        </span>
       </div>
       <app-error-messages
         [form]="form"
--- a/front-end/src/app/shared/directives/tooltip.directive.ts
+++ b/front-end/src/app/shared/directives/tooltip.directive.ts
@@ -1,4 +1,4 @@
-import { computed, Directive, effect, inject, input, } from '@angular/core';
+import { computed, Directive, effect, inject, input } from '@angular/core';
 import { Tooltip } from 'primeng/tooltip';
 
 @Directive({
@@ -12,8 +12,10 @@ export class TooltipDirective {
   readonly content = input<string>('', { alias: 'appTooltip' });
   readonly vposition = input<'top' | 'bottom'>('top');
   readonly hposition = input<'left' | 'right'>('right');
-  readonly tooltipStyleClass = computed(() => `app-tooltip app-tooltip-${this.vposition()} app-tooltip-${this.hposition()}`);
-  readonly positionLeft = computed(() => this.hposition() === 'left' ? -122 : 118);
+  readonly tooltipStyleClass = computed(
+    () => `app-tooltip app-tooltip-${this.vposition()} app-tooltip-${this.hposition()}`,
+  );
+  readonly positionLeft = computed(() => (this.hposition() === 'left' ? -122 : 118));
 
   constructor() {
     effect(() => {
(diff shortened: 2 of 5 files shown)


Elaine-Krauss-TCG (dismissed):
This is so great, and the code looks lovely, but the carat thingy is showing up twice on the inverted version that shows when the tooltip is rendered near the bottom of the screen.

Image

Also, despite doing a clean build of the front-end, the icon isn't showing up?  This might only be a problem on my end.

front-end/src/styles.scss
```diff
@@ -947,3 +947,50 @@ p-accordion-content[data-p-active='false'] > div.p-accordioncontent-content {
...
+  margin: inherit;
+}
+
+.app-tooltip-top .p-tooltip-arrow {
+  border-top-color: #112e51 !important;
+}
+.app-tooltip-top .p-tooltip-arrow::after {
+  border-top-color: white;
+  left: -8px;
+  bottom: 3px;
```
sasha-dresden: When I was looking at this 3px made the line look super thick due to overlap, 2px looked good though.

front-end/src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,28 @@
...
+  readonly tooltipStyleClass = computed(
+    () => `app-tooltip app-tooltip-${this.vposition()} app-tooltip-${this.hposition()}`,
+  );
+  readonly positionLeft = computed(() => (this.hposition() === 'left' ? -122 : 118));
+
+  constructor() {
+    effect(() => {
+      this.primeTooltip.setOption({ tooltipLabel: this.content() });
+      this.primeTooltip.setOption({ tooltipPosition: this.vposition() });
+      this.primeTooltip.setOption({ tooltipStyleClass: this.tooltipStyleClass() });
```
sasha-dresden: I known I gave you this format but I was messing around trying to figure out an issue and realized you could just do this which is a lot cleaner.
```
this.primeTooltip.setOption({
  tooltipLabel: this.content(),
  tooltipPosition: this.vposition(),
  tooltipStyleClass: this.tooltipStyleClass(),
  positionLeft: this.positionLeft(),
});
```

front-end/src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,28 @@
...
+  hostDirectives: [Tooltip],
+})
+export class TooltipDirective {
+  private readonly primeTooltip = inject(Tooltip);
+
+  readonly content = input<string>('', { alias: 'appTooltip' });
+  readonly vposition = input<'top' | 'bottom'>('top');
+  readonly hposition = input<'left' | 'right'>('right');
+  readonly tooltipStyleClass = computed(
+    () => `app-tooltip app-tooltip-${this.vposition()} app-tooltip-${this.hposition()}`,
```
sasha-dresden: Bug. PrimeNG tooltip has a feature that if you try to open a tooltip with a direction that doesn't fit it swaps it. So if the icon is all the way at the bottom of your screen and you have vposition as 'bottom' it is smart enough to realize this tooltip won't fit at the bottom and swap it to the top position instead. However, when this happens your app-tooltip-${this.vposition()} gets thrown off cause it tries to use styles for bottom when it's on the top. Instead, you can just remove that portion like this:
`readonly tooltipStyleClass = computed(() => `app-tooltip app-tooltip-${this.hposition()}`});`
And then in the styles.css update so it's looking at p-tooltip-top instead of app-tooltip-top (and vice versa for bottom)

front-end/src/styles.scss
sasha-dresden: These styles with app-tooltip-top and app-tooltip-bottom, if you look at my comment on the directive about the bug when primeNG flips it, make them p-tooltip-top and p-tooltip-bottom respectively

sasha-dresden: Also (and this is just a stylistic recommendation) I usually prefer to put substyles inside of their parent. So instead of 
```
.p-tooltip-top .p-tooltip-arrow {
  border-top-color: #112e51 !important;
}
.p-tooltip-top .p-tooltip-arrow::after {
  border-top-color: white;
  left: -8px;
  bottom: 3px;
}
```
You can do like this:
```
.p-tooltip-top .p-tooltip-arrow {
  border-top-color: #112e51 !important;
  &::after {
    border-top-color: white;
    left: -8px;
    bottom: 3px;
  }
}
```

front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
```diff
@@ -19,9 +19,11 @@
             <span class="paren-label">{{ parenLabel() }}</span>
           }
         </span>
-        <svg pTooltip="{{ memoItemHelpText() }}" height="20" width="20">
-          <use href="assets/img/i-info-circle.svg"></use>
-        </svg>
+        <span appTooltip="{{ memoItemHelpText() }}" vposition="bottom" hposition="right">
```
sasha-dresden, line 22: Man it's really annoying you can't apply this directly to an svg. Even standard primeNG tooltip doesn't behave correctly. The one suggestion I'd make and honestly this is entirely subjective, is to use <g> instead of <span> to stay in "svg" land and not run into and potential <span> related styling.

sasha-dresden (changes requested):

front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
dheitzer: @sasha-dresden yes was disappointed to see that even pTooltip directive direction does not behave properly with svg.  I'm not sure I followed your suggestion - was it to switch to img? This was something I had considered.

front-end/src/app/shared/directives/tooltip.directive.ts
dheitzer: great suggestion and sleuthing.  Updated with suggested fix.

front-end/src/app/shared/directives/tooltip.directive.ts
dheitzer: yep makes sense, good catch

front-end/src/styles.scss
dheitzer: I do like this more as well

front-end/src/app/shared/components/inputs/memo-code/memo-code.component.html
sasha-dresden: @dheitzer oh this was to switch to <g> like this:

```
<g appTooltip="{{ memoItemHelpText() }}" vposition="bottom" hposition="right">
  <svg height="20" width="20">
    <use href="assets/img/i-info-circle.svg"></use>
  </svg>
</g>
```
Here's a description of <g> from the mdn

> The `<g>` [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG) element is a container used to group other SVG elements.
Transformations applied to the `<g>` element are performed on its child elements, and most of its presentation attributes are inherited by its children. It can also group multiple elements to be referenced later with the `<use>` element.

sasha-dresden (approved):

toddlees (approved):

PR #4029, opened 2026-06-08
FECFILE-3169: disable F3 (but not F3X) transaction links

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3169

Related PRs:
N/A

We already had an `isTransactionDisabled`, sweet!

PR #4049, opened 2026-06-15
FECFILE-3169: framework for disabling specific transactions per report type

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3169

Related PRs:
https://github.com/fecgov/fecfile-web-app/pull/4029 (previous, blunt version)

<br />
Switched from disabling all F3 transactions to adding the general ability to disable one or more specific transaction types for a given report type.


sasha-dresden (approved):
This looks great! I like how you set it up as a Partial with the types set up and everything. Looks perfect. I'm not sure it really needs to be exported though, since it seems like it should only ever be used within that file. Feels like always having to break containment for unit tests, which feels wrong.

toddlees (approved):

PR #4050, opened 2026-06-15
feature/3167 - trigger contact dialog form validation on contact select

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3167

Related PRs:


danguyf (approved):
The fix works but I'm not clear if the test is testing the bug rather than the fix.

PR #4072, opened 2026-06-22
feature/3114 - Make Edit Form24 component from old Rename F24 Dialog

sasha-dresden:
Issue [FECFILE-3114](https://fecgov.atlassian.net/browse/FECFILE-3114)

API [PR2107](https://github.com/fecgov/fecfile-web-api/pull/2107)


danguyf (approved):
Works in my testing.  Contrary to the ACs, the Save/continue button is disabled entirely while the warning displays rather than appearing upon clicking the button but it's better this way.

Elaine-Krauss-TCG (approved):

front-end/src/app/reports/f24/form-24-edit/form-24-edit.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  private readonly validatedNames = new Set<string>();
+  private navigationIntent: 'save' | 'continue' = 'save';
+
+  readonly f24Form = form(
+    this.f24Model,
+    (schema) => {
+      required(schema.typeName, { message: 'This is a required field' });
+      required(schema.form24Name, { message: 'This is a required field' });
+      debounce(schema.form24Name, 300);
+      validateHttp(schema.form24Name, {
```
toddlees: This is a cool feature of this form schema setup.

Seems like we're making more calls to the api this way than we do with the create form 24 component.  What made you go this route instead of using the existing approach in the create form 24 component?

```if (this.form24Names.value()?.includes(this.fullName()))```

Also i think we've got this dangling legacy code now in form-24-service.ts:

```
export class F24UniqueNameValidator implements AsyncValidator {
...
```

File: front-end/src/app/reports/f24/form-24-edit/form-24-edit.component.ts [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+  private readonly messageService = inject(MessageService);
+  private readonly form24Service = inject(Form24Service);
+  private readonly apiService = inject(ApiService);
+  private readonly store = inject(Store);
+
+  private readonly activeReport = this.store.selectSignal(selectActiveReport);
+  private readonly report = computed(() => this.activeReport() as Form24);
+
+  private readonly f24Model = signal<Form24Data>({
+    typeName: null,
```
toddlees: this is the value rather than the name isn't it? like it's `24` instead of `24-Hour`.  it makes the line ```const name = `${typeName}-Hour: ${value()}`;``` a little tricky to understand

toddlees (changes requested):

front-end/src/app/reports/f24/form-24-edit/form-24-edit.component.ts
sasha-dresden: Ah yes, nothing better than removing dead code haha.
But as for the validateHttp(), I was trying to cleanup that F24UniqueNameValidator which in the old version of edit name was running on every blur, so I made something that would hit a little bit lighter. However, after talking with you, decided we could just prefetch the list once and run synchronous validation instead.

front-end/src/app/reports/f24/form-24-edit/form-24-edit.component.ts
sasha-dresden: Naming things is hard, ok? haha. But yea I agree this could be cleaned up a bit.
I changed it up so now the form consists of these two fields "type" and "typelessName". And then I set up two computations. One is the typeHour which gives us the "24-hour:" string, and the second is the fullName which uses that typeHour and the typelessName to build the full name that will be used by the system.

toddlees (approved):

PR #4077, opened 2026-06-23
FECFILE-3182: frontend-error-log messages logged as ERROR instead of INFO

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3182

Related PRs:
N/A

Thankfully, this little tweak is sufficient for Kibana to log it as ERROR even when it shows up in `access_log`.


dheitzer (approved):
nice find

toddlees (approved):

Issue #560, opened 2022-09-07
Allow user to download a .FEC file from the Manage Reports Page


AureliaKhorsand:
As a user, I will be able to generate and view a *.fec file of my report so that I can see how the Form 3X will output to FEC and troubleshoot as needed.

Acceptance Criteria 

  Dev Notes

*   [ ] set up record tied to report to track state of .fec compilation (should be akin to webprintsubmission and uploadsubmission records)
*   [ ] query reports to see status of dot_fec_task for each report, show download rows for each report that has a task

The grey block is a drawer that slides up from the bottom.

1\. "Your downloads"

  font: karla regular

  hex: #112e51

  size: 16pt

2\. "schedule_a-2022-09-14..." / [name of file downloading] 

  font: karla regular

  hex: #112E51

  size: 12pt

3\. Drawer background color: #F1F1F1

![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/cecd2fe9-3012-480a-8b33-b7db3239ab3c](https://github.com/fecgov/fecfile-web-app/assets/25588182/cecd2fe9-3012-480a-8b33-b7db3239ab3))

  ***Above: The user is waiting for their download to process. The 3 dots will be animated, a .gif borrowed from FEC.***

![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/7756361b-a4e3-4d6b-a1b3-60ed2b9d2c72](https://github.com/fecgov/fecfile-web-app/assets/25588182/7756361b-a4e3-4d6b-a1b3-60ed2b9d2c7))

  ***Above: The download has completed and the 'download' button is available. The 3-dots gif has changed to a checkmark.***

![Image](https://github.com/fecgov/fecfile-web-app/assets/25588182/da84db89-2f03-4fb8-99ca-b9e5d90d38d4](https://github.com/fecgov/fecfile-web-app/assets/25588182/da84db89-2f03-4fb8-99ca-b9e5d90d38d))

  ***Above: Multiple downloads show on the same screen. They keep stacking***

[https://images.zenhubusercontent.com/306118986/854f1e70-ba3a-4311-bea1-97abf05d1a30/exportvideo.mov](https://images.zenhubusercontent.com/306118986/854f1e70-ba3a-4311-bea1-97abf05d1a30/exportvideo.mov)

### QA Notes ###
 
 null

### DEV Notes ###
 
 null

### Design ###
 
 **Given** a report in any status
[...]

PR #4084, opened 2026-06-24
feature/2680 - Update *.fec download tray to expand on click

sasha-dresden:
Issue [FECFILE-2680](https://fecgov.atlassian.net/browse/FECFILE-2680)


dheitzer (changes requested):
It looks like the loading UI doesn't look quite like the wireframe:

Image

vs. wireframe:
image

it looks like at least the wireframe text doesn't match the text in the [original ticket](https://github.com/fecgov/fecfile-web-app/issues/560), so maybe that's not an issue but not sure about the rest of the loading UI.

danguyf (approved):
This is so cool. I love it.

(Side note: I think that the ticket title does not accurately reflect the ask.)

dheitzer (approved):

PR #4085, opened 2026-06-24
feature/3046 - [Schedule B] Update logic and text for purpose descriptions based on memos

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3046

Related PRs: [validate](https://github.com/fecgov/fecfile-validate/pull/456) [api](https://github.com/fecgov/fecfile-web-api/pull/2111)

### DIFF
--- a/front-end/package-lock.json
+++ b/front-end/package-lock.json
@@ -13,7 +13,7 @@
         "@ngrx/store": "21.0.1",
         "@primeuix/themes": "2.0.3",
         "class-transformer": "0.5.1",
-        "fecfile-validate": "https://github.com/fecgov/fecfile-validate#1d2f1310492f3841e06312fa674d1f9cb6ed7857",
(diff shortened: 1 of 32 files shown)


danguyf (dismissed):
I had the strangest time testing this.  Initially I was getting errors from the API about the memo text not matching expect whenever I tried to create any transactions; rebuilding the API with `--no-cache` didn't help but clearing the browser cache did.  Go figure.

Everything works now.  My only minor quibble is that when I click "Save & add memo", either when initially creating the SchedB transaction or when going back into it later, if there are no any memos currently, I get a brief error toast in the upper right saying (I think) that the memo doesn't match -- which is quickly replaced by the "Successful ... saved" toast and then loading the memo input page.  I think some users may be concerned by the error toast flashing by.

toddlees (approved):
works locally

PR #4087, opened 2026-06-25
feature/3209 - Make submit report status refresh on click for all reports. Also updated ReportStatus to be a const.

sasha-dresden:
Issue [FECFILE-3209](https://fecgov.atlassian.net/browse/FECFILE-3209)

### DIFF
--- a/front-end/src/app/reports/submission-workflow/submit-report-status.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report-status.component.html
@@ -2,7 +2,7 @@
 <hr type="solid" />
 <div class="flex">
   <div class="col-4 pl-0">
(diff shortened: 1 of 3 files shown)


danguyf (approved):
The fix follows the suggestion in the ticket comment.

Now when I first submit the report and the "Report status" says "pending" I can click on "REPORT STATUS" in the leftnav and the page refreshes to say "received", which is desirable.

What confuses me is that the "Steps to reproduce" in the ticket involve reviewing an already submitted report, which already says "Report submission received", and wanting to be able to click REPORT STATUS in the leftnav to refresh the page -- but nothing is going to change so what's the point?!  Nevertheless, I can see that clicking REPORT STATUS causes a number of API calls so I think the AC is met even though the page does not visibly refresh.

toddlees (approved):

PR #4111, opened 2026-07-01
feature/2941 - Save contact immediately when creating new one for transaction.

sasha-dresden:
Issue [FECFILE-2941](https://fecgov.atlassian.net/browse/FECFILE-2941)

This updates both of our contact dialog/modals and both of our contact lookups so that they all create a contact. (We really need to get back to unifying that process). The one nice thing about saving the contact immediately is we no longer need to keep track of excluding FecIds from our search because now all of them will have IDs


danguyf (approved):
I love how much cleaner this is.

From the ACs in the QA Notes I definitely tested #1 and #3 and I *think* I tested #2 and #4 but I'm not 100% confident that I understood what was intended.

dheitzer (changes requested):
looks good - do we need to update e2e tests though per the dev notes?

dheitzer (approved):

PR #4124, opened 2026-07-07
feature/2912 - Remove reattribute option from specified Schedule A receipts

max-zaremba-tcg:
Ticket link: [FECFILE-2912](https://fecgov.atlassian.net/browse/FECFILE-2912)


danguyf (approved):
Works in my testing.  Good refactor and consolidation.  I wonder if `canReattribute()` can be simplified a bit but getting it all in one place is a good first step.

front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts
```diff
@@ -20,6 +22,26 @@ export enum ReattRedesTypes {
...
+        ScheduleATransactionTypes.INDIVIDUAL_RECOUNT_RECEIPT,
+        ScheduleATransactionTypes.INDIVIDUAL_NATIONAL_PARTY_RECOUNT_ACCOUNT,
+        ScheduleATransactionTypes.INDIVIDUAL_NATIONAL_PARTY_HEADQUARTERS_ACCOUNT,
+        ScheduleATransactionTypes.INDIVIDUAL_NATIONAL_PARTY_CONVENTION_ACCOUNT,
+        ScheduleATransactionTypes.EARMARK_RECEIPT_FOR_RECOUNT_ACCOUNT_CONTRIBUTION,
+        ScheduleATransactionTypes.EARMARK_RECEIPT_FOR_CONVENTION_ACCOUNT_CONTRIBUTION,
+        ScheduleATransactionTypes.EARMARK_RECEIPT_FOR_HEADQUARTERS_ACCOUNT_CONTRIBUTION,
+      ]);
+    }
+    return ReattRedesUtils.F3F3X_REATTRIBUTION_ALLOWED_TRANSACTION_TYPES;
```
sasha-dresden: So, we generally want to avoid big lists like this. We already have a property built in to test if reattribution is allowed on a transaction:
`get isReattributable()`

If you go into SchATransactionType we already have an isReattributable which handle negative amounts. The new clarification is that it should also block when ContactType is individual, so change it to this:
 ```
override get isReattributable() {
    return !this.negativeAmountValueOnly || this.contactTypeOptions === INDIVIDUAL;
  }
```

**Review by dheitzer — CHANGES_REQUESTED** [review]
I'm wondering if we want to consider combining the 3 functions that determine if you can reattribute a function somehow so that there isn't confusion in how to restrict/allow something to be reattributed.  These 3 functions are:

`canReattribute`
`isReattributable`
`getF3F3XReattributionAllowedTransactionTypes`

I am struggling to see if these 3 are needed to be separate or not or if they could be somehow combined?

File: front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts [hunk 2]
```diff
@@ -41,6 +63,44 @@ export class ReattRedesUtils {
...
+        return false;
+      }
+    }
+
+    const transactionWithOptionalFields = transaction as TransactionListRecord & {
+      entity_type?: string;
+      amount?: number | string;
+    };
+
+    const entityType = 'entity_type' in transaction ? transactionWithOptionalFields.entity_type : undefined;
```
sasha-dresden: we don't need to do any of this because we get this answer from our isReattributable() function.
I do like this canReattribute function but we could probably simplify it a lot.
```
public static canReattribute(transaction: TransactionListRecord): boolean {
    return (
      !transaction.parent_transaction_id &&
      transaction.transactionType.isReattributable &&
      !ReattRedesUtils.isAtAmountLimit(transaction)
    );
  }
```

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -171,13 +171,9 @@ export abstract class TransactionListTableBaseComponent
...
         return (
-          transaction.transactionType.isReattributable &&
+          this.reportIsEditable() &&
           !transaction.parent_transaction_id &&
-          !ReattRedesUtils.isReattRedes(transaction, [
-            ReattRedesTypes.REATTRIBUTION_FROM,
-            ReattRedesTypes.REATTRIBUTION_TO,
-          ]) &&
-          !ReattRedesUtils.isAtAmountLimit(transaction)
+          ReattRedesUtils.canReattribute(transaction, this.report().report_type)
```
sasha-dresden: With the updates to canReattribute you would be able simplify this to:
(transaction: TransactionListRecord) => this.reportIsEditable() && ReattRedesUtils.canReattribute(transaction),

front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts
```diff
@@ -41,6 +63,44 @@ export class ReattRedesUtils {
...
+
+    const amount = 'amount' in transaction ? transactionWithOptionalFields.amount : undefined;
+    if (typeof amount === 'number' && amount < 0) return false;
+    if (typeof amount === 'string' && Number(amount) < 0) return false;
+
+    return true;
+  }
+
   public static isAtAmountLimit(transaction: Transaction | TransactionListRecord | undefined): boolean {
     const txn = transaction as SchATransaction | SchBTransaction;
```
sasha-dresden, line 53: @toddlees I think we have a bug here unrelated to this ticket. `isAtAmountLimit()` says it can accept `Transaction | TransactionListRecord | undefined` but actually the only thing being sent over is `TransactionListRecord` from the `TransactionListTableBaseComponent`. That's the only place this `isAtAmountLimit()` is being called from. The issue is that `TransactionListRecord` can not be successfully cast as  `SchATransaction | SchBTransaction` and will most importantly be missing the `reatt_redes_total` property. 
I tested this out by making a Party Receipt for one committee and then fully reattributing the total to another committee, and it still allows me to reattribute even though the full amount has been reattributed. Though to be fair you get in to the reattribution screen and it won't let you any amount greater than the current total which is 0.00
Another strange thing, I could reattribute an amount of 0.00 which seems odd. To double check we also allow you to create new transactions with an amount of 0 so I guess that is understandable, just strange.

sasha-dresden (changes requested):

front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts
toddlees: Great catch.  I'll start communicating this to the rest of the team

danguyf (approved):
Cleaner and tighter. I defer to Sasha.

sasha-dresden (approved):

dheitzer (approved):
looks nice

toddlees (approved):

front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts
toddlees: adding for any archeologists out there https://fecgov.atlassian.net/browse/FECFILE-3260
this was intended behavior, but we made a ticket to clarify for users

PR #4131, opened 2026-07-08
feature/3210 - [508] Fix (8) "Fieldset missing legend" on Submit Report Pages (Form 1M, 99, 24, 3X, 3)

max-zaremba-tcg:
Ticket link: [FECFILE-3210](https://fecgov.atlassian.net/browse/FECFILE-3210)

### DIFF
--- a/front-end/src/app/reports/submission-workflow/submit-report.component.html
+++ b/front-end/src/app/reports/submission-workflow/submit-report.component.html
@@ -19,10 +19,10 @@
       <hr type="solid" />
       <div class="row">
         <div>
-          <legend class="p-0 mb-2">
+          <div class="p-0 mb-2">
             <h2>Verify physical address</h2>
-            <span>HAS YOUR ADDRESS CHANGED IN THE LAST 10 DAYS?</span>
-          </legend>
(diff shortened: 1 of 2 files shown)


toddlees (commented):
This looks good.  I recall @sasha-dresden saying that this came up because these inputs are in fieldset.  I wonder if there's a way to not use fieldset to avoid the problem.  If that makes it more complicated, this solution works

sasha-dresden (changes requested):
So, this technically fixes the 508 issue, however the problem is that when this page was set up with `<fieldset>`, that's the actual problem.[ `<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) is actually designed to be used when grouping _multiple_ inputs together. So for example if you wanted to group email 1 and and email 2 together or group all the Address information together, those would be excellent uses of `<fieldset>` however the way we did was each individual input was wrapped in it's own `<fieldset>`. The better solution would be to just convert these to `<div>`, Then you won't need to add a hidden `<legend>`.

danguyf (approved):
This makes sense in light of Sasha's feedback.

sasha-dresden (approved):
Looks great!

PR #4132, opened 2026-07-08
feature/3234 - [508] Fix Missing form labels on Account info page

max-zaremba-tcg:
Ticket link: [FECFILE/3234](https://fecgov.atlassian.net/browse/FECFILE-3234)

### DIFF
--- a/front-end/src/app/app.component.html
+++ b/front-end/src/app/app.component.html
@@ -1,5 +1,7 @@
 <div>
-  <app-second-committee-admin-dialog />
+  @if (memberService.needsSecondAdmin()) {
+    <app-second-committee-admin-dialog />
+  }
   <app-poller />
   <router-outlet />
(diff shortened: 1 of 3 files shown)


sasha-dresden (changes requested):
Ah, this one is really tricky! So the issue actually is NOT with the submit report screen. The problem is that the `SecondCommitteeAdminDialogComponent` is always present and that has an email field in it. So every single page that has an email field now has that duplicate entry. The solution is to set the component to not exist when there is no need for it. So inside the AppComponent wrap it like this:
```
@if (memberService.needsSecondAdmin()) {
  <app-second-committee-admin-dialog />
}
```

This will resolve the issue here and everywhere else it is now happening.

danguyf (approved):

sasha-dresden (approved):

PR #4133, opened 2026-07-09
Feature/2532 - Set up Duplicate Contact Box for Individual

sasha-dresden:
Issue [FECFILE-2532](https://fecgov.atlassian.net/browse/FECFILE-2532)

I also converted ContactTypes to a const in our ongoing process of converting enums to consts.
I also updated our testContact to be more accurate to what type of contact it is. The old one was essentially EVERY contact type at once. Which led to some issues since an Individual would never have a `name` so some logic (like the `getNameString()`) would behave incorrectly because of that.


danguyf (changes requested):
This is really cool and functions mostly correctly in my testing.

Functional quibbles:
* The name search is case sensitive and I think it should be case insensitive for names that people might mess up like MacArthur vs Macarthur or Macbeth vs MacBeth.
* I think it would be better if it ran as the user typed (with debounce) rather than waiting until the user clicks off. If the user skips the Middle Name field and scrolls down to address they might miss that a duplicate was found.

UI notes:
* The name row has a height greater than its contents so even though the bottom-margin is only 8px as directed it visually has way more whitespace beneath it, making "No duplicate contacts found" look aligned bottom.  The ACs are matched but it doesn't match the wireframe. Not sure if the UI team wants to address this.
* The above also applies to the "found contact" vertical centering.
* The wireframes for the "found contact" box show more padding under it, that the right side aligns with the right edge of the Middle Name input box, and the shadow looks different -- especially to the left -- despite matching the ACs.

danguyf (approved):
Both functional quibbles addressed!

front-end/src/app/shared/models/contact.model.ts
```diff
@@ -102,6 +97,16 @@ export class Contact extends BaseModel {
   }
 }
 
+export function isEntity(type: ContactTypes) {
```
toddlees, line 100: These are awesome!

toddlees (approved):

PR #4141, opened 2026-07-10
FECFILE-2134: clone transactions

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2134

Related PRs:
N/A


front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -395,6 +403,34 @@ export abstract class TransactionListTableBaseComponent
...
+      message:
+        'This will create a copy of the selected transaction. The clone will use the same contact; itemization and aggregation will be determined by the system.',
+      acceptLabel: 'Acknowledge',
+      rejectLabel: 'Cancel',
+      accept: async () => {
+        try {
+          const clone = await this.transactionService.cloneSingleTransaction(String(transaction.id), this.reportId);
+          await this.router.navigateByUrl(`/reports/transactions/report/${this.reportId}/list/${clone.id}`);
+        } catch (error) {
+          console.error('Error cloning transaction:', error);
```
dheitzer: can we add the transaction.id and reportId to this log statement?

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -395,6 +403,34 @@ export abstract class TransactionListTableBaseComponent
...
+            summary: 'Unable to clone transaction',
+            detail: 'The selected transaction could not be cloned.',
+            life: 3000,
+          });
+        }
+      },
+    });
+  }
+
+  private canClone(transaction: TransactionListRecord): boolean {
```
dheitzer: can we make this public so we can have unit tests on it directly?

front-end/src/app/shared/services/transaction.service.ts
```diff
@@ -31,6 +32,11 @@ export class TransactionService {
     return getFromJSON(response);
   };
 
+  public isCloneable(transaction: CloneEligibilityTransaction | undefined): boolean {
+    // wrapper so we don't have to import/directly call the transaction-clone.utils in components
```
dheitzer, line 36: not sure I understand the necessity - Is there an issue with the aforementioned?

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+    !transaction.loan_id &&
+    !transaction.debt_id &&
+    !transaction.reatt_redes_id &&
+    !transaction.reattribution_redesignation_tag
+  );
+}
+
+export function buildClonedTransaction(source: ScheduleTransaction, reportId: string): Transaction {
+  if (!isCloneable(source)) {
+    throw new Error('FECfile+: transaction type is not eligible for cloning.');
```
dheitzer: can we log the transaction type identifier they are trying to clone to this error message?

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  const transactionTypeIdentifier = source.transaction_type_identifier as TransactionTypes;
```
dheitzer: Do we need this type assertion? The factory accepts a string.

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+  const transactionType = TransactionTypeUtils.factory(transactionTypeIdentifier);
+  const clone = transactionType.getNewTransaction();
+  const sourceCopy = getFromJSON(source.toJson());
```
dheitzer: I'm wondering if we can use [this function](https://github.com/fecgov/fecfile-web-app/blob/727c3ce17ff23a902245f20b93410dafc6b9b52b/front-end/src/app/shared/models/transaction.model.ts#L236) to clone this?

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+    clone.memo_text = MemoText.fromJSON(clone.memo_text.toJson());
+    clone.memo_text.id = undefined;
+    clone.memo_text.report_id = reportId;
+    clone.memo_text.transaction_id_number = undefined;
+    clone.memo_text.transaction_uuid = undefined;
+  }
+
+  clone.setMetaProperties(transactionType);
+
+  return clone;
```
dheitzer, line 135: I'm wondering if [this](https://github.com/fecgov/fecfile-web-app/blob/727c3ce17ff23a902245f20b93410dafc6b9b52b/front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts#L110) could be called instead of some of the above logic and/or refactor to reduce duplicating some of the logic (like copying the memo_text for example).  Thoughts?

dheitzer (changes requested):
just a few ideas

front-end/src/app/shared/services/transaction.service.ts
danguyf: I went back and forth on this and could be easily persuaded to go the other way.

The transaction service is already used everywhere that this comes into play so having this wrapper means I don't have to import `isCloneable()` from `transaction-clone.utils` in a bunch of places.  So it kept the `diff` smaller, some calls shorter, and keeps this UI code depending on one service.

Thoughts?

dheitzer: @danguyf yeah that definitely makes sense.  Thanks for the clarification! I'm good with this as-is.

front-end/src/app/shared/utils/transaction-clone.utils.ts
danguyf: `source.transaction_type_identifier` is `string | undefined` and the factory doesn't do `undefined` so I had to type it.  I could have typed it as string instead, granted.

dheitzer: if you wrap it in `if (transaction.transaction_type_identifier)` like we do [here](https://github.com/fecgov/fecfile-web-app/blob/c999f7d6c47bd4f70b9f81b3bc946d5ab3177660/front-end/src/app/shared/models/scha-transaction.model.ts#L59) it will.

It might be worth throwing an error in this case (although it should never happen).

front-end/src/app/shared/utils/transaction-clone.utils.ts
danguyf: Yes. I also updated the error message, though, because we get this error if there's any reason a transaction isn't cloneable, including the transaction type but not limited to it.  This error should never happen as the UI will not show "Clone" if the transaction isn't cloneable; better safe than sorry, though.

front-end/src/app/shared/utils/transaction-clone.utils.ts
danguyf: Love it.

front-end/src/app/shared/utils/transaction-clone.utils.ts
danguyf: I was initially opposed to that because the `if (!isCloneable(source))` above it catches the unknown `source.transaction_type_identifier` but on further reflection missing the type identifier is a whole other ball of wax that should get its own error call out, especially now that the transaction type is included in the error of `!isCloneable`, so I will stick that check at the top and then everything is happy, no more type assertion and one fewer constants declared.

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -395,6 +403,34 @@ export abstract class TransactionListTableBaseComponent
...
+  public cloneItem(transaction: TransactionListRecord): void {
+    this.confirmationService.confirm({
+      header: 'Clone transaction?',
+      message:
+        'This will create a copy of the selected transaction. The clone will use the same contact; itemization and aggregation will be determined by the system.',
+      acceptLabel: 'Acknowledge',
+      rejectLabel: 'Cancel',
+      accept: async () => {
+        try {
+          const clone = await this.transactionService.cloneSingleTransaction(String(transaction.id), this.reportId);
```
toddlees: This makes a lot of sense.  One issue is that the transaction is not supposed to be created until saving on the follow-up screen
> NOTE: The transaction is not created when the user clicks clone. It’s opening up a new transaction that will be pre-populated with the information from the original transaction. 

I think you'll be able to use a lot of your code if you navigate them to a page like 

`/reports/transaction/report/${this.reportId}/create/${transaction.transaction_type_identifier}?clone=${transaction.id}`.  and then use the `transaction.resolver.ts` under a condition like `if (clone) {`to populate the form like we do with reattributions and redesignations

toddlees (changes requested):
wicked tests

front-end/src/app/shared/utils/transaction-clone.utils.ts
danguyf: Love this idea. Waffling slightly on my implementation thereof.

danguyf: After experimenting with a different implementation I'm sticking with this one.

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
danguyf: Riiiiiight.

Excellent suggestions.

front-end/src/app/shared/resolvers/transaction.resolver.ts
sasha-dresden, line 31: Hey I noticed there was this complexity notice here. Honestly this doesn't seem _that_ complex, I think it just doesn't like having all these if/else. I think adding that one extra if for clone pushed it over the limit haha. I was messing around and here's a version that splits it into two functions and makes that if/else structure a bit more like a

```
async resolve(route: ActivatedRouteSnapshot): Promise<Transaction | undefined> {
    const transactionId = route.paramMap.get('transactionId');
    if (transactionId) return this.resolveExistingTransactionFromId(transactionId);

    const reportId = route.paramMap.get('reportId');
    const transactionTypeName = route.paramMap.get('transactionType');

    if (
      !reportId ||
      !transactionTypeName ||
      isTransactionTypeDisabledForReport(this.report().report_type, transactionTypeName)
    ) {
      return undefined;
    }

    return this.resolveNewTransactionFlow(route, reportId, transactionTypeName);
  }

  private async resolveNewTransactionFlow(
    route: ActivatedRouteSnapshot,
    reportId: string,
    transactionTypeName: string,
  ): Promise<Transaction | undefined> {
    const parentTransactionId = route.paramMap.get('parentTransactionId');
    if (parentTransactionId) {
      const parentTransaction = await this.service.get(String(parentTransactionId));
      return this.getNewChildTransaction(parentTransaction, transactionTypeName);
    }

    const resolverMap: Record<string, (id: string) => Promise<Transaction | undefined>> = {
      debt: (id) => this.resolveNewRepayment(id, transactionTypeName, 'debt'),
      loan: (id) => this.resolveNewRepayment(id, transactionTypeName, 'loan'),
      clone: (id) => this.resolveNewClone(reportId, id),
      reattribution: (id) => this.resolveNewReattribution(reportId, id),
      redesignation: (id) => this.resolveNewRedesignation(reportId, id),
    };

    for (const [key, resolverFn] of Object.entries(resolverMap)) {
      const id = route.queryParamMap.get(key);
      if (id) return resolverFn(id);
    }

    return this.resolveNewTransaction(reportId, transactionTypeName);
  }
```

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,254 @@
...
+}
+
+let cloneableTransactionTypes: Set<TransactionTypes> | undefined;
+
+function getCloneableTransactionTypes(): Set<TransactionTypes> {
+  if (cloneableTransactionTypes) {
+    return cloneableTransactionTypes;
+  }
+
+  cloneableTransactionTypes = new Set<TransactionTypes>([
```
sasha-dresden: @toddlees what are your thoughts on this list? I know we've, in the past, put this sort of information on the TransactionType. For a recent example, isReattributable. On one hand it makes a lot of sense for this to be a property of the TransactionType, but on the other hand they tend to give us a massive list of which ones should be reattributable or cloneable so keeping that list together also makes sense. On the other other hand, massive lists are a pain in the butt (plus means we need to import these types into the util which can lead to massive import dependency bloat), but so is having the data spread across 50 different files.

sasha-dresden: Oh I just though, another thing is that putting it on the TransactionType itself we would remove the need for the isCloneableTransactionType() function

danguyf: Re: your original comment, fully cognizant of the recent switch from massive list to `isReattributable()`, I was waiting to see if Todd would direct me to do something similar.  I waffled back and forth for the reasons you outlined but went with a giant list in the end because I thought it made the intent clearer than spreading it across dozens of files would.

front-end/src/app/shared/resolvers/transaction.resolver.ts
danguyf: Love the refactoring of all of those `if`s into a single `for` loop!

danguyf: I stepped through this and played around with some things and ultimately landed on what I expect is your exact code.  That did indeed get rid of the Sonar complexity complaint.

front-end/src/app/shared/utils/transaction-clone.utils.ts
toddlees: This topic has come up quite a few times recently.  I think it'll be useful to discuss it in respect to other lists or other properties that are defined at a transactiontype level.   Things i'm thinking about are how many lists are there vs how many transactions are there? how similar are some of these lists?  How can we reasonably anticipate these lists changing in the future?  I think those questions will help us decide what's best

danguyf: Sold.

front-end/src/app/shared/utils/transaction-clone.utils.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function isCloneableTransactionType(
+  transactionTypeIdentifier: string | undefined,
+): transactionTypeIdentifier is TransactionTypes {
+  if (!transactionTypeIdentifier) {
+    return false;
+  }
+
+  const transactionType = TransactionTypeUtils.factory(transactionTypeIdentifier);
```
sasha-dresden: I don't think you need to go through this whole rigmarole to get the TransactionType. You're using either a fully formed Transaction or a TransactionListRecord, both of which have a fully formed TransactionType on them, so there isn't any need to build the transactionType via the identifier. I created a [PR](https://github.com/fecgov/fecfile-web-app/pull/4166) into your branch demoing it. It seems to work fine from my testing, though maybe you had a very specific reason for using the transaction_type_identifier I didn't see.

danguyf: This is undeniably cleaner. I put it through its paces and everything looks good to me, so I've merged it in and will do a little more testing.

sasha-dresden (approved):

PR #4163, opened 2026-07-17
feature/3221&3222 - Add candidate and committee duplicate checks

sasha-dresden:
Issue [FECFILE-3221](https://fecgov.atlassian.net/browse/FECFILE-3221)
Issue [FECFILE-3222](https://fecgov.atlassian.net/browse/FECFILE-3222)

### DIFF
--- a/front-end/src/app/shared/components/contact-dialog/duplicate-contact/duplicate-contact.component.ts
+++ b/front-end/src/app/shared/components/contact-dialog/duplicate-contact/duplicate-contact.component.ts
@@ -1,14 +1,23 @@
 import { Component, computed, input, output, signal } from '@angular/core';
-import { Contact, ContactTypes, isEntity } from 'app/shared/models/contact.model';
+import { Contact, ContactTypes, hasFecId } from 'app/shared/models/contact.model';
 import { StatePipe } from '../../../pipes/state.pipe';
 import { ButtonModule } from 'primeng/button';
 import { ProgressSpinnerModule } from 'primeng/progressspinner';
 
-interface Names {
+function toFullName(data: { last_name?: string; first_name?: string }): string {
+  return data.last_name && data.first_name ? `${data.last_name}, ${data.first_name}` : '';
+}
+
+interface ValidatingFields {
+  name: string;
   first_name: string;
   last_name: string;
-  name: string;
+  candidate_id: string;
+  committee_id: string;
 }
+
+const blankData = { first_name: '', last_name: '', name: '', candidate_id: '', committee_id: '' };
+const propertyMap = { CAN: 'candidate_id', COM: 'committee_id', ORG: 'name' } as const;
 
 @Component({
   selector: 'app-duplicate-contact',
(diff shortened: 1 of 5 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/shared/components/contact-dialog/duplicate-contact/duplicate-contact.component.ts
```diff
@@ -19,55 +28,61 @@ interface Names {
...
-    return this.existingContacts().filter(
-      (c) => type === c.type && name === (isEntity(c.type) ? c.name : `${c.last_name}, ${c.first_name}`)?.toLowerCase(),
-    );
-  });
-  readonly validName = computed(() => {
-    const name = this.name();
-    const splitName = name.split(',').map((item) => item.trim());
-    return !splitName.includes('');
-  });
+  readonly potentialDuplicates = computed(() => this.existingContacts().filter((c) => this.checkForMatch(c)));
```
toddlees, line 34: So given how the f24 name check ticket [went](https://github.com/fecgov/fecfile-web-app/pull/4072#discussion_r3501636859) (where we made one call for all f24s because we can reasonably expect the number of f24s in a committee account to not venture out of the 10s) I see why you went the way you went with the name check to get all the contacts and filter on the front end. For contacts, we have reason to believe that the number will grow much more drastically, like 1000s and beyond.

This does not need to hold up this pr since our users haven't gotten to the point where they're making a ton of contacts, but it will come up. So i wanted to get your thoughts on it. Since there are so many more potential contacts, i think a debounced check is more appropriate for checking collisions in this case.  

If you think it's worth discussing in more depth or don't want to hold this ticket up, let's just go with this and make a follow up. But if you do want to make a collision endpoint and call it behind a debounce for this ticket, i think it'd be a good move

toddlees (approved):
This works great!

front-end/src/app/shared/components/contact-dialog/duplicate-contact/duplicate-contact.component.ts
sasha-dresden: oh I definitely agree with you, yet another point in favor of the contact refresh, but we could probably turn this into a lookup based similar to how we do the unique fec id, honestly, we might want to tie into that one so we aren't needing to run this twice essentially. I think if I had started with candidate/committee, I probably would've approached it that way from the start, though we had no such lookup for individual/organization so I'd need to make one for that.  Hmm, thinking about it, that would probably be a lot stronger long term, the one tricky thing was getting this to work with the right contact type at the right time which was a rather simple matter with the signal form version.

toddlees: you could have the `checkForMatch` send the correct url params to a `findDuplicates/?last=x&first=y&fec_id=z` 

I also noticed that the `get_contact_id/?fec_id` is not debounced.  It gets called every keystroke.  So if you're going to change that to call a new unified endpoint, you'll want to add a debounce.  otherwise i'll make a ticket for that too

toddlees (changes requested):
comment marking that we want the collision check to be a per value api call rather than comparing against a one time request for all contacts to the front end

PR #4171, opened 2026-07-20
feature/3111 - Update so we no longer have to excuse 'expected' 404s.

sasha-dresden:
Issue [FECFILE-3111](https://fecgov.atlassian.net/browse/FECFILE-3111)

API [PR2151](https://github.com/fecgov/fecfile-web-api/pull/2151)


max-zaremba-tcg (approved):

danguyf (approved):
I love how much simpler this makes things.

front-end/src/app/shared/services/api.service.ts
```diff
@@ -38,34 +38,9 @@ export class ApiService {
   }
 
   public get<T>(endpoint: string, params?: QueryParams): Promise<T>;
-  public get<T>(endpoint: string, params?: QueryParams, allowedErrorCodes?: number[]): Promise<T>;
-  public get<T>(
-    endpoint: string,
-    params: QueryParams = {},
-    allowedErrorCodes?: number[],
-  ): Promise<T> | Promise<HttpResponse<T>> {
+  public get<T>(endpoint: string, params?: QueryParams): Promise<T>;
```
dheitzer, line 50: I think we can remove this duplicate declaration?

dheitzer (changes requested):
one minor request

front-end/src/app/shared/services/api.service.ts
sasha-dresden: Oh good catch, I'd cleaned it up removing the allowedErrorCodes, but with that gone we now have two of the same. Odd that it didn't complain about duplicate declarations, glad you caught that!

Elaine-Krauss-TCG (approved):

dheitzer (approved):

PR #4186, opened 2026-07-23
feature/3168 - [Form 3] Election state of special reports default to candidate state 

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3168

Related PRs: [api](https://github.com/fecgov/fecfile-web-api/pull/2160)

### DIFF
--- a/front-end/src/app/committee/select-committee/select-committee.component.ts
+++ b/front-end/src/app/committee/select-committee/select-committee.component.ts
@@ -54,8 +54,8 @@ export class SelectCommitteeComponent {
   }
 
   async activateCommittee(committee: CommitteeAccount): Promise<void> {
-    await this.committeeAccountService.activateCommittee(committee.id);
-    this.store.dispatch(setCommitteeAccountDetailsAction({ payload: committee }));
+    const activatedCommittee = await this.committeeAccountService.activateCommittee(committee.id);
+    this.store.dispatch(setCommitteeAccountDetailsAction({ payload: activatedCommittee }));
     this.userService.getCurrentUser().then((userLoginData) => {
       this.store.dispatch(userLoginDataRetrievedAction({ payload: userLoginData }));
     });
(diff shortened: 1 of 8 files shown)


front-end/src/app/committee/select-committee/select-committee.component.ts
```diff
@@ -54,8 +54,8 @@ export class SelectCommitteeComponent {
   }
 
   async activateCommittee(committee: CommitteeAccount): Promise<void> {
-    await this.committeeAccountService.activateCommittee(committee.id);
-    this.store.dispatch(setCommitteeAccountDetailsAction({ payload: committee }));
+    const activatedCommittee = await this.committeeAccountService.activateCommittee(committee.id);
+    this.store.dispatch(setCommitteeAccountDetailsAction({ payload: activatedCommittee }));
```
sasha-dresden, line 58: So there's a slight bug here. The committee data returned from activatedCommittee is missing some information we had from the committee returned via `/committees/` that endpoint included a list of eligible report types. Honestly, it would make more sense to have eligible_report_types only included in the activated committee rather than the full committee list since we don't need that information until you are working with the committee

sasha-dresden (changes requested):
So this works g

front-end/src/app/committee/select-committee/select-committee.component.ts
danguyf, line 1: I expect this is dovetailing on Sasha's comment. On this branch I can't select a committee account; weirdly I don't get any errors (that I could find) in console, app logs, or API logs.  If I checked out just this file from `develop` then everything else in this branch works as desired.

danguyf (changes requested):

front-end/src/app/committee/select-committee/select-committee.component.ts
dheitzer: this is a great catch @sasha-dresden .  I have addressed this and we are now sending the full committee (including eligible_report_types) in the activate endpoint response.

I analyzed removing retrieving the full committee data for the list endpoint, but since committee name comes from openfec it breaks the list display (showing just committee ids).

front-end/src/app/committee/select-committee/select-committee.component.ts
dheitzer: Good catch.  I have addressed @sasha-dresden 's comments and things seems to be working correctly now for me.

Elaine-Krauss-TCG (approved):

sasha-dresden (approved):

danguyf (approved):
Everything works for me now.

PR #4188, opened 2026-07-23
feature/3188 - Allow user to submit Form 99 without added text (except MST)

max-zaremba-tcg:
Ticket link: [FECFILE-3188](https://fecgov.atlassian.net/browse/FECFILE-3188)

Related PRs: [fecfile-web-api](https://github.com/fecgov/fecfile-web-api/pull/2174), [fecfile-validate](https://github.com/fecgov/fecfile-validate/pull/470)


front-end/src/app/reports/f99/main-form/main-form.component.ts
```diff
@@ -69,14 +69,33 @@ export class MainFormComponent extends MainFormBaseComponent<Form99> implements
...
+  }
+
   constructor() {
     super();
     effect(() => {
       this.documentType();
-      /**
-       * Reset the filing frequency field when the document type changes, but only after the initial load.
-       * On load we want the filing frequency from the data.
-       */
```
danguyf, line 79: I think this comment could still be useful above (what is now) line 99.

danguyf (changes requested):
This initially appears to work in so far as it removes the warning that text is required when you click Save.  Unfortunately, neither save button now works, as both calls return an HTTP400 with `{"message_text":["None is not of type 'string'"]}` from the API call.

I think it would be more optimal to handle this in the F99 schema in `fecfile-validate`, rather than overriding it here.  I think that might also avoid the not-saving issue.

front-end/src/app/reports/f99/main-form/main-form.component.ts
danguyf, line 1: My understanding of the how the fecfile-validate schemas are used by the frontend is vague at best so please take my comments with a grain of salt.

My hope was that the required field logic could be handled entirely in the fecfile-validate schema but now, seeing the additions here, I think I see why that is insufficient. If I'm understanding correctly, this new code forces the message_text validation to re-run when the text_code changes?

Would it be possible to have the "required" message appear as soon as the text_code changes rather than only showing once "Save" is clicked?  Or if that is not desired by the design team, I think we should at least clear the "required" message when the type is changed from Misc to something else. (This is not happening for me in current testing, even after rebuilding app and api containers completely.)

danguyf (changes requested):

danguyf (approved):
After an `npm ci` and clearing cache everything is working.  The "required" warning no longer shows up on edit so there is no need to reset it.

dheitzer (changes requested):
@max-zaremba-tcg I think the only change needed for the front-end for this would be to adjust the validate hash and add the following ngOnInit() (similar to yours but adjusted):

```
override ngOnInit(): void {
    super.ngOnInit();
    this.form
      .get('text_code')
      ?.valueChanges.pipe(takeUntil(this.destroy$))
      .subscribe(() => {
        this.form.get('message_text')?.updateValueAndValidity();
      });
  }
```

you shouldn't need to patch the form manually due to the changes you made in the validate repo.  Otherwise looks good

sasha-dresden (approved):

dheitzer (changes requested):
@max-zaremba-tcg you'll need to update the validate hash in package.json/package-lock.json and in api requirements.txt with your latest validate commit.

dheitzer (approved):

PR #4196, opened 2026-07-27
FECFILE-2134: Corrected UI/UX.

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2134

Related PRs:
https://github.com/fecgov/fecfile-web-app/pull/4141 (previous attempt)

The UI/UX ACs were not met by the previous PR. They are here.

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
@@ -1,7 +1,6 @@
-import { Component, computed, effect, inject, Signal, signal } from '@angular/core';
+import { Component, computed, effect, inject, Signal, linkedSignal } from '@angular/core';
 import { Store } from '@ngrx/store';
 import { Title } from '@angular/platform-browser';
-import { ActivatedRoute } from '@angular/router';
 import { isPulledForwardLoan, Transaction } from 'app/shared/models/transaction.model';
 import { ReattRedesTypes, ReattRedesUtils } from '../../../shared/utils/reatt-redes/reatt-redes.utils';
 import { selectActiveReport } from '../../../store/active-report.selectors';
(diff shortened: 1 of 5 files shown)


front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
```diff
@@ -55,5 +62,6 @@ export class TransactionContainerComponent {
 
   constructor() {
     effect(() => this.titleService.setTitle(this.transaction()?.transactionType?.title ?? ''));
+    this.showClonedTransactionDialog.set(!!this.activatedRoute.snapshot.queryParamMap.get('clone'));
```
sasha-dresden: So this works great for your requirements, however it kind of breaks things for my requirements on [FECFILE-3225](https://fecgov.atlassian.net/browse/FECFILE-3225) because in mine you stay on the same page essentially, so the constructor won't re-run, so you go from not a clone to a clone. Here's a way to have this be more reactive where it isn't set once in the constructor and then never touched again:

```
private readonly cloneParam = injectQueryParams('clone');
readonly showClonedTransactionDialog = linkedSignal(() => !!this.cloneParam());
```

sasha-dresden (changes requested):
Your changes work great, but there is one thing that'll cause issues when merged with my branch, I was trying to decide if I should just merge these into mine and then fix it there, but it seems like it'd probably be best to have it set correctly from the get-go.

front-end/src/app/reports/transactions/transaction-container/transaction-container.component.ts
danguyf: Done!

sasha-dresden (approved):

dheitzer (approved):

PR #4203, opened 2026-07-28
FECFILE-3294: Don't disassociate on reattr/redes

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3294

Related PRs:
N/A

* Updated `getPayloads()` to check `to.report_ids` and `reattRedes.report_ids` for values and set both to those values.
* Updated `getPayloadTransaction()` to defensively set `report_ids` to `activeReportId` if `report_ids` is empty. This is not necessary for the fix and I'm not sure if it's a good idea -- maybe it'd be better to fail loudly if/when encountering another situation where `report_ids` is empty.
* Added some tests for the above.

### DIFF
--- a/front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
+++ b/front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
@@ -357,6 +357,9 @@ export class TransactionFormUtils {
     const secondaryReportId = form.get('linkedF3xId')?.value;
     if (secondaryReportId) {
       payload['report_ids'] = [activeReportId, secondaryReportId];
+    } else if (!payload.report_ids?.length) {
+      // If the payload doesn't have any report_ids, defensively add the active report id to it
+      payload.report_ids = transaction.report_ids?.length ? [...transaction.report_ids] : [activeReportId];
     }
 
     return payload;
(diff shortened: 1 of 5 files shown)


front-end/src/app/shared/components/transaction-type-base/transaction-form.utils.ts
```diff
@@ -357,6 +357,9 @@ export class TransactionFormUtils {
     const secondaryReportId = form.get('linkedF3xId')?.value;
     if (secondaryReportId) {
       payload['report_ids'] = [activeReportId, secondaryReportId];
+    } else if (!payload.report_ids?.length) {
+      // If the payload doesn't have any report_ids, defensively add the active report id to it
+      payload.report_ids = transaction.report_ids?.length ? [...transaction.report_ids] : [activeReportId];
```
danguyf, line 362: Good idea? Bad idea?

front-end/src/app/shared/utils/reatt-redes/reatt-redes.utils.ts
```diff
@@ -97,6 +97,16 @@ export class ReattRedesUtils {
...
+    const reportIds = to.report_ids?.length
+      ? [...to.report_ids]
+      : reattRedes.report_ids?.length
+        ? [...reattRedes.report_ids]
+        : undefined;
+    if (reportIds?.length) {
+      to.report_ids = reportIds;
+      reattRedes.report_ids = reportIds;
+    }
+
```
danguyf, line 109: The entirety of the fix. Everything else is unit tests or a possibly misguided defensive shim.

PR #4216, opened 2026-07-30
FECFILE-3311: Add CLEAN flag to force running npm ci.

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3311

Related PRs:
N/A

I also added a fix for an issue I was seeing where the browser reload signal was sent early (and in error) before the build was complete and then again (correctly) when the build completed.  This was because we were creating `.tmp/` and `chown`ing it at the start of every rebuild when that only needed to happen during the initial preparation.

### DIFF
--- a/front-end/scripts/compose-build-render.sh
+++ b/front-end/scripts/compose-build-render.sh
@@ -46,10 +46,12 @@ cleanup() {
 
 trap cleanup EXIT
 
-mkdir -p "$FRONTEND_DIR/.tmp"
-if [[ "$HOST_UID" != "0" ]]; then
-  # immediately chown .tmp to host user
-  chown -R $HOST_UID:$HOST_GID "$FRONTEND_DIR/.tmp" 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


lbeaufort (approved):
Nice improvement

front-end/scripts/compose-build-render.sh
```diff
@@ -91,14 +97,6 @@ else
...
 fi
 
-if npm run "$BUILD_SCRIPT" 2>&1 | tee "$BUILD_LOG_PATH"; then
-  :
-else
-  echo "Initial build failed. Running npm ci and then retrying build..."
-  npm ci
-  npm run "$BUILD_SCRIPT"
-fi
-
```
danguyf, line 101: This appeared to be a stray copy/paste or erroneous merge deconfliction artifact as these lines repeat lines 88-94, albeit outside of the conditional check for `PREPARE_MODE`.  I can see no reason why I would have purposefully run the build a second time.

PR #4223, opened 2026-07-31
Patch/3301 - Fix Reattribution from previous report not showing in dropdown

sasha-dresden:
Issue [FECFILE-3301](https://fecgov.atlassian.net/browse/FECFILE-3301)

Issue was that the reattribute action item should not have been blocked by the reports editable status.
Added an e2e test as well.
You won't be able to save the reattribution for the same problem that exists for redesignation on [FECFILE-3302](https://fecgov.atlassian.net/browse/FECFILE-3302)

### DIFF
--- a/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
+++ b/front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
@@ -177,7 +177,7 @@ export abstract class TransactionListTableBaseComponent
     new TableAction(
       'Reattribute',
       this.createReattribution.bind(this),
-      (transaction: TransactionListRecord) => this.reportIsEditable() && ReattRedesUtils.canReattribute(transaction),
+      (transaction: TransactionListRecord) => ReattRedesUtils.canReattribute(transaction),
       () => true,
     ),
     new TableAction(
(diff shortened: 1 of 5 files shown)


Elaine-Krauss-TCG (approved):

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
```diff
@@ -177,7 +177,7 @@ export abstract class TransactionListTableBaseComponent
     new TableAction(
       'Reattribute',
       this.createReattribution.bind(this),
-      (transaction: TransactionListRecord) => this.reportIsEditable() && ReattRedesUtils.canReattribute(transaction),
```
toddlees, line 180: looking into why this was added recently
it also looks like we removed the condition that this is not an existing to or from.  concerned that that might be another problem.  testing

toddlees (changes requested):

danguyf (approved):
Meets the ticket requirements. Todd raises a good question.

front-end/src/app/reports/transactions/transaction-list/transaction-list-table-base.component.ts
sasha-dresden: Ah good point, got that fixed as well.

PR #4235, opened 2026-08-04
Feature/3330 - Update modal when user clones a transaction

sasha-dresden:
Issue [FECFILE-3330](https://fecgov.atlassian.net/browse/FECFILE-3330)

### DIFF
--- a/front-end/src/styles.scss
+++ b/front-end/src/styles.scss
@@ -903,9 +903,7 @@ p-accordion-content[data-p-active='false'] > div.p-accordioncontent-content {
 
 .primary-icon {
   color: var(--primary-color);
-}
-svg {
-  overflow: visible;
+  flex-shrink: 0;
 }
 
 .warning-icon {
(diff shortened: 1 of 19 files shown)


danguyf (approved):

front-end/src/styles.scss
```diff
@@ -904,6 +904,9 @@ p-accordion-content[data-p-active='false'] > div.p-accordioncontent-content {
 .primary-icon {
   color: var(--primary-color);
 }
+svg {
```
dheitzer: do we want to change this globally like this? I wonder if we would want to just target this dialog unless we plan to check other svgs across the app.

dheitzer (commented):
just one thought

front-end/src/styles.scss
sasha-dresden: That's a good point. The actual problem was that this specific svg is inside a flexbox and therefore shrinking to let the text beside it expand. So what I did instead was set flex-shrink:0 so that the icon will not shrink

dheitzer (approved):

PR #4237, opened 2026-08-05
Feature/3308 - allow very long select inputs text to wrap

sasha-dresden:
Issue [FECFILE-3308](https://fecgov.atlassian.net/browse/FECFILE-3308)

### DIFF
--- a/front-end/src/styles.scss
+++ b/front-end/src/styles.scss
@@ -352,7 +352,8 @@ label.disabled {
 }
 
 .p-select {
-  height: var(--input-height);
+  min-height: var(--input-height);
+  height: fit-content;
 }
 
 .dialog-label {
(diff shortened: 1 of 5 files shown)


front-end/src/styles.scss
```diff
@@ -352,7 +352,8 @@ label.disabled {
 }
 
 .p-select {
-  height: var(--input-height);
+  min-height: var(--input-height);
```
toddlees, line 355: looks like this changed the size for other places we use p-select.

f99 doc type
Image

toddlees (changes requested):

front-end/src/styles.scss
sasha-dresden: Ah good catch. I cleaned it up to be more consistent. It uses minimum height to get our default height, but allows it's height to expand when it's wrapped. And importantly the default top/bottom padding is 0. This is what was causing it to expand aggressively and had previously been kept in check by  setting the height to a fixed value.

danguyf (approved):
This works in my testing. The issues noted by Todd appear resolved.

PR #4245, opened 2026-08-06
[Sprint 93] feature/3315 - Fix so dialog doesn't appear until future Reports has returned.

sasha-dresden:
Issue [FECFILE-3315](https://fecgov.atlassian.net/browse/FECFILE-3315)


danguyf (approved):
No sign of the "Hang on" modal in my local testing.

PR #4249, opened 2026-08-06
feature/3105 - Implement environment banners

max-zaremba-tcg:
Ticket link: [FECFILE-3105](https://fecgov.atlassian.net/browse/FECFILE-3105) 

NOTE: To test this, change 'environmentBanner' on line 6 in 'environment.local.ts' to either 'development', 'test', or 'stage'.

### DIFF
--- a/front-end/cypress/e2e-smoke/pages/usersPage.ts
+++ b/front-end/cypress/e2e-smoke/pages/usersPage.ts
@@ -2,8 +2,8 @@ import { UserFormData } from '../models/UserFormModel';
 import { PageUtils } from './pageUtils';
 
 export class UsersPage {
-  private static readonly addDialogSelector = '#content-offset app-add-committee-member-dialog dialog[open]';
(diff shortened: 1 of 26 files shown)


danguyf (changes requested):
This meets all functional requirements.

My only quibble is that if I click the 'x' to close the banner on the initial "sign in!" page then the banner is back after I sign-in.  From then on it stays closed after you close it but closing it from the sign-in page doesn't persist. Can we save the state in local storage and persist it just for the current session?

danguyf (approved):
Since hiding the banner is only for UI testing and it should be displayed 98% of the time I have no concern about hiding it persisting.

sasha-dresden (changes requested):
This looks pretty good. I did notice a small issue when cookies were disabled. I submitted a [PR](https://github.com/fecgov/fecfile-web-app/pull/4259) into this one which addresses that issue and also should help make this a bit more forward leaning avoiding the use of lifecycles.

sasha-dresden (approved):

PR #4265, opened 2026-08-11
[Sprint 93] feature/3335 - After auditing the only pinning required now is undici and uuid

sasha-dresden:
Issue [FECFILE-3335](https://fecgov.atlassian.net/browse/FECFILE-3335)


danguyf (approved):
Nice!

